PR #5, opened 2017-02-13
Remove an unreachable return statement

olysonek:
This also makes a pylint error message go away.


yarda:
Thanks, it seems useless.

PR #6, opened 2017-02-13
Raise NotImplementedError instead of NotImplemented

olysonek:
NotImplemented is not an exception type. This patch makes some pylint
errors go away.


yarda:
Thanks.

PR #7, opened 2017-02-13
plugin_disk: Fix setting 'readahead_multiply'

olysonek:
Provide the correct number of parameters to the '_set_readahead' method.
This fixes a traceback when setting the 'readahead_multiply' option.

I assume the 'sim' parameter to '_set_readahead' means 'simulated' (it
is handled in that sense), so we should set the value to False, because
we are actually enabling/disabling the tunning here.


yarda:
Thanks.

PR #12, opened 2017-02-14
Remove tuned.gtk.manager

olysonek:
Shall we remove the tuned.gtk.manager module? It doesn't seem to have been used
anywhere _ever_, it's broken (the only class it contains calls
a constructor with incorrect number of parameters) and it seems to be
completely useless anyway.


yarda:
Thanks.

PR #13, opened 2017-02-14
Add 'lint' target to Makefile

olysonek:
The target executes pylint on all *.py files, except for those in the ./tests and ./experiments subdirectories. The tests subdirectory is out-dated (lots and lots of errors) and the experiments subdirectory doesn't seem like a place you'd want to check everytime you make a change. What do you think?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,11 +183,6 @@ test:
 	python -m unittest discover tests
 
 lint:
-	$(eval tmpfile := $(shell mktemp))
-	find . \( -path ./experiments -o -path ./tests \) -prune \
-		-or -name '*.py' \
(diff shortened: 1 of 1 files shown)


yarda:
We should revive the tests in the future, but for now keep it excluded. Experiments are more like sandbox, IIRC it's not currently used in the release.

yarda:
Why not just use?
`$ pylint -E tuned *.py`

or similarly, it's much faster at least.

olysonek:
Nice, it even works around a bug in pylint. I liked the fact that my solution directly prints the path to the affected file, but the execution speed can't be argued with.

yarda:
Yup, it shows module names, so you can derive the path especially if you know the code, but it would be nice RFE, could you file it to pylint upstream, to add option for showing the file name.

PR #14, opened 2017-02-16
Fix logging to file in the current directory

olysonek:
Previously if you ran 'tuned -l log', it would crash with:
OSError: [Errno 2] No such file or directory: ''
This patch fixes the bug.


yarda:
Thanks.

PR #16, opened 2017-02-17
Implement instance priority

olysonek:
Resolves: rhbz#1246172

Allow specifying order in which individual plugin instances are to be
executed.

Each instance can be given the 'instance_priority' option
in the profile configuration. The value must be an integer (it can be
negative). If no priority is specified, it is inherited from the parent
profile, or set to the default instance priority, if no parent profile
specifies the priority for that instance.

The default instance priority is by default zero, but it can be set to
a different value using the 'default_instance_priority' option in the
global configuration file.

Creating plugin instances, assigning devices to instances, initializing
instances, applying, updating and verifying tuning are all done in order
from lowest numerical instance priority to highest. Unapplying tuning and
destroying instances is done in reversed order.

### DIFF
--- a/tuned-main.conf
+++ b/tuned-main.conf
@@ -26,3 +26,6 @@ recommend_command = 1
 # If enabled these sysctls will be re-appliead after Tuned sysctls are
 # applied, i.e. Tuned sysctls will not override system sysctls.
 reapply_sysctl = 1
+
+# Default priority assigned to instances
+default_instance_priority = 0
--- a/tuned/consts.py
+++ b/tuned/consts.py
@@ -56,6 +56,7 @@ CFG_SLEEP_INTERVAL = "sleep_interval"
 CFG_UPDATE_INTERVAL = "update_interval"
 CFG_RECOMMEND_COMMAND = "recommend_command"
 CFG_REAPPLY_SYSCTL = "reapply_sysctl"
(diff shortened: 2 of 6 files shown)


yarda:
Thanks.

yarda:
One question regarding the instance_priority option name. Why not name it just "priority"? IIRC we have discussion about it in the past, but currently it seems the only name conflict there is with the "priority" inside the plugin implementation, thus it shouldn't cause problems, because it's two distinct things - plugin implementation and profile configuration. Most people writing profiles will never write plugins, thus it shouldn't be confusing for majority. For me the "priority" in profile configuration seems more intuitive and easier to remember (especially for newcomers who needn't know what the instance is nor that Tuned use plugin instances).

olysonek:
The point of that was to not make it confusing for me and possibly other tuned contributors :). But it does seem a bit redundant in the profile configuration. I propose that we change the option name in the profile configuration to 'priority', but keep the current naming in the code (i.e. variable names etc.) and also in the global configuration file. I think that if we changed it there, then the name 'priority', taken out of context, wouldn't be descriptive enough. This might make it seem a bit inconsistent, but I think it's a good trade-off.

yarda:
Renamed instance_priority to priority in commit 82f1a9a19d427d0de71f882d948a45ea3e135c99

PR #19, opened 2017-02-27
Remove the '_devices' attribute in plugins

olysonek:
The _device attribute should always be the same as the union of
_assigned_devices and _free_devices, therefore it's redundant.
Its existence in my opinion brings only confusion and a potential
source of inconsistencies.


yarda:
Thanks. I think it's still worth more cleanup (e.g. sometimes in the future).

PR #20, opened 2017-02-28
Implement udev-based device matching

olysonek:
A new option 'devices_udev_regex' can be used in profile
configuration to specify devices to which a plugin instance
should be applied.

The option can contain a python regular expression, as specified
in https://docs.python.org/2/library/re.html#regular-expression-syntax.
The expression is effectively matched against the output of
udevadm info --query=property -n <device_path>

If the option 'devices_udev_regex' is specified, the 'devices' option
is ignored. If it is not specified, then the matching is done the same
way as previously, i.e. against 'devices'.


olysonek:
Hm, yes of course I forgot about hotplug again. I'll deal with that later.

yarda:
Thanks.

PR #21, opened 2017-03-01
Fix plugin cleanup

olysonek:
The plugins were not getting cleaned up properly, so e.g. after
a profile switch instances of the Plugin class would get piled up.

This fixes commit 363d74815a.


olysonek:
The problem applied only to hotplug plugins. The plugin instances were being held by the hardware inventory.

yarda:
Thanks, I have already noticed this behaviour. You saved me some time debugging it :)

PR #24, opened 2017-03-08
Perform command cleanup in reversed order

olysonek:
Clean up device and non-device commands it reversed order, compared to
the order in which they were applied. It makes more sense that way to me.

Related: rhbz#1246176


olysonek:
Our test suite passed after this change.

yarda:
Probably yes, currently I cannot see any problem with the reversed order. Thanks.

PR #25, opened 2017-03-10
Introduce ${PROFILE_DIR} variable

olysonek:
See the individual commit messages for more detail.

Perhaps we could rename the new 'PROFILE_DIR' variable to 'TUNED_PROFILE_DIR' or '_PROFILE_DIR' to avoid possible conflicts with user-defined profiles. We could then say in the documentation that variable names starting with the 'TUNED_' or '_' prefix are reserved for future use. What do you think?

### DIFF
--- a/tuned/consts.py
+++ b/tuned/consts.py
@@ -99,3 +99,12 @@ STR_VERIFY_PROFILE_FAIL = "verify: failed: '%s'"
 
 # timout for tuned-adm operations in seconds
 ADMIN_TIMEOUT = 600
+
+# Constants which can be used in profiles just like regular variables (i.e. ${var_name}).
+# The constant names are reserved, they cannot be redefined in profiles. When adding a new
+# constant, add it to TUNED_RESERVED_CONSTANTS.
+#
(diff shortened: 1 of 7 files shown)


olysonek:
Sorry, github autoformatting messed it up. I meant to say "... 'TUNED_PROFILE_DIR' or '[underscore]PROFILE_DIR' ..."

yarda:
Nice. I understood the first comment even after github improvements :)

Underscore is not bad, but maybe just extend already present concept of built-in functions, i.e.:

${f:FUNC:ARGS} is built-in function, so maybe go with:
${b:PROFILE_DIR}, 'b' - like built-in or maybe 'i' - like internal, or 'r' like runtime or maybe different letter.

Also please switch current profiles to use it in script, i.e. patch script= in profiles directory and keep backward compatibility for relative path for 3rd party profiles which are out of our control.

olysonek:
Sounds good, I will look into it.

olysonek:
I opened a new PR for this: https://github.com/redhat-performance/tuned/pull/26

PR #27, opened 2017-03-13
Introduce script_pre and script_post options

olysonek:


jeremyeder:
I understand PROFILE_DIR, but curious what use-cases there are for the script pre/post and how they're intended to be used?  Any examples?

yarda:
It's related to https://bugzilla.redhat.com/show_bug.cgi?id=1246176 and it also brings more flexibility. I.e. you could have script which will do custom tuning of block devices added through hotplug and/or can trigger various things.

yarda:
Thanks.

PR #28, opened 2017-03-15
Allow running verify_profile_ignore_missing to all users

olysonek:
The policy for running the 'verify_profile_ignore_missing' D-Bus method
was missing, so only root could run it. This commit adds the policy, so
that it is the same as the policy for regular 'verify_profile'.


yarda:
Nice catch, thanks.

PR #29, opened 2017-03-27
Log errors when applying a profile fails

olysonek:
Log errors when applying a profile fails and also log explicitly
when we're trying to load the previously applied profile.

Resolves: rhbz#1434360


yarda:
Great, thanks.

Issue #1, opened 2017-01-21
Not seeing hdd spindown


jflorian:
My understanding is that hdd spindown should occur automatically with the `balanced` profile, right?  I have a simple setup where CentOS 7 is installed on sda and sdb/sdc are joined with LVM to make an iSCSI target (using tgtd).  To rule out disk activity keeping sd[bc] spun up, I've stopped the tgtd service so these two drives should be completely idle.  Yet, whenever I query them:

```
$ sudo smartctl --nocheck standby -i /dev/sdb
smartctl 6.2 2013-07-26 r3841 [x86_64-linux-3.10.0-514.2.2.el7.x86_64] (local build)
Copyright (C) 2002-13, Bruce Allen, Christian Franke, www.smartmontools.org

=== START OF INFORMATION SECTION ===
Model Family:     Western Digital Caviar Black
Device Model:     WDC WD1502FAEX-007BA0
Serial Number:    WD-WMAY03425998
LU WWN Device Id: 5 0014ee 0584b74fe
Firmware Version: 05.01D05
User Capacity:    1,500,301,910,016 bytes [1.50 TB]
Sector Size:      512 bytes logical/physical
Device is:        In smartctl database [for details use: -P show]
ATA Version is:   ATA8-ACS (minor revision not indicated)
SATA Version is:  SATA 2.6, 6.0 Gb/s (current: 3.0 Gb/s)
Local Time is:    Sat Jan 21 13:43:44 2017 EST
SMART support is: Available - device has SMART capability.
SMART support is: Enabled
Power mode is:    ACTIVE or IDLE

$ sudo smartctl --nocheck standby -i /dev/sdc
smartctl 6.2 2013-07-26 r3841 [x86_64-linux-3.10.0-514.2.2.el7.x86_64] (local build)
Copyright (C) 2002-13, Bruce Allen, Christian Franke, www.smartmontools.org

=== START OF INFORMATION SECTION ===
Model Family:     Western Digital Caviar Green (AF, SATA 6Gb/s)
Device Model:     WDC WD20EZRX-00D8PB0
Serial Number:    WD-WMC4M2715769
LU WWN Device Id: 5 0014ee 6aef40058
Firmware Version: 80.00A80
User Capacity:    2,000,398,934,016 bytes [2.00 TB]
Sector Sizes:     512 bytes logical, 4096 bytes physical
Rotation Rate:    5400 rpm
Device is:        In smartctl database [for details use: -P show]
[...]

PR #31, opened 2017-03-28
disk: Don't apply lower spindown value if drive has spun down

olysonek:
Don't run 'hdparm -S' with a lower spindown value than previously
applied, if the drive has already spun down. Executing the command
in that case makes the drive spin up with some HDDs.

Resolves https://github.com/redhat-performance/tuned/issues/1

### DIFF
--- a/tuned/plugins/plugin_disk.py
+++ b/tuned/plugins/plugin_disk.py
@@ -94,6 +94,7 @@ class DiskPlugin(hotplug.Plugin):
 			instance._device_idle = {}
 			instance._stats = {}
 			instance._idle = {}
+			instance._spindown_change_delayed = {}
 		else:
 			instance._has_dynamic_tuning = False
 			instance._load_monitor = None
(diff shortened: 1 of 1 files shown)


olysonek:
"In such case it means we need better performance." - I don't understand what you mean by this. I think it's quite the opposite (?). I added the check so that if we're switching to a more performance-oriented level, the change gets applied. Otherwise we could, at least theoretically, get stuck in a very powersaving level (large spindown value) even when the drive is relatively active.

I agree with the second point.

yarda:
If I am not mistaken I can see level_change > 0, which means higher index in levels table, i.e. lower spindown, shouldn't it be in reverse?

olysonek:
Unless I'm crazy, no :D. If we're setting a lower spindown (we want to save more power), then we want to suppress the change, if the drive is not moving, so we don't wake it up unnecessarily.

yarda:
OK. With the delay flag I think the check needn't be there.

olysonek:
I would rather keep it. I really don't think we should delay increasing the spindown, it could potentially result in poor performance.

yarda:
Well I think it's unlikely that the metrics which is based on the disk activity will increase the spindown timeout when the disk doesn't rotate. The minimum spindown timeout we set is 300 seconds, the metric is recalculated every 10 seconds, so I think there shouldn't be difference w/wo the level check.

But, the update interval is user configurable, so theoretically the user could set it to e.g. 400 seconds or higher (I don't know why anybody would do it, but the possibility is there) and in such case it could be performance problem, so merging it as is, thanks.

PR #32, opened 2017-04-12
[WIP] Video plugin improvements

olysonek:
Support the 'dpm' power method on Radeon GPUs. Implement automatic selection of the best power method and use it in the balanced profile.

Resolves: rhbz#1417659

### DIFF
--- a/tuned/plugins/plugin_video.py
+++ b/tuned/plugins/plugin_video.py
@@ -29,6 +29,7 @@ class VideoPlugin(base.Plugin):
 	def _get_config_options(self):
 		return {
 			"radeon_powersave" : None,
+			"radeon_dpm_state" : None,
 		}
 
 	def _instance_init(self, instance):
@@ -43,6 +44,9 @@ class VideoPlugin(base.Plugin):
 			"method" : "/sys/class/drm/%s/device/power_method" % device,
 			"profile": "/sys/class/drm/%s/device/power_profile" % device,
 		}
+
+	def _radeon_dpm_state_file(self, device):
+		return "/sys/class/drm/%s/device/power_dpm_state" % device
 
 	@command_set("radeon_powersave", per_device=True)
 	def _set_radeon_powersave(self, value, device, sim):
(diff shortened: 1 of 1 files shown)


olysonek:
Do not merge. I doesn't work.

yarda:
Is it ready for merge?

yarda:
I do not have HW to test :)

olysonek:
I don't know, I can't test it either. Let's wait for Oliver.

tuned/plugins/plugin_video.py
```diff
@@ -61,6 +65,12 @@ def _set_radeon_powersave(self, value, device, sim):
 			if not sim:
 				self._cmd.write_to_file(sys_files["method"], "dynpm")
 			return "dynpm"
+		elif value == "dpm":
+			if not sim and self._get_radeon_powersave(device) != "dpm":
+				log.error("Cannot set dpm power method through the video plugin." \
+						+ "Use the sysctl plugin with the 'radeon.dpm=1' option")
```
oliverhenshaw: I don't think this is strictly correct. The radeon.dpm option is a boot-time parameter, not one that can be changed with sysctl.

olysonek: Thanks, I guess I meant to say the bootloader plugin.

olysonek:
>A couple of niggles though - adding "radeon_dpm_state=balanced" is not enough to avoid complaints in
>tuned.log and from "tuned-adm verify". I need to add "radeon_powersave=dpm" too.

I don't understand. What exactly are you seeing in the log? Adding "radeon_powersave=dpm" should actually _cause_ an error message.

>Secondly, "tuned-adm verify" doesn't seem to do anything to check radeon_dpm_state - I tried setting it to a nonsense value and although I got an error when changing profile I didn't see any complaints when I ran the verify subcommand.

The verify command seems to behave badly in some cases when setting nonsensical values. The same happens in the CPU plugin with governor for example. We'll have to deal with that later in some more generic way. But I think the verify command should now correctly detect that the current value is different from a _legal_ value in the tuned profile.

yarda:
Thanks, LGTM, I will also probably update the profiles to use the DPM by default.

yarda:
I rewrote it a bit in eef09f1185476ed54bbd3f08eac28e4e8e02418e. Unfortunately I do not have compatible video card, so I wasn't able to test it.

PR #33, opened 2017-04-13
plugin_net: Fix invocation of execute()

olysonek:
Fix an invocation of execute() after its API changed.

Reproducer:
Run tuned with the following profile, and you'll see error messages
'tuned.utils.commands: Executing ethtool error: ethtool: bad command line argument(s)'
in the log. Replace 'enp0s25' with your ethernet interface:

[net]
devices=enp0s25
coalesce=rx-usecs 4


yarda:
Thanks, very nice catch.

PR #34, opened 2017-04-18
try to get unitdir and tmpdir from pkg-config

evgeni:
pkg-config is the most reliable and cross-distro way to get information about the systemd folders on a system.
only rely on RPM macros when pkg-config is not installed or systemd.pc is not available.


yarda:
Thanks.

PR #38, opened 2017-04-18
Start dbus exports after a profile is applied

olysonek:
So this was my initial idea of what the patch might look like. But maybe we could remove the timeout, so that it waits indefinitely for a profile to be applied / for exports? It would be better for debugging - things wouldn't time out when stepping through the code. We would still keep the log messages 'Waiting for ...', so that the user knows what's going on, in case things are taking too long.


olysonek:
Closing in favor of https://github.com/redhat-performance/tuned/pull/74.

PR #39, opened 2017-04-21
List available profiles on 'tuned-adm profile'

olysonek:
List available profiles when 'tuned-adm profile' is run
without a profile name.

Resolves: rhbz#988433


yarda:
Thanks.

jeremyeder:
:+1:

PR #40, opened 2017-04-22
fix NF/FI macros in manpages

evgeni:
the macros are called .nf and .fi, not .NF and .FI

Test:

    % LC_ALL=en_US.UTF-8 MANROFFSEQ='' MANWIDTH=80  man --warnings \
      -E UTF-8 -l -Tutf8 -Z <file> >/dev/null
    53: warning: macro `NF' not defined
    74: warning: macro `FI' not defined

Spotted by rpmlint and lintian.

### DIFF
--- a/man/tuned-adm.8
+++ b/man/tuned-adm.8
@@ -72,7 +72,7 @@ performance or balanced profile if unsure.
 Unload tunings.
 
 .SH "FILES"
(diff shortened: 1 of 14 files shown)


yarda:
You are right, thanks for catching it.

PR #42, opened 2017-04-22
drop the encoding keyword from the desktop file

evgeni:
it's deprecated:
 https://specifications.freedesktop.org/desktop-entry-spec/1.0/apc.html


yarda:
Thanks.

PR #50, opened 2017-05-10
Updated help outputs

voxadam:
Added note about restarting tuned using 'systemclt' in addition to the existing reference to 'service'.


yarda:
Thanks.

PR #51, opened 2017-05-15
Add negation support for cpulist_upack function

### LINKED ISSUE #49: [RFE] Support for negation "^" in the cpulist_unpack function
In OpenStack, some of the configuration support negation, like, having the cpu list as ["1-5", "^3"], should eventually add the list as "1,2,4,5". We have a scenario where we need to pass the same list of parameter to both OpenStack and tuned (tuned-profile-cpu-partitioning's isolated_cores). 

The expansion can be added at the caller, and it has to be added in multiple places. If it is done in the tuned it would be nice to have feature and has consistency.

krsacme:
cpulist_unpack supports range, it would be good add negation
support via ^ or !, inorder to exclude cpus from the specified
range.
Fixes #49

### DIFF
--- a/tuned/utils/commands.py
+++ b/tuned/utils/commands.py
@@ -260,7 +260,9 @@ class commands:
 	# Unpacks CPU list, i.e. 1-3 will be converted to 1, 2, 3, supports
 	# hexmasks that needs to be prefixed by "0x". Hexmasks can have commas,
 	# which will be removed. If combining hexmasks with CPU list they need
-	# to be separated by ",,", e.g.: 0-3, 0xf,, 6
+	# to be separated by ",,", e.g.: 0-3, 0xf,, 6. It also supports negation
+	# cpus by specifying "^" or "!", e.g.: 0-5, ^3, will output the list as
+	# "0,1,2,4,5" (excluding 3). Note: negation supports only cpu numbers.
 	def cpulist_unpack(self, l):
 		rl = []
 		if l is None:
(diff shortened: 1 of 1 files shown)


krsacme:
@yarda ptal.

yarda:
Thanks, merged, but is it exactly what you need? It seems it only removes negated CPUs (which the original request seems to be about), but is it enough? I.e.:

${f:cpulist_unpack:1-3,!2}

expands to 1, 3, but on 8 core machine shouldn't it also add CPUs 0 and 4-7, i.e. for the result to be 0, 1, 3, 4, 5, 6, 7?

krsacme:
@yarda the negation will be applied only on the range which has been added.

krsacme:
OpenStack Reference-

https://docs.openstack.org/ocata/config-reference/compute/config-options.html
vcpu_pin_set
	
(String) Defines which physical CPUs (pCPUs) can be used by instance virtual CPUs (vCPUs).

Possible values:

A comma-separated list of physical CPU numbers that virtual CPUs can be allocated to by default. Each element should be either a single CPU number, a range of CPU numbers, or a caret followed by a CPU number to be excluded from a previous range. For example:
vcpu_pin_set = “4-12,^8,15”

yarda:
Thanks for info.

krsacme:
@yarda any input on how the release cycle works and when I can expect the package in fdp channels?

yarda:
If you need this functionality in RHEL, please open Red Hat bugzilla. I think backport/z-stream/FDP update shouldn't be a problem. Otherwise it will probably not get there sooner than in RHEL-7.5.

PR #56, opened 2017-07-12
Remove unnecessary shebang from gtk python modules.

felixsch:
Currently the modules in tuned/gtk/* include shebangs. Since these modules are not intended for executing remove them.

This will also silence rpmlint warnings.


yarda:
Thanks, merged.

PR #61, opened 2017-08-15
Increase udev monitor buffer size, make it configurable

olysonek:
The udev monitor buffer size was increased to 1MB. Also, a new
configuration option 'udev_buffer_size' can be specified in
/etc/tuned/tuned-main.conf to set a different buffer size.
Values such as '204800', '200KB', '200 KB' (with space) or '1 MB'
are recognized. If no unit is specified, the number is taken as
the number of bytes.

Resolves: rhbz#1442306


yarda:
Thanks, merged.

PR #62, opened 2017-08-21
cpu: Don't log error in execute() if EPB is not supported

olysonek:
Don't log error in commands.execute() if MSR_IA32_ENERGY_PERF_BIAS
is not supported. A message with level WARN is already logged
afterwards:

WARNING  tuned.plugins.plugin_cpu: your CPU doesn't support
         MSR_IA32_ENERGY_PERF_BIAS, ignoring CPU energy performance bias

Resolves: rhbz#1443182


yarda:
Thanks, merged.

PR #63, opened 2017-08-28
Improve logging

olysonek:
Note: With commit 66d4843 you get tons of warnings in the logs when the isolated_cores parameter is set, such as the following:
tuned.plugins.plugin_scheduler: Affinity of PID 9534 is not changeable.
tuned.plugins.plugin_scheduler: Affinity of PID 522 is not changeable.

It seems all the PIDs it's complaining about are kernel threads. Maybe we could change the behaviour so that:
* Tuned doesn't complain about not being able to change affinity of kernel threads, and/or
* Log only one message after the migration is completed saying that _some_ tasks could not be migrated

### DIFF
--- a/tuned/plugins/base.py
+++ b/tuned/plugins/base.py
@@ -457,9 +457,9 @@ class Plugin(object):
 				log.warn("cannot compare new value '%s' with current value '%s' by operator '%s', using '%s' directly as new value" % (val, current_value, op, new_value))
 		return new_value
 
-	def _get_current_value(self, command, device = None):
+	def _get_current_value(self, command, device = None, ignore_missing=False):
 		if device is not None:
-			return command["get"](device)
+			return command["get"](device, ignore_missing=ignore_missing)
 		else:
 			return command["get"]()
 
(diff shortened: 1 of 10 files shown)


yarda:
Thanks, the current plugin_scheduler code is mess, but could you use e.g. procfs for '_pid_exists' implementation? It's just about checking that the /proc/PID file exists, using the whole exec/ps machinery for it is a big overhead. Also I wouldn't do it by just checking that the path /proc/PID exists, I would use procfs library for it. There will be still false positives, but I think it will improve the situation.

yarda:
> There will be still false positives, but I think it will improve the situation.
I meant that the PID can vanish everywhere between the call of taskset and the check and in such case different errors could be wrongly suppressed. But it shouldn't happen often.

olysonek:
> I meant that the PID can vanish everywhere between the call of taskset and the check and in such case different errors could be wrongly suppressed. But it shouldn't happen often.

Well if the PID vanishes between the call to taskset and the check for its existence, then we don't really care if we failed to set its affinity, since it's already gone anyway right? So I think logging it with debug level is fine.

olysonek:
For the record: We decided to solve the problem in my initial comment by logging failures to set affinity in kernel threads and zombie processes (empty /proc/[pid]/cmdline) as DEBUG. The rationale is that kernel threads seem to be often pinned to specific CPUs and there's nothing to do about it, so complaining about it in the logs is annoying and pointless. For zombie processes, if we fail to set their affinity, then we don't really care either as those processes won't run on the CPU ever again.

yarda:
Thanks, merged.

PR #67, opened 2017-09-11
Fix race between DBus exporter startup and applying a profile

olysonek:
Previously if the profile was applied before all methods and
signals were exported on DBus, the daemon thread (_thread_code()
in daemon.py) could crash after attempting to emit a signal on DBus
announcing that a profile was applied.

A regular user is unlikely to run into the race in practice, but it
frequently happens when applying an empty profile for testing purposes.


yarda:
Thanks, merged.

PR #70, opened 2017-09-13
Replace ethtool invocation in scripts by built-in functionality

olysonek:


yarda:
Please drop eth0, I guess it should work just with the line deleted, because there is: https://bugzilla.redhat.com/show_bug.cgi?id=1490399

yarda:
It seems it's not correct according to linked bug, maybe we could make this option optional through the variable.

yarda:
Thanks, merged

Issue #64, opened 2017-08-29
Tuned downgrade is troublesome because of new automatically/manually set profile functionality


olysonek:
The current implementation of the functionality where we differentiate between automatically and manually set profile is causing problems when downgrading tuned. We have two lines in /etc/tuned/active_profile, e.g.:
balanced
manual

The profile name will be incorrectly recognized by old tuned from such a file as "balanced\nmanual". Let's keep the info whether the profile was set automatically/manually in a separate file, e.g. /etc/tuned/active_profile_mode.

PR #71, opened 2017-09-15
Store profile selection mode in a separate file

olysonek:
The profile selection mode is now written to /etc/tuned/profile_mode
instead of the second line of /etc/tuned/active_profile.

Here are the rules for interpreting the contents of the files. If
either of the files does not exist, it is treated the same as if
it were empty.

If active_profile is empty:
    If profile_mode contains 'manual':
        Manual mode, no profile will be used, tuned will run without a profile
    else if profile_mode is empty or contains 'auto':
        Automatic mode, recommended profile is used
    else:
        Error
else:
    If profile_mode contains 'manual', or is empty (for compatibility reasons):
        Manual mode, the profile in active_profile will be used
    else if profile_mode contains 'auto':
        Automatic mode, recommended profile is used
    else:
        Error

Resolves: https://github.com/redhat-performance/tuned/issues/64


olysonek:
Hrmm, see the commit message if you want it displayed properly :).

yarda:
Please also update the spec file - add there profile_mode

olysonek:
Done

yarda:
Thanks, merged.

PR #72, opened 2017-09-19
plugin_video: Prefix DPM states with 'dpm-'

olysonek:
In the end, I like your version better. It's short and sweet. But let's prefix DPM states with 'dpm-' to prevent future name collisions. Graphics cards manufacturers are very creative, I'm sure they will come up with a completely new power management mechanism soon :).

Also, change the state in powersave profiles to 'battery' - for some strange reason, there's
no such DPM state as 'powersave' and the lowest power consumption state is called 'battery'.

### DIFF
--- a/profiles/balanced/tuned.conf
+++ b/profiles/balanced/tuned.conf
@@ -13,7 +13,7 @@ energy_perf_bias=normal
 timeout=10
 
 [video]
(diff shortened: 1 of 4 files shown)


yarda:
Thanks, merged.

PR #77, opened 2017-10-16
Expand variables in the 'devices' option

olysonek:
Variables in the 'devices' option were not being expanded. This is
bad. It shows that the current variable expansion system is broken.
We need to come up with a consistent, well-defined way of expanding
variables. Preferably it should be done all in one place.

The expansion is probably still not done for options 'type', 'enabled',
'replace', 'devices_udev_regex', 'script_pre' and 'script_post'. Let's
deal with that later, as we don't need it in our profiles right now.

Related: rhbz#1490399


yarda:
Thanks, merged.

PR #79, opened 2017-10-20
port realtime virtual host profile to Fedora

bonzini:
RHEL ships tscdeadline_latency.flat as part of the qemu-kvm-tools package, but Fedora does not. In addition, the script that runs it refers to /usr/libexec/qemu-kvm that doesn't exist in Fedora (where qemu-kvm is simply on the path).

The solution is to package the benchmark in tuned, and once we do this we can look it in /usr/share/tuned instead of /usr/share/qemu-kvm.

Signed-off-by: Paolo Bonzini <pbonzini@redhat.com>


yarda:
More details in:
https://bugzilla.redhat.com/show_bug.cgi?id=1504680

yarda:
Thanks.

PR #85, opened 2018-01-14
Python3 support

olysonek:
An alternative to using sed to rewrite shebangs in commit b716ebfe55 might be to rewrite the build system using python setuptools. Any thoughts?

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -4,3 +4,7 @@ Installation instructions
 The tuned daemon is written in pure Python. Nothing requires to be built.
 
 For installation use 'make install'. Optionally DESTDIR can be appended.
+By default, the tuned modules are installed to the Python2 destination
+(e.g. /usr/lib/python2.7/site-packages/) and shebangs in executable
+Python files are modified to use Python2. If you want tuned to use Python3
+instead, use 'make PYTHON=python3 install'.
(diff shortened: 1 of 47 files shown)


yarda:
Thanks, LGTM. One minor thing, for python2 installation I would use:
+BuildRequires: python2, python2-devel
+Requires: python2-decorator, ...

I.e. the 2 suffix, it would allow rebuilding with python2 support on f27+ (if somebody has such requirement). And adding 'with' macro support to SPEC (i.e. %bcond_with ...) would be also nice. But it can be resolved later, so merging.

olysonek:
> One minor thing, for python2 installation I would use:
> +BuildRequires: python2, python2-devel
> +Requires: python2-decorator, ...

The problem is that on RHEL-7, there is only e.g. python-decorator, not python2-decorator. So it would need some more conditionalizing. But I can add it, if you think it's worth it.

yarda:
I am not sure whether there will be anybody with such use case (tuned on python2 on f27+), but maybe it could help us checking whether some RHEL-7 problem is reproducible with new Fedora kernel, so I would probably add it.

I would use something like:
```
%bcond_without python3
%if %{rhel} < 8
%global py python
%else
%if %{with python3}
%global py python3
%else
%global py python2
%fi
%fi
...
Requires: %{py}-decorator
```

PR #87, opened 2018-01-19
Add support for CentOS

oldmikeyang:
Add the centos grub.cfg file to the list


olysonek:
Perhaps it would be better to use the symlink /etc/grub2-efi.cfg instead of /boot/efi/EFI/*/grub.cfg. @yarda, what do you think?

PR #88, opened 2018-01-20
Locally embedded web fonts

mrksu:
- The Overpass web font is now embedded locally instead of downloading it from Google Fonts. It includes regular, bold, italic, and bold-italic styles.

- The Hack monospace font has also been added as an embedded web font. The repo contains four styles, but only regular is currently enabled for quicker page loading.

    I've chosen Hack mostly because, as opposed to Overpass Mono, it supports italics, which will be useful later on for the actual Tuned documentation. Roboto Mono is also an alternative if you prefer.

- Text alignment is now set to justified.

- List items are indented also from the right to be slightly shorter and more compact.


yarda:
Thanks, merged.

PR #90, opened 2018-02-05
Makefile: refactor

olysonek:
Introduce a function for installing python scripts and make use of it in the Makefile. It makes it look a lot cleaner.

We talked about using setuptools in tuned for installation, but after working on it for a while a came to the conclusion that it is really, really not a good idea. Basically, distutils/setuptools don't have the functionality we need and so we have to implement it ourselves. And I don't see the point of using setuptools, if we have to implement a lot of the functionality ourselves anyway. Namely, setuptools do not support at least the following:
 * renaming files
 * creating empty files
 * installing desktop files
 * installing scripts to arbitrary locations (/usr/sbin, /usr/libexec)

I created an implementation of the first three things here:
https://github.com/olysonek/tuned/blob/setuptools/setup.py

I'm still struggling to get the last thing to work. And my partial implementation of it brings the line count of setup.py up to 275 SLOC, which is quite a lot compared to the Makefile in this PR (223 SLOC), considering the Makefile includes a lot of additional functionality.

Even if we disregard the size of setup.py, using setuptools is still a big maintainance burden. It could happen that something in distutils changes, breaking our setup script. Or perhaps a user finds out that the setup.py script doesn't work in his particular situation and we have to fix it (this is a very realistic scenario BTW, because extending setuptools requires understanding it, which I can't claim to do). By having a simple Makefile, we don't create any obligation to support what other setup.py scripts support.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,12 @@ TUNED_RECOMMEND_DIR = $(TUNED_PROFILESDIR)/recommend.d
 TUNED_USER_RECOMMEND_DIR = $(SYSCONFDIR)/tuned/recommend.d
 BASH_COMPLETIONS = $(DATADIR)/bash-completion/completions
 
+copy_executable = install -Dm 0755 $(1) $(2)
+rewrite_shebang = sed -i -r -e $(SHEBANG_REWRITE_REGEX) $(1)
+restore_timestamp = touch -r $(1) $(2)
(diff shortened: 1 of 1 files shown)


olysonek:
If you want, I can keep working on setup.py, but I really think sticking to the Makefile is better.

yarda:
Thanks for looking into it. It makes sense.

yarda:
Merged.

PR #92, opened 2018-03-01
Consistently use 'balance' instead of 'rebalance' in cpu-partitioning

olysonek:


yarda:
Thanks, merged.

PR #94, opened 2018-03-07
tuned: Add readahead comment to throughput-performance tuned.conf

jtpittman195:
A common source of confusion within tuned is the lack of suffix
on the readahead specification in the throughput-performance
tuned.conf. This confusion is resolved, in general, by comparing
the value, for example, to blockdev or lvdisplay output. Avoid
future end-user confusion by adding a comment to the
throughput-performance tuned.conf file, explaining that the default
is in KiB, and noting the allowance of a sector suffix as well.

Signed-off-by: John Pittman <jpittman@redhat.com>

### DIFF
--- a/profiles/throughput-performance/tuned.conf
+++ b/profiles/throughput-performance/tuned.conf
@@ -11,6 +11,9 @@ energy_perf_bias=performance
 min_perf_pct=100
 
 [disk]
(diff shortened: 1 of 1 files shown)


yarda:
Thanks, merged.

PR #100, opened 2018-05-14
Fix setting EPB on Linux 4.13+, support specifying alternative EPB values

olysonek:
Resolves: rhbz#1508468


yarda:
Thanks, merged.

PR #104, opened 2018-06-11
Print log excerpt in tuned-adm when changing the profile

olysonek:
Resolves: rhbz#1538745


olysonek:
I added one more commit.

yarda:
Thanks, merged.

PR #111, opened 2018-07-09
tuned-adm: Fix a traceback when run without arguments

olysonek:
Running tuned-adm without arguments using Python 3 results in
a traceback. This is because in this case parse_args does not
exit with a usage message in Python 3 and the 'action' option
is then undefined. So let's check for this case and print the
usage message ourselves. There may be a better solution to this,
but I wasn't able to come up with anything in a reasonable amount
of time.

### DIFF
--- a/tuned-adm.py
+++ b/tuned-adm.py
@@ -94,6 +94,9 @@ if __name__ == "__main__":
 	parser_profile_mode = subparsers.add_parser("profile_mode", help="show current profile selection mode")
 	parser_profile_mode.set_defaults(action="profile_mode")
 
+	if len(sys.argv) < 2:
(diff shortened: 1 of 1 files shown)


yarda:
Thanks, in python 3.7 add_subparsers method has 'required' option which should IMHO introduce the same behaviour as in 2.7, but there is no such option in 3.6, so this commit is probably the right way to go.

yarda:
This needs different approach, because:
```
$ python3 ./tuned-adm.py -d -a
  File "./tuned-adm.py", line 106, in <module>
    action_name = options.pop("action")
KeyError: 'action'
```
Probably better way is to add exception handler for KeyError.

PR #114, opened 2018-07-10
tuned-gui: Sort plugins based on their name

olysonek:
Previously the sorting was done by comparing the objects themselves,
which is not what we want and it doesn't work in Python 3 - TypeError
is raised, e.g:
TypeError: '<' not supported between instances of 'BootloaderPlugin' and 'MountsPlugin'


yarda:
Thanks, LGTM.

PR #119, opened 2018-08-14
Show rollback info message only when in daemon mode.

jmencak:
This commit removes misleading info message about rolling back changes when running tuned in one-shot mode.

### DIFF
--- a/tuned/daemon/daemon.py
+++ b/tuned/daemon/daemon.py
@@ -173,8 +173,11 @@ class Daemon(object):
 			# do full cleanup
 			full_rollback = False
 			if self._full_rollback_required():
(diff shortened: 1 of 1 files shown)


olysonek:
Thanks, LGTM. @yarda ?

olysonek:
But maybe add a hint to the new message saying that Tuned was running in oneshot mode.

olysonek:
I see no problem with that. But @jmencak , please add that hint about oneshot mode to the message, it may be useful for us when we need to analyze customer logs.

yarda:
OK, thanks, merging, we will add the hint ourselves.

PR #120, opened 2018-09-05
log: add two options(log_file_count and log_file_maxbytes) in the main

ldzhong:
config file

Currently the logrotation is not configurable. On system with large
amount of cpus a tuned start creates a lot of entries and older log
entries are lost.

### DIFF
--- a/tuned-main.conf
+++ b/tuned-main.conf
@@ -32,3 +32,9 @@ default_instance_priority = 0
 
 # Udev buffer size
 udev_buffer_size = 1MB
+
+# Log file count
+log_file_count = 2
+
+# Log file max bytes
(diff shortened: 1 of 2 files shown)


olysonek:
Nice, I've been meaning to do this. Thanks for doing it for me :). LGTM.

ldzhong:
Thanks for review. But I am closing this pull request because classmethod can't access object attribute.
Will send a new version patch.

olysonek:
That seemed fishy to me indeed, but I assumed you tested it... Also consider renaming `log_file_maxbytes` to `log_file_maxsize` or something. The get_size() method supports parsing `12 MB` and similar.

PR #124, opened 2018-09-24
Add support for syspurpose and chassis type matching in recommend.conf

olysonek:


yarda:
Thanks, I didn't test it, but LGTM.

PR #129, opened 2018-10-09
commands: Fix chassis type matching

olysonek:
Resolves: rhbz#1565598


yarda:
Thanks, merged.

PR #130, opened 2018-10-10
Make python-dmidecode a weak dependency

olysonek:
It's only available on x86_64.

Resolves: rhbz#1565598


yarda:
Thanks, merged.

PR #132, opened 2018-10-18
Make virtual-host identical to latency-performance

### LINKED ISSUE #131: Changes required in virtual-host profile
The default tuned profile for an oVirt cluster is "virtual-host". Recently, testing at customer sites and performance labs has shown that "latency-performance" yields better results. Therefore the configuration in virtual-host profile should be made identical to latency-performance. In the future it can be tweaked as understanding regarding the contribution of the specific configuration items to ovirt nodes performance grows.

See: https://bugzilla.redhat.com/1588932 for further information

oliel:
The default tuned profile for an oVirt cluster is "virtual-host".
Recently, testing at customer sites and performance labs has shown
that "latency-performance" yields better results. This patch makes
virtual-host profile identical to latency-performance. This seems
better than uniting them since this way we maintain the possibility
to tweak them in different ways in the future according to further
findings.

fixes #131 

Change-Id: I1a64fac22364b17de1279a3e1f37a6b7eb0fe61b
Bug-Url: https://bugzilla.redhat.com/1588932
Signed-off-by: Ori_Liel <oliel@redhat.com>

### DIFF
--- a/profiles/virtual-host/tuned.conf
+++ b/profiles/virtual-host/tuned.conf
@@ -4,42 +4,4 @@
 
 [main]
 summary=Optimize for running KVM guests
-
-[cpu]
-force_latency=1
-governor=performance
-energy_perf_bias=performance
-min_perf_pct=100
-
-[sysctl]
-# ktune sysctl settings for rhel6 servers, maximizing i/o throughput
-#
-# Minimal preemption granularity for CPU-bound tasks:
-# (default: 1 msec#  (1 + ilog(ncpus)), units: nanoseconds)
-kernel.sched_min_granularity_ns=10000000
-
-# If a workload mostly uses anonymous memory and it hits this limit, the entire
-# working set is buffered for I/O, and any more write buffering would require
-# swapping, so it's time to throttle writes until I/O can catch up.  Workloads
-# that mostly use file mappings may be able to use even higher values.
-#
-# The generator of dirty data starts writeback at this percentage (system default
-# is 20%)
-vm.dirty_ratio=10
-
-# Start background writeback (via writeback threads) at this percentage (system
-# default is 10%)
-vm.dirty_background_ratio=3
-
-# The swappiness parameter controls the tendency of the kernel to move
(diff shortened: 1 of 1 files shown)


oliel:
@yarda could you please approve this patch? Thanks

yarda:
@oliel thanks for the patch, but what about just including the latency-performance? In such case you could just tweak what different in the future.

We will need RHEL Tuned bugzilla to fix it in RHEL.

jeremyeder:
agree with @yarda suggestion - don't use latency-performance itself, just make yours include it (we have a few other profiles that already do that, and then add other custom stuff on top of latency-performance).

oliel:
Thanks that makes sense, I will do that.

oliel:
Done, looks ok?

yarda:
@oliel I am OK with it now, thanks.

@mwperina I think I wasn't in the thread containing Douglas recommendations. So the current PR is not what you want?

oliel:
I wasn't aware of Douglas recommendations, I will change the patch again

yarda:
So we are back with the throughput-performance? And the only change to the current profile is addition of force_latency=70?

oliel:
That's the idea, yes

yarda:
Meged, thanks.

jeremyeder:
the /dev/cpu_dma_latency value of 70 isn't guaranteed to get you into C3, it just so happens to be the case on many of the current model Intel CPUs.  I'm sure you know this, so just adding context to this github issue in case it bites us in the future.

More future-proof thing to do would be to identify the C3 exit latency by looking at the system's sysfs, then programming that via tuned.

yarda:
Well, we could add this functionality to Tuned, e.g. when you specify latency as C*, e.g. C3 it will find out the exact number from the sysfs. Feel free to open RFE bugzilla.

rgolangh:
@yarda @jeremyeder @oliel one of the main points against aliasing is if latency profile will do something which is not balanced for virtualization - please wait before approving that.

PR #136, opened 2018-11-21
Disable KSM only once, re-enable it only on full rollback

olysonek:
Disabling the ksm and ksmtuned services during boot seems to cause
problems, so do it only once when the profile is first applied and
reenable the services only when full rollback is required.

Resolves: rhbz#1622239


yarda:
Thanks, merged.

PR #137, opened 2018-11-21
Add range feature for cpu exclusion

TomasKorbar:
- It is now possible to exclude cpu range by !1-3
- Also edited test for cpulist_unpack method to test this new
feature
- Resolves rhbz#1533908


yarda:
Thanks, merged.

PR #139, opened 2018-12-11
logs: actually use SystemRandom source for generating a random string

mgerstner:
The returned string was only pseudo random. It is only used for the
log_capture_start() mechanism at the moment. Therefore an attacker might
be able to guess the tokens of other users. Since this data it publicly
available anyways it should not be a big issue, however.


yarda:
Thanks, it makes sense, merged.

PR #141, opened 2018-12-12
fix disk plugin/plugout problem

TomasKorbar:
when udev sends remove event it does not send any device
attributes with it so _device_is_supported check must be omitted
also when getting devices sys_name ensure it is string object
by using str()
Resolves: rhbz#1595156

### DIFF
--- a/tuned/monitors/base.py
+++ b/tuned/monitors/base.py
@@ -99,13 +99,13 @@ class Monitor(object):
 		self._refresh_updating_devices()
 
 	def add_device(self, device):
-		assert isinstance(device, str)
+		assert (isinstance(device,str) or isinstance(device,unicode))
 		if device in self._available_devices:
 			self._devices.add(device)
 			self._updating_devices.add(device)
 
 	def remove_device(self, device):
-		assert isinstance(device, str)
+		assert (isinstance(device,str) or isinstance(device,unicode))
 		if device in self._devices:
 			self._devices.remove(device)
 			self._updating_devices.remove(device)
--- a/tuned/plugins/plugin_disk.py
+++ b/tuned/plugins/plugin_disk.py
@@ -50,7 +50,7 @@ class DiskPlugin(hotplug.Plugin):
 		self._hardware_inventory.unsubscribe(self)
 
 	def _hardware_events_callback(self, event, device):
-		if self._device_is_supported(device):
+		if self._device_is_supported(device) or event == "remove":
 			super(DiskPlugin, self)._hardware_events_callback(event, device)
 
 	def _added_device_apply_tuning(self, instance, device_name):


tuned/plugins/hotplug.py
```diff
@@ -32,7 +32,7 @@ def _hardware_events_callback(self, event, device):
 			self._remove_device(device)
 
 	def _add_device(self, device):
-		device_name = device.sys_name
+		device_name = str(device.sys_name)
```
olysonek: I don't have a problem with this per se, but is there a particular problem this is trying to solve?

TomasKorbar: Yes, the problem here was that sys_name was not always string object and because of that assertions on lines https://github.com/redhat-performance/tuned/blob/master/tuned/monitors/base.py#L102 and https://github.com/redhat-performance/tuned/blob/master/tuned/monitors/base.py#L108 were failing.

olysonek: So what was the content of sys_name? Do you know how to reproduce it? I'd like to take a look.

TomasKorbar: https://tkorbar.fedorapeople.org/reproduce_sysname_problem.py
If you will run this file with python2 then you will see that unicode type is printed out. In pyudev 0.21 which is running on python2 sys_name is unicode object.

olysonek: I see. Unless there's also some other reason to convert it, I would instead modify the assertion to accept unicode objects. Converting unicode objects like that could, I think, have unintended consequences.

Also, this seems like a more or less unrelated change, so please do it in a separate commit with proper justification in the commit message, so that we can later easily find out why the change was made. In other words, try to follow the "1 logical change per patch" rule, as enforced e.g. by the kernel (sections 2) and 3) of this document could be helpful, though you don't need to follow all of it: https://www.kernel.org/doc/html/latest/process/submitting-patches.html#describe-your-changes)

TomasKorbar: Done

olysonek: Well, that's not what I meant :). Now it'll fail if the object is a `str` object. It should accept both `str` and `unicode`.

olysonek: Use `assert (isinstance(device,str) or isinstance(device,unicode))`, in this order, because `unicode` is undefined in python3.

olysonek: This would actually be a fix for commit f563c7d756cc92. Please mention that in the commit message.

TomasKorbar: Ah, sorry. Changed and tested on python2/3.

yarda:
Thanks, merged.

PR #142, opened 2018-12-12
Fix unit configuration reading

TomasKorbar:
- 1 is not equal to "1" in python
- Resolves: rhbz#1613379

### DIFF
--- a/tuned/profiles/unit.py
+++ b/tuned/profiles/unit.py
@@ -11,8 +11,8 @@ class Unit(object):
 	def __init__(self, name, config):
 		self._name = name
 		self._type = config.pop("type", self._name)
-		self._enabled = config.pop("enabled", True) in [True, "true", 1]
-		self._replace = config.pop("replace", False) in [True, "true", 1]
+		self._enabled = config.pop("enabled", True) in [True, "true", 1, "1"]
+		self._replace = config.pop("replace", False) in [True, "true", 1, "1"]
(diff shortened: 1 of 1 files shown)


tuned/profiles/unit.py
```diff
@@ -11,8 +11,8 @@ class Unit(object):
 	def __init__(self, name, config):
 		self._name = name
 		self._type = config.pop("type", self._name)
-		self._enabled = config.pop("enabled", True) in [True, "true", 1]
-		self._replace = config.pop("replace", False) in [True, "true", 1]
+		self._enabled = config.pop("enabled", True) in [True, "true", "1"]
+		self._replace = config.pop("replace", False) in [True, "true", "1"]
```
olysonek: Personally, I would append "1" instead of replacing the 1, to be safe. Are you sure there's no way the value will be an integer?

TomasKorbar: I tested this with both python2 and 3 but the value got by this approach was always string, so i think it is safe but to be sure we can just append "1".

olysonek: Well there's also a boolean in the list of possible values. Perhaps there is a reason for each value. Keep in mind that the code is called from multiple places. So I'd prefer to just append the "1".

TomasKorbar: Ok. Solved.

yarda:
Thanks, merged.

PR #144, opened 2018-12-14
scheduler: Keep polling file objects alive long enough

olysonek:
Make sure the file objects returned by evlist.get_pollfd()
don't go out of scope and get destroyed too soon. This is
a workaround for python3-perf rhbz#1659445.

Resolves: rhbz#1659140


yarda:
Thanks, merged.

PR #151, opened 2019-01-25
Differ archs and x86 CPU vendors in plugin_cpu

watologo1:
This avoids wrong warning about a missing intel tool (x86_energy_perf_bias)

Remember arch and x86 CPU vendor to allow arch/vendor specific CPU tunings
in the future


yarda:
LGTM, thanks merged.

PR #153, opened 2019-01-30
Added profile for HPC compute workloads

nealepetrillo:
Added a profile for common HPC workloads. It sets basic VM policies common to compute nodes and enables a few network settings that optimize for large, contiguous, file operations.

### DIFF
--- a/profiles/hpc-compute/tuned.conf
+++ b/profiles/hpc-compute/tuned.conf
@@ -4,9 +4,25 @@
 
 [main]
 summary=Optimize for HPC compute workloads
-include=throughput-performance
+description=Configures virtual memory, CPU governors, and network settings for HPC compute workloads.
+include=latency-performance
+
+[vm]
+transparent_hugepages=always
+
+[disk]
+readahead=>4096
 
 [sysctl]
+# Forces hugepages to be allocated on non-hotpluggable memory
+vm.hugepages_treat_as_movable=0
+vm.min_free_kbytes=135168
+vm.zone_reclaim_mode=1
+
+# Most HPC applications have NUMA knowledge. Disabling automatic NUMA
+# balancing prevents unwanted memory unmapping. 
+kernel.numa_balancing=0
+
 # Busy polling helps reduce latency in the network receive path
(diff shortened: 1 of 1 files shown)


yarda:
Thanks for contribution. Do you have any HPC benchmark results w/wo these tuning applied?

yarda:
And isn't better to base this profile on 'throughput-performance'? E.g. with no C state locking which can negatively affect cluster power consumption.

nealepetrillo:
To benchmark these profiles, we chose to run the Intel Message Passing Benchmark (IMB) to assess MPI communication performance, the Interleaved or Random (IOR) benchmark to assess network IO performance, and the High Performance LINPACK (HPL) benchmark to assess compute performance. We used the throughput-performance, latency-perfromance, and this tuning profile. 

MPI performance was sporadic regardless of tuning profile. The Exchange benchmark showed a notable improvement of about 5% in throughput using this tuning profile however, latency suffered about a 3% decrease compared to either the latency-perfromance or throughput-performance profiles. Both of these numbers were within the error region. 

HPL performance increased about 2% using this profile compared to throughput-performance and about 5% compared to latency-performance. This was, again, in the error region. 

IOR was much more consistent with error rates in the 1% region. The throughput-performance profile showed the least performance. The latency-performance profile showed a 3% increase in read throughput and a 1.5% increase in write throughput compared to throughput-performance. This profile showed a 2.5% increase in read performance and a 2.5% increase in write performance over the throughput-performance profile. 

We did not consider power consumption as an important factor although I realize that this is of concern to some other organizations.

yarda:
Thanks, for info. It's under review.

yarda:
One minor redundancy spotted - you are setting
`kernel.sched_migration_cost_ns=5000000`
But it's already set in the latency-performance profile you are including. Could you please update the PR and drop it? Thanks.

nealepetrillo:
Removed extra `kernel.sched_migration_cost_ns=5000000` with commit 8f64333

yarda:
Thanks, merged.

PR #157, opened 2019-02-08
Increase max log file size in the default config file

olysonek:
100KB is not a lot, it can fill up quickly.


yarda:
Thanks, merged.

PR #158, opened 2019-02-13
conservative governor may be compiled as module and needs explicit lo…

watologo1:
…ading

Tested on an AMD machine with acpi-cpufreq kernel driver providing
ondemand, conservative,... governors.

Without this line, one gets:
tuned.plugins.plugin_cpu: ignoring governor 'conservative' on cpu 'cpu7', it
is not supported

when switching to balanced profile.
With this line you see:
tuned.plugins.plugin_cpu: setting governor 'ondemand' on cpu 'cpu2'

Be careful: Order matters!
tuned seem to read profile config file line by line or section by section
and directly tries to apply things.

If order of these lines are exchanged:
[modules]
cpufreq_conservative=+r

[cpu]
governor=conservative

you fall back to previous error again. Therefore the newly introduced
[modules] section is not put at the end, but nearly at top.

### DIFF
--- a/profiles/balanced/tuned.conf
+++ b/profiles/balanced/tuned.conf
@@ -5,6 +5,9 @@
 [main]
 summary=General non-specialized tuned profile
 
+[modules]
+cpufreq_conservative=+r
+
 [cpu]
 governor=conservative
 energy_perf_bias=normal


yarda:
Thanks for contribution.

Regarding the plugin instances priority it's not yet perfect. Currently, it uses dict and the order depends on python internals, so the order cannot be relied on. It could be reworked. But there is simple hack you could use for it, add option 'priority', e.g.:

[cpu]
priority=10
governor=conservative

Every instance has priority which is by default 0 and the plugin instances are processed from the lowest number (meaning highest priority) to higher numbers. The priority can be also negative.

Could you please update your PR?

yarda:
NP, I will update it myself. Thanks for the contribution, merged.

PR #159, opened 2019-02-19
Makefile: Drop a workaround for a pylint bug and run pylint on tests too

olysonek:
The bug is no longer reproducible with
python3-pylint-2.1.1-2.fc29.noarch.


yarda:
Thanks, merged.

PR #160, opened 2019-02-25
Apply sysctl settings by writing to /proc/sys directly

### LINKED ISSUE #146: Tuned takes too long to reload/start when "ulimit -n" is high.
When `ulimit -n` is `1048576`, tuned takes around ~20s to reload due to subprocess.Popen(... close_fds = True ...) setting all the "sysctls" via the call.

Do we need `close_fds=True`?

Links:
https://github.com/redhat-performance/tuned/blob/master/tuned/utils/commands.py#L230
https://www.python.org/dev/peps/pep-0446/#closing-all-open-file-descriptors
https://bugs.python.org/issue1663329

Workaround: set `ulimit -n 1024` or `close_fds=False`.

/cc @jeremyeder

olysonek:
Fixes #146
Resolves: rhbz#1663412

### DIFF
--- a/tuned/plugins/plugin_sysctl.py
+++ b/tuned/plugins/plugin_sysctl.py
@@ -5,8 +5,12 @@ import tuned.logs
 from subprocess import *
 from tuned.utils.commands import commands
 import tuned.consts as consts
+import errno
+import os
 
 log = tuned.logs.get()
+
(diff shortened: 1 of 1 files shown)


yarda:
OK, thanks. Regarding the bound methods -> functions, it depends whether we want Tuned to be strictly object code or not, but personally I don't care.

PR #163, opened 2019-03-05
Update tuned.conf(5) and powertop2tuned to use ${i:PROFILE_DIR}

olysonek:


yarda:
Thanks, LGTM, merged.

PR #164, opened 2019-03-05
Fix checking for None values

olysonek:
In python, 'is None' should be used instead of '== None' to check if
something is None.


yarda:
LGTM, but there is conflict, could you please update the PR?

PR #168, opened 2019-03-07
Fix issues from pullrequest 138

TomasKorbar:
Hi,
In pullrequest 138 i overlooked some issues and this pullrequest is meant to fix them.
Please pay attention to code which replaced list profiles and plugins subcommands in tuned-adm. I would like to know your opinion on this. Maybe its little bit hacky but this way we can achieve the same behaviour like with subcommands in python3.

### DIFF
--- a/tuned/gtk/gui_profile_loader.py
+++ b/tuned/gtk/gui_profile_loader.py
@@ -91,7 +91,7 @@ class GuiProfileLoader(object):
     def _load_all_profiles(self):
         for d in self.directories:
             for profile in os.listdir(d):
-                if os.path.isdir(d + '/' + profile):
+                if self._is_dir_profile(os.path.join(d, profile)):
                     try:
                         self.profiles[profile] = p.Profile(profile,
                                 self.load_profile_config(profile, d))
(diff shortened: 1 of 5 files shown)


olysonek:
Thank you for doing this, I'll take a look asap.

TomasKorbar:
Done. I hope it is clear.

tuned/gtk/gui_profile_loader.py
```diff
@@ -103,6 +103,11 @@ def _load_all_profiles(self):
 #                         raise managerException.ManagerException("Can not make profile")
 #                         print "can not make \""+ profile +"\" profile without correct config with path: " + d
 
+    def _is_dir_profile(self, path):
+        is_dir = os.path.isdir(path)
```
olysonek: You can (and should, to avoid unnecessary execution) immediatelly return False if it's not a directory.

TomasKorbar:
> "new command which haves" -> "which has" ;)
... I dont really know how did this happen. Fixed.

tuned/gtk/gui_profile_loader.py
TomasKorbar: Changed.

TomasKorbar:
> So the methods `_action_dbus_list_plugins`, `_action_list_plugins`, `_action_dbus_list_profiles` and `_action_list_profiles` can no longer be reached using the mechanism in the `action` method, right? It would probably be more readable if you renamed them to something like `list_plugins` and `list_plugins_dbus`, but it's not important. Let's keep it as it is for now.

Well, they can be reached by action method because they have `_action` prefix. But user can not use them directly without passing through `_action_dbus_list` or `_action_list`. Now as you say it i think action prefix should be removed from `_action_dbus_list_plugins`, `_action_list_plugins`, `_action_dbus_list_profiles` and `_action_list_profiles` to ensure clarity of code.

olysonek:
Looks good to me, thank you.

yarda:
Thanks, merged.

PR #171, opened 2019-03-13
Improve logging of tuned-adm recommend

TomasKorbar:
Hi,
recently we've noticed mutliple error messages warning user about running virt-what without root privileges. This pull request makes from these multiple messages one which says exactly what is wrong.
Problem occured when executing:
$ tuned-adm recommend

### DIFF
--- a/tuned/utils/profile_recommender.py
+++ b/tuned/utils/profile_recommender.py
@@ -32,7 +32,11 @@ class ProfileRecommender:
 		profile = consts.DEFAULT_PROFILE
 		if hardcoded:
 			return profile
-		matching = self.process_config(consts.RECOMMEND_CONF_FILE)
+		has_root = os.geteuid() == 0
+		if not has_root:
+			log.warning("Profile recommender is running without root privileges. Profiles with virt recommendation condition will be omitted.")
+		matching = self.process_config(consts.RECOMMEND_CONF_FILE,
+									   has_root=has_root)
 		if matching is not None:
 			return matching
 		files = {}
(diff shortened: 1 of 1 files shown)


tuned/utils/profile_recommender.py
```diff
@@ -32,7 +32,13 @@ def recommend(self, hardcoded = False):
 		profile = consts.DEFAULT_PROFILE
 		if hardcoded:
 			return profile
-		matching = self.process_config(consts.RECOMMEND_CONF_FILE)
+		has_root = os.getuid() == 0
```
olysonek: This should use `os.geteuid()`. The effective UID matters, not the real one.

tuned/utils/profile_recommender.py
```diff
@@ -32,7 +32,13 @@ def recommend(self, hardcoded = False):
...
+		if not has_root:
+			log.warning(("Profile recommender is running without root"
+						 " privileges. Profiles with virt recommendation"
+						 " condition will be omitted."))
```
olysonek: I prefer to put log messages on a single line, even if it makes the line too long. It makes it easier to `grep` for the message.

tuned/utils/profile_recommender.py
olysonek: @yarda Can you please fix that up once you merge this? I fixed one more new occurence of the problem here https://github.com/redhat-performance/tuned/pull/172

TomasKorbar: Fixed.

tuned/utils/profile_recommender.py
TomasKorbar: Fixed

olysonek:
Thanks, LGTM.

yarda:
Thanks, merged.

PR #172, opened 2019-03-21
Check for root privileges by inspecting the effective UID

olysonek:
The effective UID is what matters when it comes to the ability to
perform privileged actions, not the real UID.


olysonek:
This fixes a newly added problem, but it's a very minor one, so we don't have to include it 2.11.

yarda:
OK, I finally merged it to 2.11, it's related. Thanks.

PR #177, opened 2019-04-25
virtual-host: Fix setting force_latency

olysonek:
The force_latency parameter belongs to the cpu plugin, so that's the
section it needs to be in.

Fixes #132
Resolves: rhbz#1569375


yarda:
Thanks, merged.

PR #178, opened 2019-04-30
[Docs] Improved the description of the script plug-in

mrksu:
This PR fixes https://bugzilla.redhat.com/show_bug.cgi?id=1671107

### DIFF
--- a/doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
+++ b/doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
@@ -85,15 +85,32 @@ It sets Aggressive Link Power Management (ALPM) to the value specified by the [o
 Enables or disables barriers for mounts according to the Boolean value of the [option]`disable_barriers` option. 
 
 `script`::
-Can be used for the execution of an external script that is run when the profile is loaded or unloaded.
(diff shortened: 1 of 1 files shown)


mrksu:
Sorry about the force pushes on my branches. I added more commits and squashed them afterwards -- nothing scary happening.

olysonek:
No problem, that is the correct way to do it. The changes look good. It's much better now, thanks! I'll let Jarda merge it.

yarda:
@mrksu please update the script description, that it can be any executable/script, e.g. Bash script

olysonek:
You mean this, right? "Can be used for the execution of an external Bash script"
Agreed.

(I thought that there was some magic happening inside the `process` function in the Bash library, but I was wrong.)

olysonek:
Well, there is no magic, however importing the Bash library does bring some useful variables, such as `STORAGE_PERSISTENT`. That's why I think Bash scripts are preferable...

mrksu:
OK, I've generalized the text to say that you can run an "executable", which can be either a "script" or a "binary". The parts about Bash scripts specifically say Bash.

yarda:
Thanks, merged.

Issue #180, opened 2019-05-02
tuned polkit rules must be more restrictive - ordinary X users must not be able to start/stop tuned


watologo1:
These should all be auth_admin by default, X users must not be able to disable/start/stop/reload the systemd tune daemon?

com.redhat.tuned.disable                        auth_admin:auth_admin:yes
com.redhat.tuned.reload                         auth_admin:auth_admin:yes
com.redhat.tuned.start                          auth_admin:auth_admin:yes
com.redhat.tuned.stop                           auth_admin:auth_admin:yes

Also security people at SUSE have problems with the newly introduced:
get_all_plugins and friends which ends up in load_all_plugins() and which is suggested to be executable/triggerable by all users yes:yes:yes.
Ideally this should be auth_admin as well.

yarda:
All of them currently have:
```
      <allow_any>auth_admin</allow_any>
      <allow_inactive>auth_admin</allow_inactive>
      <allow_active>yes</allow_active>
```
Which IMHO allows only people logged directly into the machine on a TTY or X display (i.e. mostly people with physical access to the machine) without auth. It's same for many other policies. E.g. if the user is already allowed to reboot or power-off the machine, I think that changing the performance profile cannot cause more harm, e.g. on Fedora:
```
$ pkaction -v --action-id org.freedesktop.login1.reboot
org.freedesktop.login1.reboot:
  description:       Reboot the system
  message:           Authentication is required for rebooting the system.
  vendor:            The systemd Project
  vendor_url:        http://www.freedesktop.org/wiki/Software/systemd
  icon:              
  implicit any:      auth_admin_keep
  implicit inactive: auth_admin_keep
  implicit active:   yes
  annotation:        org.freedesktop.policykit.imply -> org.freedesktop.login1.set-wall-message

$ pkaction -v --action-id org.freedesktop.login1.power-off
org.freedesktop.login1.power-off:
  description:       Power off the system
  message:           Authentication is required for powering off the system.
  vendor:            The systemd Project
  vendor_url:        http://www.freedesktop.org/wiki/Software/systemd
  icon:              
  implicit any:      auth_admin_keep
  implicit inactive: auth_admin_keep
  implicit active:   yes
  annotation:        org.freedesktop.policykit.imply -> org.freedesktop.login1.set-wall-message
```
Just my two cents - I am open to discussion.

[...]

PR #181, opened 2019-05-03
Lower permissions for plugin info dbus interfaces

watologo1:
As these interfaces allow loading and thus executing code from plugins
the default permission policy is lowered from:
yes:yes:yes
to
auth_admin:auth_admin:yes
So only users logged into the X server are allowed to use this interface
without extra root/admin authentication.

This was found by security review by Matthias Gerstner
<matthias.gerstner@suse.com>

Also see:
https://github.com/redhat-performance/tuned/issues/180


olysonek:
Looks good to me.

yarda:
LGTM, merged, thanks.

PR #183, opened 2019-05-09
functions: Return an ordered cpu list in cpulist_online

olysonek:
cpulist_unpack, which is used in the function, returns an ordered cpu
list, however by converting it to a set, the order is lost. Rewrite the
operation in a way that preserves the order.

Resolves: rhbz#1706171


yarda:
Thanks, merged.

PR #184, opened 2019-05-14
Fix verifying sysctl options with tabs

kbotc:
sysctl options such as `net.ipv4.tcp_wmem` and `net.ipv4.tcp_rmem` include tabs. When you use `tuned-adm verify`, the changes report back as broken as the whitespaces are different because one side has been sanitized but the other side has not. This pull request will fix that.


olysonek:
Thanks! Looks good to me. I'll let @yarda merge it.

Btw, that was me who broke this (commit 3166940015), sorry! I wish we did test-driven development - this would have been easily caught by a unit test.

yarda:
@kbotc thanks, good catch, merged. I wonder why it has been caught by our QA :)

PR #186, opened 2019-05-28
Create log file directory with more limited permissions

watologo1:
This patch is originally from:
Matthias Gerstner <matthias.gerstner@suse.com>
If tuned creates /var/log/tuned
permissions are now 750 instead of 755

### DIFF
--- a/tuned/logs.py
+++ b/tuned/logs.py
@@ -134,6 +134,7 @@ class TunedLogger(logging.getLoggerClass()):
 			log_directory = '.'
 		if not os.path.exists(log_directory):
 			os.makedirs(log_directory)
+			os.chmod(log_directory, 0o750)
 
 		cls._file_handler = logging.handlers.RotatingFileHandler(
 			filename, maxBytes = int(maxBytes), backupCount = int(backupCount))


olysonek:
First of all, are you sure the old permissions are 777? I got 775 in my testing.

Second, this change will need to be justified in the commit message. Why do you think the permissions should be limited? Namely, why should not the log directory be accessible to others?

olysonek:
> First of all, are you sure the old permissions are 777? I got 775 in my testing.

I even get 755 if I actually run Tuned instead of just `os.makedirs` in the interpreter. Are you sure this is not a more general problem in your environment, not specific to Tuned? Don't get me wrong, I'm definitely not against specifying the permissions explicitly.

There's also a small race in the proposed patch between creating the directory and changing the permissions. There must be an atomical way to do it.

And finally, if we're changing the permissions of the log directory, we should probably also change the permissions of the log files themselves.

Anyway, thanks for bringing this up! I have myself been wondering whether the log files should be world-readable, however right now I can't think of a reason why they shouldn't be.

watologo1:
@olysonek Yes, sorry for the quickly written, wrong comment. I could edit this on the fly and corrected it to: permissions are now 750 instead of 755

> There's also a small window for a race in the proposed patch between creating the directory and
> changing the permissions. There must be an atomical way to do it.
This certainly is a minor security issue, as no data is written to the directory yet at that point...
But yes, os.makedirs also takes a permission parameter which is much nicer.
I opened a new pull request and will close this one:
https://github.com/redhat-performance/tuned/pull/188

PR #187, opened 2019-05-29
Fix a bare 'except' statement

olysonek:
A bare 'except' statement can catch (and hide) all sorts of errors,
including NameError. Fix it to catch only the errors we really expect.


yarda:
Thanks, merged.

PR #188, opened 2019-05-29
Create log file directory not world readable

watologo1:
If tuned creates /var/log/tuned
permissions are now 750 instead of 755

CC: Matthias Gerstner <matthias.gerstner@suse.com>

### DIFF
--- a/tuned/logs.py
+++ b/tuned/logs.py
@@ -133,7 +133,7 @@ class TunedLogger(logging.getLoggerClass()):
 		if log_directory == '':
 			log_directory = '.'
 		if not os.path.exists(log_directory):
-			os.makedirs(log_directory)
+			os.makedirs(log_directory, 0o750)
 
 		cls._file_handler = logging.handlers.RotatingFileHandler(
 			filename, maxBytes = int(maxBytes), backupCount = int(backupCount))


olysonek:
I'm still missing a justification why the directory should not be world-readable. This will hurt user experience, so it needs to be well justified.

watologo1:
I tried to find a hint on FHS (Filesystem Hierarchy Standard), but I couldn't find a hint about permissions for this directory.
IMO it makes a lot of sense to remove the world readable/access bits on systemd service logfiles.
Matthias Gerstner who suggested this, arguemented (I'll add him):
> We try to harden our distributions with regards to log file permissions. In short, we want to avoid
> world readable log files, because they might at some point contain sensitive data that could facilitate
> other types of attacks.

mgerstner:
Basically this is a hardening measure we're currently following up in SUSE distributions. Trigger for this was an [older security finding](https://seclists.org/oss-sec/2015/q2/758), where passwords got exposed in log files.

I don't know whether Fedora/Red Hat products by know implement the suggested "secure by default config and log file permissions" policy refered to in above post. Maybe it's something not for upstream but only for distro adjustment. In the latter case it would help if the used permissions could be a compiler or config file setting that can easily be adjusted in packaging.

PR #189, opened 2019-05-29
sysctl: Ignore non-existent settings from system sysctl configs

olysonek:
Ignore non-existent sysctl settings from the system configuration files
(/etc/sysctl.conf, etc.). Logging errors about these settings hurts user
experience, if the non-existent settings are in fact real settings that
are just temporarily unavailable. For example, the following settings
(from /usr/lib/sysctl.d/00-system.conf on RHEL-7) are not available until
the br_netfilter module is loaded. However once that module is loaded,
it is often desirable to set these, so having them in a RHEL-provided
configuration file makes sense.

net.bridge.bridge-nf-call-ip6tables = 0
net.bridge.bridge-nf-call-iptables = 0
net.bridge.bridge-nf-call-arptables = 0

This change restores the old behaviour before the recent rewrite of the
sysctl plugin away from using the 'sysctl' program (running
'sysctl --system' ignores missing settings).

Resolves: rhbz#1714595

### DIFF
--- a/tuned/plugins/plugin_sysctl.py
+++ b/tuned/plugins/plugin_sysctl.py
@@ -133,7 +133,7 @@ def _apply_sysctl_config_line(path, lineno, line):
 				% (path, lineno))
 		return
 	value = value.strip()
-	_write_sysctl(option, value)
+	_write_sysctl(option, value, ignore_missing = True)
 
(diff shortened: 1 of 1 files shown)


yarda:
Thanks, LGTM, merged.

PR #190, opened 2019-05-30
Drop sap-hana-vmware profile, fix other SAP profiles

olysonek:
Please don't merge yet.

Resolves: rhbz#1708418
Resolves: rhbz#1701394


olysonek:
Resolves: rhbz#1715541

olysonek:
> Please don't merge yet.

I marked this PR as work-in-progress because I'm not sure whether we should change the value of `kernel.sem` in the `sap-hana` profile - I'm waiting for response to https://bugzilla.redhat.com/show_bug.cgi?id=1701394#c10. However I think it's likely that the change is supposed to be done in `sap-hana`, so I added the change to this PR. We can revert it later, if needed, however I don't think that it will be needed.

I propose we merge this now.

olysonek:
I've received confirmation that the change also applies to `sap-hana`, so this PR can be merged as it is. https://bugzilla.redhat.com/show_bug.cgi?id=1701394#c13

yarda:
LGTM, thanks, merged.

Issue #192, opened 2019-06-28
Avoid printing ERROR logs even if they are not really errors


xzpeter:
Some log examples on my laptop:

2019-06-26 18:08:19,339 ERROR    tuned.utils.commands: Executing modinfo error: modinfo: ERROR: Module cpufreq_conservative not found.
2019-06-26 18:08:19,339 ERROR    tuned.plugins.plugin_modules: kernel module 'cpufreq_conservative' not found, skipping it

or:

2019-06-26 18:08:19,341 ERROR    tuned.utils.commands: Error when reading file '/sys/class/drm/card0/device/power_method': '[Errno 2] No such file or directory: '/sys/class/drm/card0/device/power_method''

Some more from some RT servers:

2019-06-26 02:15:16,376 ERROR    tuned.plugins.plugin_sysctl: Failed to read sysctl parameter 'kernel.numa_balancing', the parameter does not exist
2019-06-26 02:15:16,376 ERROR    tuned.plugins.plugin_sysctl: sysctl option kernel.numa_balancing will not be set, failed to read the original value.

or:

2019-06-26 03:01:16,847 ERROR    tuned.plugins.plugin_scheduler: Failed to set affinity of PID 4084 to '[0, 2, 4, 6, 8, 10]': [Errno 22] Invalid argument

AFAICT all these ERRORs that I pasted are not vital or even many of them are just informative.  Would it make sense that we change these ERRORs into WARNINGs or even INFOs (we obviously know we can skip cpufreq_conservative)?  These could help people who is not yet familiar with tuned to better understand what has really went wrong, instead of thinking that tuned was not working properly.  Also, people can grep ERROR for real errors.

yarda:
Thanks for the report. I agree with your comments. This is something we are trying to gradually improve. Feel free to provide PR.

PR #193, opened 2019-07-03
misc: Lower logging level for some errors

xzpeter:
There're some errors that do not affect tuned from running, even some
of them are quite natural and all these ERRORs are misleading to
newcomers of tuned when they read the logs.  Examples:

https://github.com/redhat-performance/tuned/issues/192

This patch tries to lower quite a few of these cases to use either
warn() or even info() for these errors so the real errors could be
more obvious to users.

One case in _get_radeon_powersave() is special, I simply checked the
file existance before read_file().

Signed-off-by: Peter Xu <peterx@redhat.com>

### DIFF
--- a/tuned/plugins/plugin_cpu.py
+++ b/tuned/plugins/plugin_cpu.py
@@ -133,7 +133,7 @@ class CPULatencyPlugin(base.Plugin):
 			try:
 				self._cpu_latency_fd = os.open(consts.PATH_CPU_DMA_LATENCY, os.O_WRONLY)
 			except OSError:
-				log.error("Unable to open '%s', disabling PM_QoS control" % consts.PATH_CPU_DMA_LATENCY)
+				log.info("Unable to open '%s', disabling PM_QoS control" % consts.PATH_CPU_DMA_LATENCY)
 				self._has_pm_qos = False
 			self._latency = None
 
--- a/tuned/plugins/plugin_modules.py
+++ b/tuned/plugins/plugin_modules.py
@@ -49,12 +49,15 @@ class ModulesPlugin(base.Plugin):
 			module = self._variables.expand(option)
 			v = self._variables.expand(value)
 			if not skip_check:
-				retcode, out = self._cmd.execute(["modinfo", module])
+				# modinfo could fail as expected when module is not
+				# there. Capture the error instead of printing an
+				# ERROR every time even it's a very valid case
+				retcode, out, err = self._cmd.execute(["modinfo", module], return_err=True)
 				if retcode < 0:
 					skip_check = True
-					log.warn("'modinfo' command not found, not checking kernel modules")
+					log.info("'modinfo' command not found, not checking kernel modules")
 				elif retcode > 0:
-					log.error("kernel module '%s' not found, skipping it" % module)
+					log.info("kernel module '%s' not found, skipping it" % module)
 			if skip_check or retcode == 0:
 				if len(v) > 1 and v[0:2] == "+r":
 					v = re.sub(r"^\s*\+r\s*,?\s*", "", v)
--- a/tuned/plugins/plugin_scheduler.py
+++ b/tuned/plugins/plugin_scheduler.py
@@ -318,15 +318,15 @@ class SchedulerPlugin(base.Plugin):
 	def _convert_sched_params(self, str_scheduler, str_priority):
 		scheduler = self._dict_schedcfg2num.get(str_scheduler)
 		if scheduler is None and str_scheduler != "*":
-			log.error("Invalid scheduler: %s. Scheduler and priority will be ignored."
-					% str_scheduler)
+			log.warn("Invalid scheduler: %s. Scheduler and priority will be ignored."
+					 % str_scheduler)
 			return (None, None)
 		else:
 			try:
 				priority = int(str_priority)
 			except ValueError:
-				log.error("Invalid priority: %s. Scheduler and priority will be ignored."
-							% str_priority)
+				log.warn("Invalid priority: %s. Scheduler and priority will be ignored."
+						 % str_priority)
 				return (None, None)
 		return (scheduler, priority)
 
@@ -336,8 +336,8 @@ class SchedulerPlugin(base.Plugin):
 		else:
 			affinity = self._cmd.hex2cpulist(str_affinity)
 			if not affinity:
-				log.error("Invalid affinity: %s. It will be ignored."
-						% str_affinity)
+				log.warn("Invalid affinity: %s. It will be ignored."
+						 % str_affinity)
 				affinity = None
 		return affinity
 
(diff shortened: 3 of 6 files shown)


tuned/plugins/plugin_modules.py
```diff
@@ -49,12 +49,15 @@ def _instance_apply_static(self, instance):
...
+				# there. Capture the error instead of printing an
+				# ERROR every time even it's a very valid case
+				retcode, out, err = self._cmd.execute(["modinfo", module], return_err=True)
 				if retcode < 0:
 					skip_check = True
-					log.warn("'modinfo' command not found, not checking kernel modules")
+					log.info("'modinfo' command not found, not checking kernel modules")
 				elif retcode > 0:
-					log.error("kernel module '%s' not found, skipping it" % module)
+					log.info("kernel module '%s' not found, skipping it" % module)
```
olysonek: If a kernel module is not found, it can be a real problem (e.g. the module name might be misspelled in the Tuned profile, or the admin forgot to build the required kernel module). So I don't think it makes sense to hide this message. It might make sense to change it to `log.warn`, though.

It might make sense to merge the message printed by `self._cmd.execute` and by the line above into one message, so that we don't print two messages regarding one problem unnecessarily. I think we could do something similar to the following here:
`log.warn("modinfo failed, skipping the module '%s': %s" % (module, err))`
so that the message becomes something like
`WARNING ... modinfo failed, skipping the module 'cpufreq_conservative': modinfo: ERROR: Module cpufreq_conservative not found.`

... but that's getting a bit long, isn't it?

Anyway, note that the error message
`Executing modinfo error: modinfo: ERROR: Module cpufreq_conservative not found.`
in issue #192 can IMO be considered misconfiguration in the `balanced` profile. We're going to fix the error another way - we'll load the `cpufreq_conservative` module in the `cpu` plugin, if necessary, instead of doing it through the `modules` plugin.

tuned/plugins/plugin_video.py
```diff
@@ -80,6 +80,8 @@ def _set_radeon_powersave(self, value, device, sim):
 	@command_get("radeon_powersave")
 	def _get_radeon_powersave(self, device, ignore_missing = False):
 		sys_files = self._radeon_powersave_files(device)
+		if not os.path.exists(sys_files["method"]):
+			return None
```
olysonek: The message regarding the missing `power_method` file could be made into a warning instead of an error, however I don't think it makes sense to drop it altogether. The user should be able to learn that Tuned failed to apply the profile, whatever the reason for that is.

Note that the real solution to make the error message go away is to override the Tuned profile and drop the `radeon_powersave` tuning in the overriden profile.

tuned/plugins/plugin_sysctl.py
```diff
@@ -51,8 +51,8 @@ def _instance_apply_static(self, instance):
 		for option, value in list(instance._sysctl.items()):
 			original_value = _read_sysctl(option)
 			if original_value is None:
-				log.error("sysctl option %s will not be set, failed to read the original value."
-						% option)
+				log.warn("sysctl option %s will not be set, failed to read the original value."
+						 % option)
```
olysonek: We'll probably solve this another way - we'll add a way to differentiate sysctl settings that are required vs. not required. Failures to set required settings will result in error messages, whereas failures to set non-required settings will result in warning messages (this was requested in https://bugzilla.redhat.com/show_bug.cgi?id=1666562#c4). Until that's implemented, I think it's safer to keep printing error messages.

I have a work-in-progress solution to that here: https://github.com/olysonek/tuned/commit/adc926b756761e0e6784bc51b017cb1112dffb63

You can override the Tuned profile and drop the missing sysctl settings to make the error message go away.

tuned/plugins/plugin_scheduler.py
```diff
@@ -318,15 +318,15 @@ def _tune_process(self, pid, cmd, sched, prio, affinity):
 	def _convert_sched_params(self, str_scheduler, str_priority):
 		scheduler = self._dict_schedcfg2num.get(str_scheduler)
 		if scheduler is None and str_scheduler != "*":
-			log.error("Invalid scheduler: %s. Scheduler and priority will be ignored."
-					% str_scheduler)
+			log.warn("Invalid scheduler: %s. Scheduler and priority will be ignored."
+					 % str_scheduler)
```
olysonek: I'd say error is the correct log level here, as it can only happen due to misconfiguration. All the valid scheduler names are defined in `_dict_schedcfg2num`. If anything else is used in the profile, it's misconfiguration.

tuned/plugins/plugin_scheduler.py
```diff
@@ -318,15 +318,15 @@ def _tune_process(self, pid, cmd, sched, prio, affinity):
...
 			return (None, None)
 		else:
 			try:
 				priority = int(str_priority)
 			except ValueError:
-				log.error("Invalid priority: %s. Scheduler and priority will be ignored."
-							% str_priority)
+				log.warn("Invalid priority: %s. Scheduler and priority will be ignored."
```
olysonek: Again, this can only happen if an non-integer priority is specified in the profile, so only due to misconfiguration. I think error level is appropriate here.

tuned/plugins/plugin_scheduler.py
```diff
@@ -336,8 +336,8 @@ def _convert_affinity(self, str_affinity):
 		else:
 			affinity = self._cmd.hex2cpulist(str_affinity)
 			if not affinity:
-				log.error("Invalid affinity: %s. It will be ignored."
-						% str_affinity)
+				log.warn("Invalid affinity: %s. It will be ignored."
+						 % str_affinity)
```
olysonek: This can also happen only due to misconfiguration, so I think error is appropriate here.

tuned/plugins/plugin_cpu.py
```diff
@@ -133,7 +133,7 @@ def _instance_init(self, instance):
 			try:
 				self._cpu_latency_fd = os.open(consts.PATH_CPU_DMA_LATENCY, os.O_WRONLY)
 			except OSError:
-				log.error("Unable to open '%s', disabling PM_QoS control" % consts.PATH_CPU_DMA_LATENCY)
+				log.info("Unable to open '%s', disabling PM_QoS control" % consts.PATH_CPU_DMA_LATENCY)
```
olysonek: If the file does not exist, we should consider PM_QoS control unsupported, so we should log it as a warning. If the file could not be opened for other reasons, error level seems appropriate. So you could add an if-else statement here and change the log level as appropriate. I don't think hiding the message with an info level makes sense.

A related change we might want to do here is open the file descriptor only when it's needed - I believe there are tunings supported by the plugin that don't require the file descriptor.

tuned/plugins/plugin_scheduler.py
```diff
@@ -584,8 +584,10 @@ def _set_irq_affinity(self, irq, affinity, restoring):
 				log.debug("Setting SMP affinity of IRQ %s is not supported"
 						% irq)
 			else:
-				log.error("Failed to set SMP affinity of IRQ %s to '%s': %s"
-						% (irq, affinity_hex, e))
+				# managed IRQs would have smp_affinity file as RO, so
+				# it could be fine
+				log.warn("Failed to set SMP affinity of IRQ %s to '%s': %s"
+						 % (irq, affinity_hex, e))
```
olysonek: I think I'll wait for a resolution in our email thread regarding whether Tuned should manage IRQ affinities at all before digging into this :).

PR #194, opened 2019-07-04
Apply irqbalance service tuning for realtime-virtual-{host|guest} profiles

xzpeter:
The two realtime profiles needs to do cpu isolation just like cpu-partitioning profiles.  Do proper configuration for irqbalance service for them just like the cpu-partioning profile otherwise irqbalance might collapse with tuned on configuring system IRQs.

### DIFF
--- a/profiles/cpu-partitioning/script.sh
+++ b/profiles/cpu-partitioning/script.sh
@@ -39,8 +39,7 @@ start() {
     mkdir -p "${TUNED_tmpdir}/usr/lib/dracut/hooks/pre-udev"
     cp /etc/systemd/system.conf "${TUNED_tmpdir}/etc/systemd/"
     cp 00-tuned-pre-udev.sh "${TUNED_tmpdir}/usr/lib/dracut/hooks/pre-udev/"
-    sed -i '/^IRQBALANCE_BANNED_CPUS=/d' /etc/sysconfig/irqbalance
-    echo "IRQBALANCE_BANNED_CPUS=$TUNED_isolated_cpumask" >>/etc/sysconfig/irqbalance
+    irqbalance_banned_cpus_setup "$TUNED_isolated_cpumask"
     setup_kvm_mod_low_latency
     disable_ksm
 
@@ -52,7 +51,7 @@ start() {
 stop() {
     if [ "$1" = "full_rollback" ]
     then
-        sed -i '/^IRQBALANCE_BANNED_CPUS=/d' /etc/sysconfig/irqbalance
+        irqbalance_banned_cpus_clear
         teardown_kvm_mod_low_latency
         enable_ksm
     fi
--- a/profiles/functions
+++ b/profiles/functions
@@ -473,6 +473,17 @@ restore_logs_syncing() {
 	mv -Z $RSYSLOG_SAVE $RSYSLOG_CFG || mv $RSYSLOG_SAVE $RSYSLOG_CFG
 }
 
+irqbalance_banned_cpus_clear() {
+    sed -i '/^IRQBALANCE_BANNED_CPUS=/d' /etc/sysconfig/irqbalance
+}
+
+irqbalance_banned_cpus_setup() {
+    irqbalance_banned_cpus_clear
+    if [ -n "$1" ]; then
+        echo "IRQBALANCE_BANNED_CPUS=$1" >> /etc/sysconfig/irqbalance
+    fi
+}
+
 #
 # HARDWARE SPECIFIC tuning
 #
--- a/profiles/realtime/script.sh
+++ b/profiles/realtime/script.sh
@@ -3,10 +3,12 @@
 . /usr/lib/tuned/functions
 
 start() {
+    irqbalance_banned_cpus_setup "$TUNED_isolated_cpumask"
     return 0
 }
 
 stop() {
+    irqbalance_banned_cpus_clear
     return 0
 }
 
--- a/profiles/realtime/tuned.conf
+++ b/profiles/realtime/tuned.conf
@@ -18,6 +18,7 @@ assert1=${f:assertion_non_equal:isolated_cores are set:${isolated_cores}:${isola
 # Non-isolated cores cpumask including offline cores
 not_isolated_cpumask = ${f:cpulist2hex_invert:${isolated_cores}}
 isolated_cores_expanded=${f:cpulist_unpack:${isolated_cores}}
+isolated_cpumask=${f:cpulist2hex:${isolated_cores_expanded}}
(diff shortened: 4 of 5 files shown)


xzpeter:
I tested it on my host to make sure the irqbalance configs are setup correct and also I'm sure the env is setup correctly for irqblance daemon.

profiles/cpu-partitioning/script.sh
```diff
@@ -39,8 +39,7 @@ start() {
     mkdir -p "${TUNED_tmpdir}/usr/lib/dracut/hooks/pre-udev"
     cp /etc/systemd/system.conf "${TUNED_tmpdir}/etc/systemd/"
     cp 00-tuned-pre-udev.sh "${TUNED_tmpdir}/usr/lib/dracut/hooks/pre-udev/"
-    sed -i '/^IRQBALANCE_BANNED_CPUS=/d' /etc/sysconfig/irqbalance
-    echo "IRQBALANCE_BANNED_CPUS=$TUNED_isolated_cpumask" >>/etc/sysconfig/irqbalance
+    irqbalance_banned_cpus_setup $TUNED_isolated_cpumask
```
olysonek: Off the top of my head, I don't know what the format of `$TUNED_isolated_cpumask` is, however I think it would be safer to enclose it in double quotes.

profiles/realtime-virtual-guest/script.sh
```diff
@@ -4,10 +4,12 @@
 
 start() {
     systemctl start rt-entsk
+    irqbalance_banned_cpus_setup $TUNED_isolated_cpumask
```
olysonek: Same here, enclose `$TUNED_isolated_cpumask` in double quotes, to be safe.

profiles/realtime-virtual-host/script.sh
```diff
@@ -98,12 +98,13 @@ start() {
         echo `cat $CACHE_VALUE_FILE` > $KVM_LAPIC_FILE
     fi
     systemctl start rt-entsk
-
+    irqbalance_banned_cpus_setup $TUNED_isolated_cpumask
```
olysonek: Same here, enclose `$TUNED_isolated_cpumask` in double quotes.

xzpeter: Sure thing, I'll re-push soon.  Thanks for the quick reviews!

xzpeter:
Hello, Ondřej,

Do you think this new version could be accepted?

My understanding is that we might need a better way to fix this at last. Say, we may not want tuned to operate on the irqs but instead we may want tuned to depend on irqbalance package and we tell irqbalance to not bind irqs upon real-time cpus using IRQBALANCE_BANNED_CPUS.  However I'm thinking whether we would allow this change to be in first so we avoid the spikes earlier even without that bigger change (no matter how we'd like to do with that finally).

Thanks,

olysonek:
> Do you think this new version could be accepted?

Yes, I think it's good to go. Thanks for the patch.

> My understanding is that we might need a better way to fix this at last. Say, we may not want tuned to operate on the irqs but instead we may want tuned to depend on irqbalance package and we tell irqbalance to not bind irqs upon real-time cpus using IRQBALANCE_BANNED_CPUS.

Sounds good to me.

> However I'm thinking whether we would allow this change to be in first so we avoid the spikes earlier even without that bigger change (no matter how we'd like to do with that finally).

I have no problem with that at all. The change is a good one.

I'll let @yarda review and merge this PR.

olysonek:
> On second thought, shouldn't `irqbalance_banned_cpus_{setup,clear}` be used in the underlying `realtime` profile instead of `realtime-virtual-{host,guest}`? I think the fix is applicable to the base `realtime` as well.

Just a hint on how to implement that: looking at the variable handling code WRT inheritance, it looks to me like it should be sufficient to define the `isolated_cpumask` variable in the `realtime` profile. So no changes to `realtime-virtual-{guest,host}` should be necessary.

xzpeter:
> > On second thought, shouldn't `irqbalance_banned_cpus_{setup,clear}` be used in the underlying `realtime` profile instead of `realtime-virtual-{host,guest}`? I think the fix is applicable to the base `realtime` as well.
> 
> Just a hint on how to implement that: looking at the variable handling code WRT inheritance, it looks to me like it should be sufficient to define the `isolated_cpumask` variable in the `realtime` profile. So no changes to `realtime-virtual-{guest,host}` should be necessary.

Hi, Ondřej,

Yes it seems to work.  Actually I thought about doing this before but I noticed that most of the RT documents will only document to update /etc/tuned/realtime-virtual-{host|guest}.conf so I thought it would not work if we do it in realtime profile.  But lately I tried your suggestion and it actually works nicely even if the isolated_cores parameters are configured in the realtime-virtual-{host|guest}.conf files.  I'll push another version soon. 

Thanks,

olysonek:
Yes - when the profiles are merged, the `include` option in the `[variables]` section from the `realtime` profile gets overwritten by the `include` option from the `realtime-virtual-{guest,host}` profile. So `/etc/tuned/realtime-variables.conf` is not loaded at all.

Thanks for the updated patch. However it seems that rollback of the `/etc/sysconfig/irqbalance` modifications doesn't work. Weird...

olysonek:
> Thanks for the updated patch. However it seems that rollback of the `/etc/sysconfig/irqbalance` modifications doesn't work. Weird...

To be clear, it doesn't work for the `realtime-virtual-{guest,host}` only.

olysonek:
The problem is here: https://github.com/redhat-performance/tuned/blob/b514cda168878c330b70416676d6f692fb88d02a/tuned/plugins/plugin_script.py#L50

In my testing, `/etc/tuned/realtime-virtual-guest/script.sh` exits with exit code 5, so `/etc/tuned/realtime/script.sh` is not executed at all. We should probably drop that return statement.

olysonek:
Can you please rebase on top of this branch? It contains a fix for the issue.
https://github.com/olysonek/tuned/tree/scripts_plugin_fix

xzpeter:
I still don't understand why it failed for you (because it worked here), but sure I can rebase to your branch and push again.

olysonek:
It's just a problem with my environment - apparently I don't have the necessary dependencies installed, so it fails with the following:
```
Failed to execute operation: No such file or directory
Failed to get unit file state for ksm.service: No such file or directory
Failed to execute operation: No such file or directory
Failed to get unit file state for ksmtuned.service: No such file or directory
Failed to stop rt-entsk.service: Unit rt-entsk.service not loaded.
```

My patch should make it more robust.

Thanks. I think this is good to go now.

yarda:
Thanks, merged.

olysonek:
For the record:
Resolves: rhbz#1720042

PR #196, opened 2019-09-05
Start using CircleCi

TomasKorbar:
Hi,
as we discussed Travis is not suitable for multiple checks which we need. CircleCI will provide us easily extensible platform for future testing.
Feel free to propose any change in configuration file.


.circleci/config.yml
```diff
@@ -0,0 +1,77 @@
...
+            pip install pylint
+      - run:
+          name: download tool for comparing of pylints outputs
+          command: |
+            git clone https://github.com/TomasKorbar/pylint_regexer ./regexer
+      - run:
+          name: check current code with pylint
+          command: |
+            . venv/bin/activate
+            pylint -E --output-format=parseable -s n --exit-zero ./tuned/ > /tmp/new_pl_output
```
olysonek: This checks only the `tuned` module, right? Ideally, it should check also the python scripts in the top level directory and tests. Take a look at the `lint` target in the Makefile.

TomasKorbar:
Commits squashed. We are good to go.

yarda:
Thanks, but we finally ended on Packit (well, it currently doesn't work, but this is different story :), so I think this should be stripped down just to Travis removal.

olysonek:
Hm, I completely forgot that we currently have a travis config in `master`. It doesn't seem to run for pull requests, though..

Remember that we don't currently run unit tests in Packit, and I'm not sure Packit is a good solution for that. Perhaps we might like to keep Travis..

yarda:
> Hm, I completely forgot that we currently have a travis config in `master`. It doesn't seem to run for pull requests, though..
> 
> Remember that we don't currently run unit tests in Packit, and I'm not sure Packit is a good solution for that. Perhaps we might like to keep Travis..

We run unit tests in Packit, they are run during the build in copr.

olysonek:
> We run unit tests in Packit, they are run during the build in copr.

I forgot about that. I'm ok with removing travis.

TomasKorbar:
I suppose we do not need the dockerfile in tests folder anymore too, am i right?

olysonek:
> I suppose we do not need the dockerfile in tests folder anymore too, am i right?

I guess not.

TomasKorbar:
Then this is ready to be merged.

yarda:
Thanks, merging.

PR #197, opened 2019-09-10
Do not rely on systemctl if it is failing

yrobla:
When running in a container or chroot, systemctl
is not working, returning a warning message and
stopping.
So all these checks for systemctl return unexpected
output, causing not desired rollbacks. So detect
if we are in chroot and just return no full
rollback is required for that case.

### DIFF
--- a/tuned/daemon/daemon.py
+++ b/tuned/daemon/daemon.py
@@ -122,11 +122,18 @@ class Daemon(object):
 		return errstr
 
 	def _full_rollback_required(self):
-		retcode, out = self._cmd.execute(["systemctl", "is-system-running"], no_errors = [0])
+		retcode, out, err_msg = self._cmd.execute(["systemctl", "is-system-running"], no_errors = [0], return_err = True)
 		if retcode < 0:
 			return False
+
+		# if there is some stderr and not stdout, it means systemctl runs in container
+		# always rollback in that case
+		if len(out) == 0 and len(err_msg) > 0:
+			return True
 		if out[:8] == "stopping":
 			return False
+		if out[:7] == "offline":
+			return True               
 		retcode, out = self._cmd.execute(["systemctl", "list-jobs"], no_errors = [0])
 		return re.search(r"\b(shutdown|reboot|halt|poweroff)\.target.*start", out) is None
 


yarda:
Thanks, maybe you could run the stat with the stderr redirected to the /dev/null, in such case it will not output any error in case it's running in the chroot without /proc mounted.

And sorry for delay, I was on vacation :)

yrobla:
ok ,done

yarda:
Thanks. But the blocker is that the code you wrote is in Bash, and it should be Python, so it will not work (thanks @olysonek for pointing this out :). Could you fix it?

yrobla:
Oh, you are right, sorry that i did the fix quickly. I'll redo asap...

tuned/daemon/daemon.py
```diff
@@ -122,6 +122,16 @@ def _notify_profile_changed(self, profile_names, result, errstr):
 		return errstr
 
 	def _full_rollback_required(self):
+                # if we are in chroot, we cannot use systemctl
+                retcode_root, out_root = self._cmd.execute(["stat", "-c", "%d:%i", "/"])
```
olysonek: Please use `os.stat()` (https://docs.python.org/2/library/os.html#files-and-directories). It is highly desirable to execute external programs as little as possible. And there's no need to execute the `stat` program here, as far as I can tell.

Also, notice that you're indenting with spaces, wheres the rest of the file is indented with tabs. It won't even bytecompile with python3.

yrobla: Ok so i can change the indentation, sorry that it is just related to my editor.
Regarding the stat, can you help me see how to get the right info from python os.stat? so i am basing the condition on device number and inodes, that i get with -c %d:%i flag is bash stat. I looked at the result of os.stat() but i cold not find the same results that i needed

yrobla: I actually could see there are other ways (the check for inode doesn't seem to pass in all cases). Like, for example: https://stackoverflow.com/questions/20010199/how-to-determine-if-a-process-runs-inside-lxc-docker
Do you prefer that alternative for checking if we are in a container, or do you hae some other suggestions that could be better?

olysonek:
I think the following would have the same effect, without adding the execution of the `systemctl status` beast:
```
diff --git a/tuned/daemon/daemon.py b/tuned/daemon/daemon.py
index fbf10c7..5748ca3 100644
--- a/tuned/daemon/daemon.py
+++ b/tuned/daemon/daemon.py
@@ -125,7 +125,7 @@ class Daemon(object):
 		retcode, out = self._cmd.execute(["systemctl", "is-system-running"], no_errors = [0])
 		if retcode < 0:
 			return False
-		if out[:8] == "stopping":
+		if out[:8] == "stopping" or out[:7] == "offline":
 			return False
 		retcode, out = self._cmd.execute(["systemctl", "list-jobs"], no_errors = [0])
 		return re.search(r"\b(shutdown|reboot|halt|poweroff)\.target.*start", out) is None
```

But let's take a step back. Can you help me understand what exactly you are trying to achieve and why? I think this change is going to affect everyone who runs tuned in a container (I think OpenShift does this) and I'm not sure it's desirable.

olysonek:
> But let's take a step back. Can you help me understand what exactly you are trying to achieve and why? I think this change is going to affect everyone who runs tuned in a container (I think OpenShift does this) and I'm not sure it's desirable.

To be clear, the change I'm refering to is that previously, the function would return True in a container (AFAICT), now it returns False.

yrobla:
So the problem that i have is that as the checks for a system stopped do not work on containers. So all the times that the host is rebooted a rollback of the changes is performed. While a rollback only needs to be done if tuned is stopped manually.
Openshift is using tuned, but they started to use the daemon in 4.2, so the problems were not experienced until that version

olysonek:
So if you wanted to rollback all the changes done by Tuned, how would you do that? I don't see a way to do that, given that `_full_rollback_required` will now always return False in a container. I mean, you could do hacks like apply an empty profile, or start Tuned outside a container and stop it immediatelly. But those are hacks :).

yrobla:
Could it be done by detecting some signal? The problem with containers, is that the current way of discriminating between stopping manually / stopping because system is down, is not valid...

olysonek:
I agree. I'm asking all these questions simply because the proposed changes seem to be an incomplete solution to the problem :).

@jmencak Can you comment on this PR? Is it something you'd like to see merged, or is it going to cause problems in the OpenShift tuned operator? Do you currently have some mechanism to tell tuned whether it should do full rollback or not?

yrobla:
> I think the following would have the same effect, without adding the execution of the `systemctl status` beast:
> 
> ```
> diff --git a/tuned/daemon/daemon.py b/tuned/daemon/daemon.py
> index fbf10c7..5748ca3 100644
> --- a/tuned/daemon/daemon.py
> +++ b/tuned/daemon/daemon.py
> @@ -125,7 +125,7 @@ class Daemon(object):
>  		retcode, out = self._cmd.execute(["systemctl", "is-system-running"], no_errors = [0])
>  		if retcode < 0:
>  			return False
> -		if out[:8] == "stopping":
> +		if out[:8] == "stopping" or out[:7] == "offline":
>  			return False
>  		retcode, out = self._cmd.execute(["systemctl", "list-jobs"], no_errors = [0])
>  		return re.search(r"\b(shutdown|reboot|halt|poweroff)\.target.*start", out) is None
> ```
> 
> But let's take a step back. Can you help me understand what exactly you are trying to achieve and why? I think this change is going to affect everyone who runs tuned in a container (I think OpenShift does this) and I'm not sure it's desirable.

About that ... i guess this "offline" result depends on the runtime you use. With podman and with cri-o i get different results. If i run systemctl is-system-running on containers with cri-o i get:

systemctl is-system-running
Failed to get D-Bus connection: Operation not permitted
[root@test-master-0 /]# echo $?
1
So the "offline" check won't work, but capturing the output will.

jmencak:
> @jmencak Can you comment on this PR? Is it something you'd like to see merged, or is it going to cause problems in the OpenShift tuned operator? Do you currently have some mechanism to tell tuned whether it should do full rollback or not?

Thanks for bringing me in!  Yes, this would break what we're trying do in OCP 4.2, rollback of node settings and this would basically render the new functionality introduced in 4.2 useless.  Yolanda is aiming at future support for the bootloader plugin, which we do not support in OCP now.  We currently have no way to tell tuned whether it should do a full rollback or not.  I didn't have enough time to investigate, but I believe there needs to be a way to tell the tuned daemon "do not roll back" + have the same functionality in the Node Tuning Operator (NTO) if the bootloader functionality is to be supported via the tuned daemon.  But take this with a pinch of salt, will try to revisit this asap.

yrobla:
> > @jmencak Can you comment on this PR? Is it something you'd like to see merged, or is it going to cause problems in the OpenShift tuned operator? Do you currently have some mechanism to tell tuned whether it should do full rollback or not?
> 
> Thanks for bringing me in! Yes, this would break what we're trying do in OCP 4.2, rollback of node settings and this would basically render the new functionality introduced in 4.2 useless. Yolanda is aiming at future support for the bootloader plugin, which we do not support in OCP now. We currently have no way to tell tuned whether it should do a full rollback or not. I didn't have enough time to investigate, but I believe there needs to be a way to tell the tuned daemon "do not roll back" + have the same functionality in the Node Tuning Operator (NTO) if the bootloader functionality is to be supported via the tuned daemon. But take this with a pinch of salt, will try to revisit this asap.

But see that you are doing a rollback now when system is shutting down, but relying on a broken code. The method is returning false not because you determined that it was intented, but due to a bug in systemctl check for containers...

olysonek:
> systemctl is-system-running
> Failed to get D-Bus connection: Operation not permitted
> [root@test-master-0 /]# echo $?
> 1
> So the "offline" check won't work, but capturing the output will.

Which output do you mean?

> But see that you are doing a rollback now when system is shutting down, but relying on a broken code. The method is returning false not because you determined that it was intented, but due to a bug in systemctl check for containers...

I agree that it's relying on broken code, however that doesn't mean we can change it if we know it's going to cause problems. Merging the change as it is now would cause an undesirable behaviour change with no way to get the old behaviour. A bit more thought will have to be put into this. At the moment I'm not sure how to solve it so that everyone is satisfied.

yrobla:
> > I think the following would have the same effect, without adding the execution of the `systemctl status` beast:
> > ```
> > diff --git a/tuned/daemon/daemon.py b/tuned/daemon/daemon.py
> > index fbf10c7..5748ca3 100644
> > --- a/tuned/daemon/daemon.py
> > +++ b/tuned/daemon/daemon.py
> > @@ -125,7 +125,7 @@ class Daemon(object):
> >  		retcode, out = self._cmd.execute(["systemctl", "is-system-running"], no_errors = [0])
> >  		if retcode < 0:
> >  			return False
> > -		if out[:8] == "stopping":
> > +		if out[:8] == "stopping" or out[:7] == "offline":
> >  			return False
> >  		retcode, out = self._cmd.execute(["systemctl", "list-jobs"], no_errors = [0])
> >  		return re.search(r"\b(shutdown|reboot|halt|poweroff)\.target.*start", out) is None
> > ```
> > 
> > 
> > But let's take a step back. Can you help me understand what exactly you are trying to achieve and why? I think this change is going to affect everyone who runs tuned in a container (I think OpenShift does this) and I'm not sure it's desirable.
> 
> About that ... i guess this "offline" result depends on the runtime you use. With podman and with cri-o i get different results. If i run systemctl is-system-running on containers with cri-o i get:
> 
> systemctl is-system-running
> Failed to get D-Bus connection: Operation not permitted
> [root@test-master-0 /]# echo $?
> 1
> So the "offline" check won't work, but capturing the output will.

I did same test on cri-o and podman. On cri-o, systemctl is-system-running shows : Failed to get D-Bus connection: Operation not permitted. See how i tested:
[root@test-master-0 tmp]# cat test-systemctl.json 
{
  "metadata": {
      "name": "tuned-test"
  },
  "image":{
      "image": "quay.io/yrobla/cluster-node-tuned"
  },
  "args": [
    "systemctl is-system-running && echo 'returns ok' || echo $?"
  ],
  "command": [
      "/bin/sh", "-c"
  ],
  "linux": {
  }
}

Result:
Failed to get D-Bus connection: Operation not permitted
1

Where if you run that with podman, the command does not fail , but gives "offline" as status:

[root@38d2a4ff67ad /]# systemctl is-system-running
offline
[root@38d2a4ff67ad /]# echo $?
1

They have in common the exit code, 1. Maybe we can rely on that?

olysonek:
> They have in common the exit code, 1. Maybe we can rely on that?

No, we can't. According to the systemctl man page, almost all system states are associated with a non-zero exit code, including `initializing` and `starting`.

yrobla:
> > They have in common the exit code, 1. Maybe we can rely on that?
> 
> No, we can't. According to the systemctl man page, almost all system states are associated with a non-zero exit code, including `initializing` and `starting`.

Given that different behaviour then, what can we do to detect it properly? Search for the two different outputs, the "offline" and the "D-Bus" ones?
I also tried to look at ways to detect if we run in container, but again, the behaviour between docker/podman/cri-o, is different.

olysonek:
I'm not sure. Searching for specific error messages does not seem like a good solution either. I'll try to look into it later.

olysonek:
So I've got an idea. I assume the error message "Failed to get D-Bus connection: Operation not permitted" is printed to stderr, is that correct? If that's the case, we could check that stdout is empty and/or stderr is non-empty.

But we still have to solve the OpenShift problem.

yrobla:
> So I've got an idea. I assume the error message "Failed to get D-Bus connection: Operation not permitted" is printed to stderr, is that correct? If that's the case, we could check that stdout is empty and/or stderr is non-empty.
> 
> But we still have to solve the OpenShift problem.

Ok that's right. The message is shown at stderr, not stdout. So we'd need to check the "offline" case for podman, and the stderr case for crio.
For openshift, what we could do is return true in case the conditionals for containers are met, instead of returning false. That would maintain the current behaviour.

olysonek:
>  For openshift, what we could do is return true in case the conditionals for containers are met, instead of returning false. That would maintain the current behaviour.

Sure, but would that work for your use case? I thought you wanted the function to return False.

yrobla:
> > For openshift, what we could do is return true in case the conditionals for containers are met, instead of returning false. That would maintain the current behaviour.
> 
> Sure, but would that work for your use case? I thought you wanted the function to return False.

Well, after some conversations with Jiri i think that it makes sense to isolate my use case , and try to solve it on another commit. The intention of running tuned on the containers is having a rollback when tuned stops (either container or system), so i think it makes sense to just return true in that case, and figure later.

tuned/daemon/daemon.py
```diff
@@ -122,11 +122,18 @@ def _notify_profile_changed(self, profile_names, result, errstr):
 		return errstr
 
 	def _full_rollback_required(self):
-		retcode, out = self._cmd.execute(["systemctl", "is-system-running"], no_errors = [0])
+		retcode, out, err_msg = self._cmd.execute(["systemctl", "is-system-running"], no_errors = [0], return_err = True)
 		if retcode < 0:
 			return False
+
+                # if there is some stderr and not stdout, it means systemctl runs in container
```
olysonek: There are spaces instead of tabs again ;) (in all the added lines).

olysonek:
> Well, after some conversations with Jiri i think that it makes sense to isolate my use case , and try to solve it on another commit. The intention of running tuned on the containers is having a rollback when tuned stops (either container or system), so i think it makes sense to just return true in that case, and figure later.

OK then. Thanks!

olysonek:
I think this can be merged now. @yarda @jmencak ?

tuned/daemon/daemon.py
```diff
@@ -122,11 +122,18 @@ def _notify_profile_changed(self, profile_names, result, errstr):
 		return errstr
 
 	def _full_rollback_required(self):
-		retcode, out = self._cmd.execute(["systemctl", "is-system-running"], no_errors = [0])
+		retcode, out, err_msg = self._cmd.execute(["systemctl", "is-system-running"], no_errors = [0], return_err = True)
 		if retcode < 0:
 			return False
+
+		# if there is some stderr and not stdout, it means systemctl runs in container
```
jmencak: I believe that for "containers" this is a wrong assumption to make and I have no idea how I'd test for "containers" since "containers" is such a wide term.  Anyways, this "works" in openshift-tuned container shipped in OCP 4.2:

Running "in a container" 1)
```
sh-4.2# systemctl is-system-running >/tmp/stdout 2>/tmp/stderr ; echo $?
0
sh-4.2# cat /tmp/stdout
sh-4.2# cat /tmp/stderr
Running in chroot, ignoring request.
```
but this does not:

Running "in a container" 2)
```
sh-4.2# SYSTEMD_IGNORE_CHROOT=1 systemctl is-system-running >/tmp/stdout 2>/tmp/stderr ; echo $?
0
sh-4.2# cat /tmp/stdout
running
sh-4.2# cat /tmp/stderr
sh-4.2# 
```

You can probably detect you're running in a chroot()ed env though if you make sure `SYSTEMD_IGNORE_CHROOT=0`.  The question is, whether this is what we want.  I really only do not want to regress at this point.  If we keep this code, the comment should be rephrased and very likely a search for "Running in chroot" introduced not to make any wrong assumptions.  Will that error message "Running in chroot" never change though?  This all seems very brittle.  Can someone remind me exactly where is the original code broken given one uses `SYSTEMD_IGNORE_CHROOT=1` when using tuned?  BTW, I do not use `SYSTEMD_IGNORE_CHROOT=1` when running tuned in a container, so perhaps I should fix my stuff?  Thanks!

PR #198, opened 2019-10-13
added new profile for k8s-container

jianzzha:
This new profile is targeted for Kubernetes/openshift worker nodes running non real time RHEL. It's based off the existing cpu-partitioning profile, with kvm and ksm related setup removed. Also, "nohz" and "nohz_full" are removed due to this BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1760644

### DIFF
--- a/profiles/k8s-container/tuned.conf
+++ b/profiles/k8s-container/tuned.conf
@@ -1,0 +1,57 @@
+# tuned configuration
+#
+
+[main]
+summary=Optimize for CPU partitioning
+include=network-latency
+
+[variables]
+# User is responsible for updating variables.conf with variable content such as isolated_cores=X-Y
+include=/etc/tuned/k8s-container-variables.conf
+
+isolated_cores_assert_check = \\${isolated_cores}
+# Fail if isolated_cores are not set
+assert1=${f:assertion_non_equal:isolated_cores are set:${isolated_cores}:${isolated_cores_assert_check}}
+
+# tmpdir
+tmpdir=${f:strip:${f:exec:/usr/bin/mktemp:-d}}
+# Non-isolated cores cpumask including offline cores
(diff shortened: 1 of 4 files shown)


jianzzha:
cpu-partitioning has no isolcpu kernel option, but I keep it here, do you see issue with it? @lcapitulino

profiles/k8s-container/tuned.conf
```diff
@@ -0,0 +1,57 @@
+# tuned configuration
+#
+
+[main]
+summary=Optimize for CPU partitioning
```
yarda, line 5: Here should be some good summary what the profile is good for. It's for users to orientate in the profiles.

profiles/k8s-container/tuned.conf
```diff
@@ -0,0 +1,57 @@
...
+[scheduler]
+isolated_cores=${isolated_cores}
+ps_blacklist=.*pmd.*;.*PMD.*;^DPDK
+
+[bootloader]
+priority=10
+initrd_remove_dir=True
+initrd_dst_img=tuned-initrd.img
+initrd_add_dir=${tmpdir}
+cmdline_cpu_part=+isolcpus=${isolated_cores} rcu_nocbs=${isolated_cores} tuned.non_isolcpus=${not_isolated_cpumask} intel_pstate=disable nosoftlockup
```
yarda, line 57: The option should be named `cmdline_k8s_container` to differentiate that it belongs to the k8s_container profile.

jianzzha:
@yarda The situation has greatly changed since this PR was open. On openshift side we have node tuning operator and performance addon operator which now can do pretty much everything we requested and we have confirmed that they worked via performance test. I guess RHEL based openshift has much chance to be an option. Let's close this PR unless there is K8S real use case on RHEL - we can always adopt what  tuning we put in the node tuning operator and performance addon operator if we need to do it on RHEL.

PR #199, opened 2019-10-14
Add null value check for kernel option when a tuned profile is applied

jianzzha:
This is to address https://bugzilla.redhat.com/show_bug.cgi?id=1755854

There are cases where the user wants to use a tuned profile without isolcpu; when a empty string is put in  variables.conf file, such as isolated_cores="", the kernel option line will have a null value key-pair like 'isolated_cores='. Even though this has no negative impact, it doesn't look good.

The fix is to simply remove 'key=' or 'key=""'

### DIFF
--- a/profiles/k8s-container/00-tuned-pre-udev.sh
+++ b/profiles/k8s-container/00-tuned-pre-udev.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+type getargs >/dev/null 2>&1 || . /lib/dracut-lib.sh
+
+cpumask="$(getargs tuned.non_isolcpus)"
+
+file=/sys/devices/virtual/workqueue/cpumask
+
+log()
+{
+  echo "tuned: $@" >> /dev/kmsg
+}
+
+if [ -n "$cpumask" ]; then
+  log "setting workqueue CPU mask to $cpumask"
+  if ! echo $cpumask > $file 2>/dev/null; then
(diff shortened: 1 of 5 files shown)


yarda:
@jianzzha could you please drop the ba54074 commit from this PR?

yarda:
Personally, I don't think that removing the 'OPTION=' strings is good thing to do. IMHO there are kernel parameters that relies on the 'OPTION=' syntax, so this patch wouldn't allow users to use them, e.g.:
```
acpi_osi=               # disable all strings
```
And some kernel modules can also rely on it.

If you don't like that `isolcpus=` can appear in the cmdline, I think the right thing is to set the parameters trough the helper function, e.g.:
```
cmdline=+${f:cmdline_drop_empty:isolcpus=${managed_irq}${isolated_cores}}
```
And put your NULL removal code to the cmdline_drop_empty built-in function.

Or add assert not allowing user to set the profile with empty isolated_cores, e.g.:
```
assert3=${f:assertion_non_equal:isolated_cores are not empty:${isolated_cores}:}
```

jianzzha:
@yarda cmdline_drop_empty is a better option. I will close this PR.

yarda:
> @yarda cmdline_drop_empty is a better option. I will close this PR.

Probably yes, but it's not yet implemented. Feel free to provide PR adding such function plugin, functions plugins are in tuned/profiles/functions directory.

PR #200, opened 2019-10-21
tests-beakerlib: Add first tests.

marmolak:
Upstream first.

Signed-off-by: Robin Hack <rhack@redhat.com>


yarda:
Robin thanks for this.

I propose adding the tests under the 'tests/beakerlib' directory. Currently under the 'tests' there are unit tests, we could move them to e.g. 'tests/unit'.

I will keep this open for others to comment their opinions.

olysonek:
I thought that the test metadata would be converted to fmf first and there would be no `Makefile` nor `PURPOSE` file.

TomasKorbar:
I had similar expectations.

PR #205, opened 2019-11-11
Enable copr builds and add packit config

dhodovsk:
Let us introduce [packit service](https://packit.dev) to you - the automation to integrate upstream open source projects into Fedora operating system.

After merging this PR, you are just a few steps away from RPM builds being automatically triggered on your PR's.
It means, that you'll be able to try and play with your change, packaged as an RPM.

But there is more. By using packit, you can for example enable adding new releases into Fedora Rawhide.

What are the next steps?
* Install [Packit-as-a-Service github application](https://github.com/marketplace/packit-as-a-service) in your repo
* In case you are first user, wait for account approval
* Enjoy the built RPMs!

For more info, please:
 * check out the documentation: https://packit.dev/docs/
 * contact @packit-service

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,14 @@
+downstream_package_name: tuned
+jobs:
+- job: copr_build
+  metadata:
+    targets:
+    - fedora-30-x86_64
+    - fedora-31-x86_64
+    - fedora-rawhide-x86_64
+  trigger: pull_request
+specfile_path: tuned.spec
+synced_files:
+- tuned.spec
+- .packit.yaml
+upstream_package_name: tuned


yarda:
@dhodovsk thanks, but:
- we currently cannot install the Packit app, because this is organization account, not personal account, we are trying to resolve it on the organizational level
- I think we currently don't want to sync the upstream spec file to the Fedora, could you please edit the PR for Packit not to do it?

yarda:
Packit app should be deployed now.

dhodovsk:
The syncing to fedora is not turned on in this config, for enabling it this should be added: 
```
jobs:
- job: sync_from_downstream
  trigger: commit
```
But with the configuration, as it is now, you are able to propose to downstream with using packit CLI tool.

I agree that it is unfortunate, having all these data required in packit.yaml. We aim to minimalize it as possible in the near future (for example [here](https://github.com/packit-service/packit/issues/140))

yarda:
> The syncing to fedora is not turned on in this config, for enabling it this should be added:
> 
> ```
> jobs:
> - job: sync_from_downstream
>   trigger: commit
> ```
> 
> But with the configuration, as it is now, you are able to propose to downstream with using packit CLI tool.
> 
> I agree that it is unfortunate, having all these data required in packit.yaml. We aim to minimalize it as possible in the near future (for example [here](https://github.com/packit-service/packit/issues/140))

@dhodovsk thanks. From the project docs the current behaviour wasn't clear to me, so it's probably OK now, so merging.

PR #207, opened 2019-11-15
scheduler: skip verification of IRQs with non-changeable affinity

yarda:
During the verification of the IRQs affinity if the verification fails
for the IRQ, try to set the current affinity. If it fails, skip the IRQ,
because it means that the IRQ affinity cannot be set. This is low cost
workaround which should be harmless.

Resolves: rhbz#1729936


tuned/plugins/plugin_scheduler.py
```diff
@@ -640,6 +640,13 @@ def _verify_irq_affinity(self, irq_description, correct_affinity,
 			log.info(consts.STR_VERIFY_PROFILE_VALUE_OK
 					% (irq_description, current_affinity))
 		else:
+			if not self._set_irq_affinity(self, irq, current_affinity, False):
```
olysonek: I don't think making changes to the system during verification is desirable. Also, I'm afraid it could potentially lead to security issues, as verification can be performed by any user on the system (based on our polkit policy).

olysonek:
I implemented a different solution to the problem in https://github.com/redhat-performance/tuned/pull/215. It doesn't add any writes to verification, and it respects `ignore_missing`, as you suggested.

yarda:
Considering the fact that verification is currently supported only with the running daemon the #215 is better solution, so closing.

PR #208, opened 2019-11-17
spec: added %check to run unit tests

yarda:
Also shipped 'tests' in sources.


yarda:
OK, thanks, merging.

PR #209, opened 2019-11-17
Use python3 interpreter by default

yarda:
This may ease development, because development environment will
be probably already on python3.


olysonek:
I think you may have missed `systemtap/varnetload`.

yarda:
Hopefully fixed.

PR #210, opened 2019-11-18
tests-beakerlib: Convert metadata to fmf format.

marmolak:
Converted by tmt tool.
Remove old metadata format.


yarda:
@marmolak thanks, could you please fix the contact field? Remove the 'None' and add your name.

yarda:
I need to reorg directories a bit, so I am merging it as is. @marmolak I will remove the None if you are not against.

PR #215, opened 2019-11-21
scheduler: Make IRQ SMP affinity verification respect ignore_missing

olysonek:
Changing the SMP affinity of some IRQs is not supported. The semantics of ignore_missing is that unsupported tunings are not treated as errors during verification. IRQ SMP affinity being unchangeable can be regarded as an unsupported tuning, so skip verification of these IRQs when ignore_missing is set.

The code assumes that the IRQ numbers point to the same device/event throughout the Tuned run time. We might need to break this assumption in the future (or maybe not; I haven't looked into how IRQ numbers are allocated). Either way, the assumption is already present in the code - rollback assumes the same thing.

Resolves: rhbz#1729936

### DIFF
--- a/tuned/plugins/plugin_scheduler.py
+++ b/tuned/plugins/plugin_scheduler.py
@@ -45,6 +45,8 @@ class IRQAffinities(object):
 	def __init__(self):
 		self.irqs = {}
 		self.default = None
+		# IRQs that don't support changing CPU affinity:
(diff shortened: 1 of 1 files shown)


olysonek:
Rebased on top of current master, so that we get results from Packit...

yarda:
Thanks, merged.

PR #216, opened 2019-11-22
Use re.DOTALL in the udev device matcher

olysonek:
In order to allow matching multiple parameter-value pairs in a single regex, we need to use re.DOTALL, because the parameter-value pairs are on separate lines. Consider the following regex:

^ID_MODEL=SD_MMC$.*^ID_MODEL_ID=0316$

Previously it would not much a string such as the following:

ID_MODEL=SD_MMC
ID_MODEL_ENC=SD\x2fMMC\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20
ID_MODEL_ID=0316

Now it does match.

It should be noted that this change is not entirely backwards-compatible. Some user-written regexes can now start to match where they shouldn't. For example the following regex will now match even if 'ID_MODEL_ID' and '0316' are on different lines.

ID_MODEL_ID.*0316

The primary motivation for this change is making the udev matcher behave the same as the cpuinfo matcher that will be written to resolve rhbz#1748965.

### DIFF
--- a/tuned/hardware/device_matcher_udev.py
+++ b/tuned/hardware/device_matcher_udev.py
@@ -21,4 +21,4 @@ class DeviceMatcherUdev(device_matcher.DeviceMatcher):
 		for key, val in list(items):
 			properties += key + '=' + val + '\n'
 
-		return re.search(regex, properties, re.MULTILINE) is not None
+		return re.search(regex, properties, re.MULTILINE | re.DOTALL) is not None


yarda:
Thanks, you could probably also drop the 'is not None' because the matcher object returns True upon match, but I think it's on more places in the code, so we can clean it later.

yarda:
> In order to allow matching multiple parameter-value pairs in a single regex, we need to use re.DOTALL, because the parameter-value pairs are on separate lines. Consider the following regex:
> 
> ^ID_MODEL=SD_MMC$.*^ID_MODEL_ID=0316$
> 

I think we could go without the re.DOTALL (and not introduce change in the behaviour), e.g. the following regex works for the above example without the re.DOTALL:

^ID_MODEL=SD_MMC$(.|\n)*^ID_MODEL_ID=0316$

IMHO the only thing re.DOTALL does is that you can write just '.' instead of the '(.|\n)'. It's more simple syntax for multiline matching, but slightly more complicated and confusing if you want to match just one line, e.g. consider the following regex:

^ID_MODEL=.*SD_MMC$

with re.DOTALL flag the following string will also match:
ID_MODEL=HDD
ID_MODEL_ID=0316
COMMENT=ULTRA_SD_MMC

Which may be not what the user initially wanted. Of course, the regex can be rewritten as:
^ID_MODEL=[^\n]*SD_MMC$

So instead of the simple '.' user now has to write '[^\n]'.

Both approaches have the same power. The approach we will choose needs to be properly documented.

olysonek:
You're absolutely right. My change doesn't make sense. Somehow, during my experimentation I came to the conclusion that you can't match strings on different lines at all.

Thanks for looking into it!

I'll submit a revert PR.

PR #217, opened 2019-11-22
cpu: Update checking if EPB is supported

olysonek:
Update checking if EPB is supported so that it works with recent versions of the x86_energy_perf_policy tool. Newer versions of x86_energy_perf_policy, unlike older versions, exit with a zero exit code even if the CPU doesn't support EPB. Newer versions of the tool give no ouput on stdout if EPB is not supported, so check for that.

In the future, we might like to determine if EPB is supported by searching /proc/cpuinfo for specific CPU flags. However the solution described in the previous paragraph should work just fine for now.

Resolves: rhbz#1690929


yarda:
Thanks, LGTM, merged.

PR #220, opened 2019-11-22
bootloader: Don't abort patching if grub.cfg is empty

olysonek:
Previously, if the grub.cfg patching code ran into a grub.cfg file that was empty, all subsequent grub.cfg files would get skipped and would not be patched. This behaviour doesn't make much sense to me. We should skip empty grub.cfg files and patch all the non-empty ones.

Resolves: rhbz#1622646


yarda:
OK, thanks, merged.

PR #222, opened 2019-11-25
Replace python flexmock module with builtin mock

TomasKorbar:
Also replace dependency in specfile and alter dockerfile.
Mock code in globals.py has been removed because it appears that when i first wrote it i misunderstood its effects.

### DIFF
--- a/tests/unit/plugins/test_base.py
+++ b/tests/unit/plugins/test_base.py
@@ -4,7 +4,6 @@ except ImportError:
 	from collections import Mapping
 import tempfile
 import unittest2
-import flexmock
 
 from tuned.monitors.repository import Repository
 import tuned.plugins.decorators as decorators
@@ -16,10 +15,6 @@ import tuned.plugins as plugins
 import tuned.consts as consts
 from tuned import storage
 import tuned.plugins.base
-
-tuned.plugins.base.log = flexmock.flexmock(info = lambda *args: None,\
-	error = lambda *args: None,debug = lambda *args: None,\
-	warn = lambda *args: None)
 
 temp_storage_file = tempfile.TemporaryFile(mode = 'r')
 consts.DEFAULT_STORAGE_FILE = temp_storage_file.name
--- a/tests/unit/storage/test_storage.py
+++ b/tests/unit/storage/test_storage.py
@@ -1,26 +1,31 @@
 import unittest
-from flexmock import flexmock
+from unittest.mock import Mock
+from unittest.mock import call
 import tuned.storage
 
 class StorageStorageTestCase(unittest.TestCase):
 	def test_set(self):
-		mock_provider = flexmock()
+		mock_provider = Mock()
 		factory = tuned.storage.Factory(mock_provider)
 		storage = factory.create("foo")
 
-		mock_provider.should_receive("set").with_args("foo", "optname", "optval").once
 		storage.set("optname", "optval")
+		mock_provider.set.assert_called_once_with("foo", "optname", "optval")
 
 	def test_get(self):
-		mock_provider = flexmock()
+		mock_provider = Mock()
+		mock_provider.get.side_effect = [ None, "defval", "somevalue" ]
 		factory = tuned.storage.Factory(mock_provider)
 		storage = factory.create("foo")
 
-		mock_provider.should_receive("get").with_args("foo", "optname", None).and_return(None).once.ordered
-		mock_provider.should_receive("get").with_args("foo", "optname", "defval").and_return("defval").once.ordered
-		mock_provider.should_receive("get").with_args("foo", "existing", None).and_return("somevalue").once.ordered
-
-		self.assertIsNone(storage.get("optname"))
+		self.assertEqual(None, storage.get("optname"))
 		self.assertEqual("defval", storage.get("optname", "defval"))
 		self.assertEqual("somevalue", storage.get("existing"))
 
+		calls = [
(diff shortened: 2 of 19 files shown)


yarda:
@TomasKorbar thanks, could you also enable %check in spec file on rhel-8?

TomasKorbar:
Done

yarda:
Thanks.

yarda:
I will wait for copr-build and if it passes I will merge it.

olysonek:
>  Mock code in globals.py has been removed because it appears that when i first wrote it i misunderstood its effects.

I don't think you're the one who added that. It looks like it was added by Jan Vcelak in commit eb4b1d94. It looks to me like it's trying to replace the logger function by a null logger. If it works (which I haven't yet verified), I think we should keep it.

yarda:
Unit tests passed. I am not sure whether we need to mock the logger (I haven't noticed any bad effects by not doing so), but doing so should be harmless at least. @TomasKorbar could you look into it? If it correctly mocks the logger, could you use python-mock for it?

TomasKorbar:
Alright. Now i understand the origin and context of this code.
When i first wrote these unittests i was running them separately one module at a time to test them. With this approach tuned.logs was not overridden by flexmock because no module imported globals module, so i used flexmock in each module to override it manually. We did not notice any effect on tests because when they were run altogether, globals module overridden tuned.logs module(there was import in tests.unit.plugin.test_base).
I replaced flexmock with simple monkey patch, because i think it is more elegant and added import of globals to each subpackage of tests. This approach will stop logging in both scenarios, running tests together and separately.

tests/unit/plugins/test_base.py
```diff
@@ -17,10 +16,6 @@
 from tuned import storage
 import tuned.plugins.base
 
-tuned.plugins.base.log = flexmock.flexmock(info = lambda *args: None,\
-	error = lambda *args: None,debug = lambda *args: None,\
-	warn = lambda *args: None)
-
```
olysonek: Yes, this is exactly what I was hoping keeping the mocking in globals.py would allow us to do. Thanks for looking into it!

olysonek:
Currently I can't get the unit tests to work with python2 (`make test PYTHON=python2`). However, this is an older problem - it's caused by cff78c537c157. I'll se what we can do about it.

olysonek:
I may be wrong here, but I think that the fact that it currently works with python3 is a python3 feature added in python3.4. The doc says the following about unittest discovery:
"Changed in version 3.4: Test discovery supports namespace packages." [1]

We now agreed with @yarda that running unittest with python2 upstream is not important for us at this point. So I don't think we need to spend time on it.

[1] https://docs.python.org/3/library/unittest.html#test-discovery

tests/unit/storage/test_storage.py
```diff
@@ -1,26 +1,30 @@
...
-		self.assertIsNone(storage.get("optname"))
-		self.assertEqual("defval", storage.get("optname", "defval"))
-		self.assertEqual("somevalue", storage.get("existing"))
+		calls = [
+				 call("foo", "optname", None),
+				 call("foo", "optname", "defval"),
+				 call("foo", "existing", None)
+				]
 
+		mock_provider.get.assert_has_calls(calls)
```
olysonek: This no longer checks the return values, right? Is there a way to fix that?

TomasKorbar: This test was badly written. 3 removed calls did the following: they created function `get` and recorded that it should be called 3 times with some arguments and then return None, "defval" and "somevalue". So the assertions were useless because mocked method always returned what the 3 removed calls told it to return.
I do not think that checking the return value here is necessary because it's sufficient implementation would be to set return value of `get` to some value and assert that the same value is returned.

olysonek: I think the point is that the assertions call `storage.get()`, not `mock_provider.get()`. I think the point of the test is to check that `storage.get` correctly passes its arguments to `mock_provider.get()` and then passes back the return value. Right now, the test doesn't check the second part - passing back the return value.

olysonek: I think we can achieve what we want using the following patch. There might be a better way to write it, but this is the best I was able to come up with and it should work.
```
diff --git a/tests/unit/storage/test_storage.py b/tests/unit/storage/test_storage.py
index ca6eef8..2d2709a 100644
--- a/tests/unit/storage/test_storage.py
+++ b/tests/unit/storage/test_storage.py
@@ -14,12 +14,13 @@ class StorageStorageTestCase(unittest.TestCase):
 
 	def test_get(self):
 		mock_provider = Mock()
+		mock_provider.get.side_effect = [ None, "defval", "somevalue" ]
 		factory = tuned.storage.Factory(mock_provider)
 		storage = factory.create("foo")
 
-		storage.get("optname")
-		storage.get("optname", "defval")
-		storage.get("existing")
+		self.assertEqual(None, storage.get("optname"))
+		self.assertEqual("defval", storage.get("optname", "defval"))
+		self.assertEqual("somevalue", storage.get("existing"))
 
 		calls = [
 				 call("foo", "optname", None),
```

olysonek: (Inspired by https://stackoverflow.com/a/24897297)

TomasKorbar: Fixed.

yarda:
Thanks, merging.

PR #223, opened 2019-11-26
Remove defirqaffinity.py

watologo1:
The last callers to this realtime profile helper script
had been nuked by this commit:
c2dfc1f80695f0cdd6782a6d72787b378128fc47

### DIFF
--- a/libexec/defirqaffinity.py
+++ b/libexec/defirqaffinity.py
@@ -82,6 +82,7 @@ def verify(shouldbemask):
 	sys.exit(0)
 
 
+sys.stderr.write("defirqaffinity.py is deprecated. Use isolated_cores or other built-in functionality instead.\n")
 
 # adjust default_smp_affinity
 cpulist = parse_def_affinity(irqpath + "default_smp_affinity")


watologo1:
@olysonek I cannot add reviewers? The original comit to remove defirqaffinity.py callers came from you. So you should certainly have a look at this...

olysonek:
Is there a reason to remove the script? People are free to use it in their own profiles. And they might be using it now. So I'd rather not remove it unless there is a compelling reason to do so.

yarda:
I think we should try convincing users to use built-in features instead of the external apps. It would be also better for maintainability to drop support of obsoleted code. So it's +1 from me to try to remove it. We can always return it back if it shows there are still people relying on it.

yarda:
> I think we should try convincing users to use built-in features instead of the external apps. It would be also better for maintainability to drop support of obsoleted code. So it's +1 from me to try to remove it. We can always return it back if it shows there are still people relying on it.

But I would mark it as obsoleted at first for few releases, instead of hard dropping it now.

olysonek:
Ok, agreed. Let's mark it as deprecated and remove it later.

olysonek:
@watologo1 I think we can modify the defirqaffinity script so that it prints a warning about it being deprecated. The warning will propagate to the Tuned log.

olysonek:
> @watologo1 I think we can modify the defirqaffinity script so that it prints a warning about it being deprecated. The warning will propagate to the Tuned log.

Forgot to say, the warning needs to be printed to stderr.

watologo1:
Fine. Do you expect me to re-submit the obsolete patch?

yarda:
> Fine. Do you expect me to re-submit the obsolete patch?

If you could do so, it would be great :) If not, we could handle it ourselves.

watologo1:
@yarda Here we go.

yarda:
@watologo1 thanks, merging.

PR #224, opened 2019-11-26
Make python not write bytecode files when unit tests are run

olysonek:
Previously, when unit tests were run, python would leave behind files with bytecompiled modules. It's not useful, it just clutters the repository. And it could potentially lead to problems - it seems to have lead to some problem for me earlier today when running git bisect to find out when `make test` started failing.


yarda:
LGTM, thanks, merged.

PR #225, opened 2019-11-26
Fix a hang on SIGHUP when the preset profile does not exist

olysonek:
Please take a look at the second commit. The `return False` when the daemon is not running has been there for ages, however I'm not sure if it serves a specific purpose. Do you think we can drop it?

### DIFF
--- a/tuned/daemon/controller.py
+++ b/tuned/daemon/controller.py
@@ -138,7 +138,11 @@ class Controller(tuned.exports.interfaces.ExportableInterface):
 			stop_ok = self.stop()
 			if not stop_ok:
 				return False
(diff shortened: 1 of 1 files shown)


olysonek:
I need to think this through. Please don't merge.

olysonek:
@yarda I'm happy with this now. Can you please review?

olysonek:
@jmencak It would be great if you could test it.

jmencak:
I can no longer reproduce the "hang" with this patch.  Thank you!

yarda:
@olysonek thanks for the patch and @jmencak thanks for testing, merging.

PR #228, opened 2019-11-27
unit: accept "True" for replace and enabled options

yarda:


yarda:
This is to accept more "Pythonic" booleans and not to relax the syntax to case insensitive. In fact we could probably drop the non-string True and 1 values, because we are using ConfigObj in string mode and not in the 'unrepr' mode, but our unit tests also checks for non string booleans so keep them for now (and also stay compatible with the unrepr mode).

yarda:
Thanks for review.

Issue #227, opened 2019-11-27
RFE: a way to return instance option to default


yarda:
For example `profile1`:
```
[cpu]
force_latency=10
no_turbo=1

[disk]
readahead=4098
```
I `profile2` I want to include `profile1` and drop just the `force_latency`. Currently I have to write:
```
[main]
include=profile1

[cpu]
replace=true
no_turbo=1
```
I.e. there is no way to just drop one option and return it to its defaults (i.e. usually None).

yarda:
I thought about multiple ways how to extend the syntax to support this:
a) `force_latency=`
This may not work for options where user needs to set empty string ""
b) special operator on the left side, e.g. `:`, e.g. `force_latency:=`
When using it on the left side we probably do not need to implement escaping, but it will require some 'hackish' implementation.
c) special operator on the first position of the option, e.g. `:`, e.g. `:force_latency=`
This is quite unclear syntax, and we will also need additional `=` because our parser requires it.
d) special operator on the right side, e.g. `:`, e.g. `force_latency=:`
This will require escaping, because there may be options where user wants just literal `:`
e) special option which will list what to drop, e.g. `drop=option1[,option2[,..]]`
I.e.:
```
[cpu]
drop=force_latency
```
And the `drop` will be not inhereted, just utilized in the profile merger.

Any other opinions?

olysonek:
@yarda I think I like best the `drop` option approach.

PR #229, opened 2019-11-27
Added drop option allowing dropping options from inherited plugins

yarda:
Consider the following:
profile_1:
[cpu]
force_latency=10
no_turbo=1

[disk]
readahead=4098

If you want to include it in profile_2 and just drop the force_latency
option now you can write:
[main]
include=profile_1

[cpu]
drop=force_latency

Previously, you had to use the replace option and add all other options
which were used in the inherited profiles like:
[main]
include=profile_1

[cpu]
replace=true
no_turbo=1

There can be named multiple options in the drop. As the delimiter comma
',' or semicolon ';' can be used, spaces are ignored, e.g.:
[cpu]
drop=force_latency, no_turbo

Fixes: #227

### DIFF
--- a/tuned/profiles/merger.py
+++ b/tuned/profiles/merger.py
@@ -43,6 +43,10 @@ class Merger(object):
 					profile_a.units[unit_name].script_pre = unit.script_pre
 				if unit.script_post is not None:
 					profile_a.units[unit_name].script_post = unit.script_post
+				if unit.drop is not None:
+					for option in unit.drop:
+						profile_a.units[unit_name].options.pop(option, None)
+					unit.drop = None
 				if unit_name == "script" and profile_a.units[unit_name].options.get("script", None) is not None:
 					script = profile_a.units[unit_name].options.get("script", None)
 					profile_a.units[unit_name].options.update(unit.options)
--- a/tuned/profiles/unit.py
+++ b/tuned/profiles/unit.py
@@ -1,11 +1,12 @@
 import collections
+import re
 
 class Unit(object):
 	"""
 	Unit description.
 	"""
 
-	__slots__ = [ "_name", "_type", "_enabled", "_replace", "_devices", "_devices_udev_regex", \
+	__slots__ = [ "_name", "_type", "_enabled", "_replace", "_drop", "_devices", "_devices_udev_regex", \
 		"_cpuinfo_regex", "_uname_regex", "_script_pre", "_script_post", "_options" ]
 
 	def __init__(self, name, config):
@@ -13,6 +14,9 @@ class Unit(object):
 		self._type = config.pop("type", self._name)
 		self._enabled = config.pop("enabled", True) in [True, "True", "true", 1, "1"]
 		self._replace = config.pop("replace", False) in [True, "True", "true", 1, "1"]
+		self._drop = config.pop("drop", None)
+		if self._drop is not None:
+			self._drop = re.split(r"\b\s*[,;]\s*", str(self._drop))
 		self._devices = config.pop("devices", "*")
 		self._devices_udev_regex = config.pop("devices_udev_regex", None)
 		self._cpuinfo_regex = config.pop("cpuinfo_regex", None)
@@ -44,6 +48,14 @@ class Unit(object):
 	@property
 	def replace(self):
 		return self._replace
+
+	@property
+	def drop(self):
+		return self._drop
+
+	@drop.setter
+	def drop(self, value):
+		self._drop = value
 
 	@property
 	def devices(self):


yarda:
No syntax check added.

yarda:
Also I am not sure whether 'drop' is the best name for this new options, maybe we could use 'defaults' or even different name?

olysonek:
> Also I am not sure whether 'drop' is the best name for this new options, maybe we could use 'defaults' or even different name?

I like `drop` - it correctly implies that the options are effectively dropped from the profile. `defaults` would imply that some default value is used for the option, which is not the case.

olysonek:
It behaves a bit oddly with multiple inheritance. Consider the following profiles.

Profile `a`:
```
[sysctl]
vm.swappiness=10
```

Profile `b1`:
```
[sysctl]
vm.swappiness=20
```

Profile `b2`:
```
[main]
include=b1

[sysctl]
drop=vm.swappiness
```

Profile `c`:
```
[main]
include=a, b2
```

If I apply profile `c`, Tuned will not set `vm.swappiness`. It would make more sense to me if it set `vm.swappiness=10`, taken from profile `a`.

yarda:
> It behaves a bit oddly with multiple inheritance. Consider the following profiles.
> 
> Profile `a`:
> 
> ```
> [sysctl]
> vm.swappiness=10
> ```
> 
> Profile `b1`:
> 
> ```
> [sysctl]
> vm.swappiness=20
> ```
> 
> Profile `b2`:
> 
> ```
> [main]
> include=b1
> 
> [sysctl]
> drop=vm.swappiness
> ```
> 
> Profile `c`:
> 
> ```
> [main]
> include=a, b2
> ```
> 
> If I apply profile `c`, Tuned will not set `vm.swappiness`. It would make more sense to me if it set `vm.swappiness=10`, taken from profile `a`.

Why? What it does is:
```
a    ->     b1    ->    b2    ->    c
10          20           DROP            => nothing
```

yarda:
> > It behaves a bit oddly with multiple inheritance. Consider the following profiles.
> > Profile `a`:
> > ```
> > [sysctl]
> > vm.swappiness=10
> > ```
> > 
> > 
> > Profile `b1`:
> > ```
> > [sysctl]
> > vm.swappiness=20
> > ```
> > 
> > 
> > Profile `b2`:
> > ```
> > [main]
> > include=b1
> > 
> > [sysctl]
> > drop=vm.swappiness
> > ```
> > 
> > 
> > Profile `c`:
> > ```
> > [main]
> > include=a, b2
> > ```
> > 
> > 
> > If I apply profile `c`, Tuned will not set `vm.swappiness`. It would make more sense to me if it set `vm.swappiness=10`, taken from profile `a`.
> 
> Why? It works the following way:
> 
> ```
> a    ->     b1    ->    b2    ->    c
> 10          20          DROP            => nothing
> ```

Just try replacing `drop=vm.swappiness` by `vm.swappiness=30` in your example and the result will be 30. So the logic is still the same. Why it should work differently for `drop`? I think it would be confusing for users.

 ```
 a    ->     b1    ->    b2    ->    c
 10          20          30            => result is 30
 ```

The full example for reference:

Profile `a`:
```
[sysctl]
vm.swappiness=10
```
Profile `b1`:
```
[sysctl]
vm.swappiness=20
 ```
Profile `b2`:
```
[main]
include=b1

[sysctl]
vm.swappiness=30
```
Profile `c`:
```
[main]
include=a, b2
```

**Comment by yarda** [comment]
The minor problem is that it doesn't expand variables, but let's call it feature for now :) I think the benefit of supporting variables in drop is low and other special options like 'replace', 'enabled', 'type' also don't support it at the moment.

**Comment by olysonek** [comment]
> Why? It works the following way:
> 
> ```
> a    ->     b1    ->    b2    ->    c
> 10          20          DROP            => nothing
> ```

How it currently works is one thing, but how it should work is another thing. The profiles form a tree, so they should be handled like a tree. The problem is in how the multiple-include feature was implemented. It should do depth-first search on the profiles. It should *not* convert the tree to a chain of profiles and then merge that chain.

```
    b1
     |
a   b2
  \/
  c
```
The merge code should first merge `b2` into `b1`, which would result in a temporary profile, let's call it `b-temp`, which would not contain `vm.swappiness`. Only then should the code merge `b-temp` into `a`, giving the final profile, `c`, containing `vm.swappiness` (originating from profile `a`).

> Just try replacing `drop=vm.swappiness` by `vm.swappiness=30` in your example and the result will be 30. So the logic is still the same. Why it should work differently for `drop`? I think it would be confusing for users.
> 
> ```
> a    ->     b1    ->    b2    ->    c
> 10          20          30            => result is 30
> ```

Notice that you will get the same result also with the 'depth-first search' merge.

The thing is that the chain merge will work as expected when the profiles only add or replace options. If the profiles remove options, be it using the `drop` directive or the `replace` directive, it no longer works correctly.

To repeat, the problem is not in this PR implementing `drop` -  the same problem will occur with the `replace` option. The problem is in the implementation of the profile merge.

yarda:
> > Why? It works the following way:
> > ```
> > a    ->     b1    ->    b2    ->    c
> > 10          20          DROP            => nothing
> > ```
> 
> How it currently works is one thing, but how it should work is another thing. The profiles form a tree, so they should be handled like a tree. The problem is in how the multiple-include feature was implemented. It should do depth-first search on the profiles. It should _not_ convert the tree to a chain of profiles and then merge that chain.
> 
> ```
>     b1
>      |
> a   b2
>   \/
>   c
> ```
> 
> The merge code should first merge `b2` into `b1`, which would result in a temporary profile, let's call it `b-temp`, which would not contain `vm.swappiness`. Only then should the code merge `b-temp` into `a`, giving the final profile, `c`, containing `vm.swappiness` (originating from profile `a`).
> 
> > Just try replacing `drop=vm.swappiness` by `vm.swappiness=30` in your example and the result will be 30. So the logic is still the same. Why it should work differently for `drop`? I think it would be confusing for users.
> > ```
> > a    ->     b1    ->    b2    ->    c
> > 10          20          30            => result is 30
> > ```
> 
> Notice that you will get the same result also with the 'depth-first search' merge.
> 
> The thing is that the chain merge will work as expected when the profiles only add or replace options. If the profiles remove options, be it using the `drop` directive or the `replace` directive, it no longer works correctly.
> 
> To repeat, the problem is not in this PR implementing `drop` - the same problem will occur with the `replace` option. The problem is in the implementation of the profile merge.

I am not against but it's change in behaviour. I think it should be discussed elsewhere because it's unrelated to this PR.

yarda:
> To repeat, the problem is not in this PR implementing `drop` - the same problem will occur with the `replace` option. The problem is in the implementation of the profile merge.

@olysonek can I take it as an approval for this PR?

olysonek:
Feel free to merge it.

yarda:
Thanks, merging.

PR #230, opened 2019-11-29
latency: support None value

yarda:
This allows writing e.g.:
[cpu]
force_latency=cstate.name:XYZ|None

In this case if C-state with name XYZ doesn't exist force_latency
will be evaluated to 'None' and nothing will be set and no
error will be reported. If '|None' is omitted, error will be
reported.

Another example:
[cpu]
force_latency=None

Here, it will also set nothing and no error will be reported.

When 'None' is encountered it stops parsing the latency, so writing
'None|1' will also result in no latency set. Both 'none' and 'None' can
be used interchangeably with the same effect.


yarda:
Thanks, merging.

PR #233, opened 2019-12-09
latency-performance: Update sysctl tunings

olysonek:
Update sysctl tunings in the latency-performance profile based on feedback from the performance team.

Resolves: rhbz#1779759


yarda:
LGTM, thanks.

PR #239, opened 2020-01-06
profiles: Make sure variables are defined before use

olysonek:
Make sure the isolated_cores and no_balance_cores variables are defined before any of the variables that use them are defined. This enforces a certain ordering of variable expansions so that child profiles can set the variables directly in the profile (tuned.conf), e.g.:

[main]
include=cpu-partitioning

[variables]
isolated_cores=3

Resolves: rhbz#1781664

### DIFF
--- a/profiles/cpu-partitioning/tuned.conf
+++ b/profiles/cpu-partitioning/tuned.conf
@@ -10,6 +10,10 @@ include=network-latency
 include=/etc/tuned/cpu-partitioning-variables.conf
 
 isolated_cores_assert_check = \\${isolated_cores}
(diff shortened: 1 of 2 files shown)


yarda:
It's hack fixing existing profiles, but for new profiles people will still have to take this into account - i.e for variables that could come from the external variables file and if they could be overridden in the parent profile, people will have to dummy define them. I think that some bigger changes of the code are probably not worth the effort (e.g. redefining the OrderedDict update method just for the variables and adding new items to the top of the OrderedDict instead of the bottom and coping with the 'include'...) so let's call this feature and document it somewhere :)

yarda:
Thanks, merging

PR #243, opened 2020-01-31
Try-restart irqbalance service upon writing /etc/sysconfig/irqbalance

jmencak:
For context: https://github.com/openshift/cluster-node-tuning-operator/pull/111

### DIFF
--- a/profiles/functions
+++ b/profiles/functions
@@ -474,14 +474,18 @@ restore_logs_syncing() {
 }
 
 irqbalance_banned_cpus_clear() {
-    sed -i '/^IRQBALANCE_BANNED_CPUS=/d' /etc/sysconfig/irqbalance
+    sed -i '/^IRQBALANCE_BANNED_CPUS=/d' /etc/sysconfig/irqbalance || return
+    if [ ${1:-restart} = restart ]; then
+        systemctl try-restart irqbalance
+    fi
 }
 
 irqbalance_banned_cpus_setup() {
-    irqbalance_banned_cpus_clear
+    irqbalance_banned_cpus_clear norestart
     if [ -n "$1" ]; then
         echo "IRQBALANCE_BANNED_CPUS=$1" >> /etc/sysconfig/irqbalance
     fi
+    systemctl try-restart irqbalance
 }
 
 #


jmencak:
@olysonek PTAL.  Many thanks!

jmencak:
Linking relevant discussion: https://bugzilla.redhat.com/show_bug.cgi?id=1784645

profiles/functions
```diff
@@ -474,13 +474,13 @@ restore_logs_syncing() {
 }
 
 irqbalance_banned_cpus_clear() {
-    sed -i '/^IRQBALANCE_BANNED_CPUS=/d' /etc/sysconfig/irqbalance
+    sed -i '/^IRQBALANCE_BANNED_CPUS=/d' /etc/sysconfig/irqbalance && systemctl try-restart irqbalance
```
olysonek: Notice that due to this change, `irqbalance` will be restarted twice in a row when `irqbalance_banned_cpus_setup` is called. This seems like unnecessary overhead and it might occasionally fail - for some reason, systemd doesn't like restarting the same service multiple times in quick succession.

jmencak: Right, in that case I guess a proper "irqbalance" plugin (as requested in BZ1784645) might be a better idea that does the restart only once.  Thoughts?

jmencak: @olysonek , would something like this work for you until we get a proper `irqbalance` plugin?
```
irqbalance_banned_cpus_clear() {
    sed -i '/^IRQBALANCE_BANNED_CPUS=/d' /etc/sysconfig/irqbalance && \
        test ${1:-restart} = restart && systemctl try-restart irqbalance
}

irqbalance_banned_cpus_setup() {
    irqbalance_banned_cpus_clear norestart
    if [ -n "$1" ]; then
        echo "IRQBALANCE_BANNED_CPUS=$1" >> /etc/sysconfig/irqbalance
    fi
    systemctl try-restart irqbalance
}
```

olysonek: > @olysonek , would something like this work for you until we get a proper `irqbalance` plugin?

Yes, that's what I had in mind.

jmencak: Thanks!  Pushed an update.

jmencak:
Is there anything else I can do so that we get this merged?  Thanks!

profiles/functions
```diff
@@ -474,14 +474,16 @@ restore_logs_syncing() {
 }
 
 irqbalance_banned_cpus_clear() {
-    sed -i '/^IRQBALANCE_BANNED_CPUS=/d' /etc/sysconfig/irqbalance
+    sed -i '/^IRQBALANCE_BANNED_CPUS=/d' /etc/sysconfig/irqbalance && \
+        test ${1:-restart} = restart && systemctl try-restart irqbalance
```
yarda, line 478: Could you please rewrite this to 'if' (the part starting with the 'test')? We didn't define API for functions, but I would take it the way that the function can return non zero exit code in case of some problem. This would return non zero (error) exit code even when there is no error and the argument is just e.g. 'norestart' which could have other unintended consequences - e.g. imagine situation when this function would be used as the last one in the script.sh (without the explicit return 0), which could cause the script.sh to return non zero exit code in case everything is OK.

jmencak: Makes sense.

olysonek:
Good point. My approval holds :).

yarda:
Thanks, merged.

Issue #246, opened 2020-02-16
bashism in Makefile


schmidicom:
The Makefile contains bashism which should be revised to make an installation work with something other than bash.

See: https://bugs.gentoo.org/706164

ekohl:
Given #247 was merged, shouldn't this be closed?

PR #247, opened 2020-02-16
drop bashism syntax, make it happy with dash

dlan17:
see bug report at
https://github.com/redhat-performance/tuned/issues/246
https://bugs.gentoo.org/706164

Signed-off-by: Yixun Lan <dlan@gentoo.org>


yarda:
@dlan17 thanks and sorry - the scripts were primary written for bash, so there may be more bash specific syntax, but we are open for fixing it :)

PR #248, opened 2020-02-25
Unapply bootcmdline settings even when grub.cfg does not exits.

jmencak:
Currently when `grub.cfg` file is not found, the [bootloader] plugin still modifies `/etc/tuned/bootcmdline` file.  However, during rollbacks, the settings applied to that file are not rolled back.  This becomes a problem in (containerized) environments when `grub.cfg` file does not exist.  This patch will make the behaviour more consistent and unapply the settings in `/etc/tuned/bootcmdline` even when no `grub.cfg` is found.

### DIFF
--- a/tuned/plugins/plugin_bootloader.py
+++ b/tuned/plugins/plugin_bootloader.py
@@ -93,10 +93,10 @@ class BootloaderPlugin(base.Plugin):
 		return self._cmd.add_modify_option_in_file(consts.BOOT_CMDLINE_FILE, d)
 
 	def _remove_grub2_tuning(self):
+		self._patch_bootcmdline({consts.BOOT_CMDLINE_TUNED_VAR : "", consts.BOOT_CMDLINE_INITRD_ADD_VAR : ""})
 		if not self._grub2_cfg_file_names:
 			log.info("cannot find grub.cfg to patch")
 			return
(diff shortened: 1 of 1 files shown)


jmencak:
> Thanks! Makes sense. The question is though, should `/etc/tuned/bootcmdline` be modified in the first place when applying the tuning, if no grub.cfg is found?

Yeah, thought of that too.  We have two options really.  Either go with this (1) approach or not to generate `/etc/tuned/bootcmdline` no `grub.cfg` is found (2).  From a containerized tuned viewpoint, calculating kernel boot parameters is useful even when `grub.cfg` doesn't exist.  So if we go with (2), for containers where no `grub.cfg` is shipped we'd need to ship it to get `/etc/tuned/bootcmdline` calculated.  I vote for (1), but I'm open to discussion.

olysonek:
> > Thanks! Makes sense. The question is though, should `/etc/tuned/bootcmdline` be modified in the first place when applying the tuning, if no grub.cfg is found?
> 
> Yeah, thought of that too. We have two options really. Either go with this (1) approach or not to generate `/etc/tuned/bootcmdline` no `grub.cfg` is found (2). From a containerized tuned viewpoint, calculating kernel boot parameters is useful even when `grub.cfg` doesn't exist. So if we go with (2), for containers where no `grub.cfg` is shipped we'd need to ship it to get `/etc/tuned/bootcmdline` calculated. I vote for (1), but I'm open to discussion.

That's interesting to know. The current solution is fine with me.

jmencak:
Is there anything I can do to help this merge?  Thanks!

PR #251, opened 2020-02-28
Adjust FILES section to point to the doc dir  containing the version info

chuckersjp:
The DOC file specified in the man tuned.8 page may actually be located under the tuned-VERSION of it but there is no indicator that that might be the case.

### DIFF
--- a/man/tuned.8
+++ b/man/tuned.8
@@ -59,7 +59,6 @@ Show version information.
 .SH "FILES"
 .nf
 /etc/tuned
-/usr/share/doc/tuned\fR[\fI-version\fR]/README
 .SH "SEE ALSO"
 .LP
 tuned.conf(5)
--- a/man/tuned.8
+++ b/man/tuned.8
@@ -59,7 +59,7 @@ Show version information.
 .SH "FILES"
 .nf
 /etc/tuned
-/usr/share/doc/tuned/README
+/usr/share/doc/tuned\fR[\fI-version\fR]/README
 .SH "SEE ALSO"
 .LP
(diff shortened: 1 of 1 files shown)


yarda:
> The DOC file specified in the man tuned.8 page may actually be located under the tuned-VERSION of it but there is no indicator that that might be the case.

This is not correct, i.e.:
https://github.com/redhat-performance/tuned/blob/398f943f7097e5b59dcc1465fa4bcd578c358d4c/Makefile#L208
And:
https://github.com/redhat-performance/tuned/blob/398f943f7097e5b59dcc1465fa4bcd578c358d4c/Makefile#L36

So by default it uses `/usr/share/doc/tuned` (i.e. without the version). Probably your distro customized the DOCDIR location to versioned dir.

Could you update the PR to drop the README from the manual page? I think it needn't be mentioned in the manual page. Otherwise we would have to generate the manual page from the template, but I think it's not worth the effort.

chuckersjp:
I see your point.  I am a bit confused how it did end up in `/usr/share/docs/tuned-{VERSION}` instead of what you have defined, especially with this being RHEL 7.5.

Let me see if I can find where I made this PR.

yarda:
> I see your point. I am a bit confused how it did end up in `/usr/share/docs/tuned-{VERSION}` instead of what you have defined, especially with this being RHEL 7.5.
> 
> Let me see if I can find where I made this PR.

The path is customized in RHEL-7 to versioned docdir, IIRC also Fedora used versioned docs for some time.

yarda:
Thanks, merged.

PR #256, opened 2020-03-23
builtin_functions: fixed check for number of arguments

yarda:
There can be lower number of arguments than the max limit.
Also fixed definition of functions which wrongly used the API.

I.e. it should work this way:
__init__("FUNC", 3, 3)

__init__("FUNC", 4, 3)

__init__("FUNC", 3)

__init__("FUNC", 0, 3)

__init__("FUNC", 0)


yarda:
Hmm, git ate the comments, I will fix it.

olysonek:
Ah. Still approved :).

PR #258, opened 2020-03-25
[WIP]: Add a test for plugin listing using internal tuned API

olysonek:
Work-in-progress. Do not merge yet.


olysonek:
The `kernel_settings` role no longer depends on the behaviour tested here, so the test case is no longer needed. Closing.

PR #260, opened 2020-04-15
Enable isolate_managed_irq to be defined in child profiles.

jmencak:
This fixes the unability to use `isolate_managed_irq` from child profiles.

### DIFF
--- a/profiles/realtime/tuned.conf
+++ b/profiles/realtime/tuned.conf
@@ -29,6 +29,10 @@ isolated_cores_online_expanded=${f:cpulist_online:${isolated_cores}}
 assert2=${f:assertion:isolated_cores contains online CPU(s):${isolated_cores_expanded}:${isolated_cores_online_expanded}}
 
 # Assembly managed_irq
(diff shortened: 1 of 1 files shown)


olysonek:
Thanks, good catch. I'll just test it once the copr build completes.

profiles/realtime/tuned.conf
```diff
@@ -29,6 +29,10 @@ isolated_cores_online_expanded=${f:cpulist_online:${isolated_cores}}
 assert2=${f:assertion:isolated_cores contains online CPU(s):${isolated_cores_expanded}:${isolated_cores_online_expanded}}
 
 # Assembly managed_irq
+# Make sure isolate_managed_irq is defined before any of the variables that
+# use it (such as regex_search_ternary) are defined, so that child profiles can set
```
olysonek, line 33: `s/regex_search_ternary/managed_irq/`

`regex_search_ternary` is a function.

jmencak: Thanks, good catch.  It was late :)

yarda:
Thanks, merged.

PR #261, opened 2020-04-17
Spectrum Scale ECE profile

bolinches:
Hi

This is a profile for Spectrum Scale Erasure Code Edition (ECE) parallel filesystem. It is mainly throughput-performance with the sysctl settings we require customers to set manually to the values set https://github.com/IBM/SpectrumScale_ECE_OS_READINESS/blob/master/sysctl.json

Idea is to offload that manual work into a profile and keep it up to date on tuned as profile in tuned for simplification instead of recommending the changes on the product documentation and checking for individual sysctl values (or other later on) we would document to use the specctrumscale-ece and check that is being used.

Thanks

EDIT: Previous Link was private internal repo, pointing to public now

### DIFF
--- a/man/tuned-profiles-spectrumscale-ece.7
+++ b/man/tuned-profiles-spectrumscale-ece.7
@@ -1,0 +1,59 @@
+.\"/* 
+.\" * All rights reserved
+.\" * Copyright (C) 2020 International Business Machines Corporation 
+.\" * Authors: Luis Bolinches
+.\" *
+.\" * This program is free software; you can redistribute it and/or
+.\" * modify it under the terms of the GNU General Public License
+.\" * as published by the Free Software Foundation; either version 2
+.\" * of the License, or (at your option) any later version.
+.\" *
+.\" * This program is distributed in the hope that it will be useful,
+.\" * but WITHOUT ANY WARRANTY; without even the implied warranty of
+.\" * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+.\" * GNU General Public License for more details.
+.\" *
+.\" * You should have received a copy of the GNU General Public License
+.\" * along with this program; if not, write to the Free Software
+.\" * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+.\" */
+.\" 
+.TH TUNED_PROFILES_SPECTRUMSCALE_ECE "7" "22 Apr 2020" "Fedora Power Management SIG" "tuned"
+.SH NAME
+tuned\-profiles\-spectrumscale-ece - description of profiles provided for Spectrum Scale Erasure Code Edition
+
+.SH DESCRIPTION
+These profiles are provided for Spectrum Scale Erasure Code Edition servers.
+
+.SH PROFILES
+The following profiles are provided:
+
+.TP
+.BI "spectrumscale-ece"
+Profile optimized for Spectrum Scale Erasure Code Edition based on throughput\-performance profile.
+It additionally enables automatic NUMA balancing and modifies some other
+performance related kernel parameters.
+
+.SH "FILES"
+.nf
+.I /etc/tuned/*
+.I /usr/lib/tuned/*
+
+.SH "SEE ALSO"
+.BR tuned (8)
+.BR tuned\-adm (8)
+.BR tuned\-profiles (7)
+.BR tuned\-profiles\-sap (7)
+.BR tuned\-profiles\-sap\-hana (7)
+.BR tuned\-profiles\-mssql (7)
+.BR tuned\-profiles\-atomic (7)
+.BR tuned\-profiles\-oracle (7)
+.BR tuned\-profiles\-realtime (7)
+.BR tuned\-profiles\-nfv\-host (7)
+.BR tuned\-profiles\-nfv\-guest (7)
+.BR tuned\-profiles\-cpu\-partitioning (7)
+.BR tuned\-profiles\-compat (7)
+.SH AUTHOR
+.nf
+Luis Bolinches <luis.bolinches@fi.ibm.com>
+
--- a/profiles/spectrumscale-ece/tuned.conf
+++ b/profiles/spectrumscale-ece/tuned.conf
@@ -1,0 +1,12 @@
+#
+# tuned configuration
+#
+
+[main]
+summary=Optimized for Spectrum Scale Erasure Code Edition Servers
+include=throughput-performance
+
+[sysctl]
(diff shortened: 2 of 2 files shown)


olysonek:
Hi! Please see https://github.com/redhat-performance/tuned/blob/master/DCO and provide your sign-off in the commit message. Also, please fix up the missing new line at the end of the file.

Would you like this profile to be distributed in Fedora?

bolinches:
Fedora is not officially supported to run ECE so you raise a really good point. I was looking more into Red Hat Enterprise Linux, and their derivates than Fedora only.

olysonek:
Ah, sorry, I didn't check the origin of the patch ;).

Anyway, if it's going to be in RHEL, it should be in Fedora too, if possible. From my point of view, it doesn't matter if it's supported or not.

I think we should put the profile in a separate subpackage. That's what we've been doing with product-specific profiles. What do you think?

Also, can you please put together a man page for the profile? You can copy and modify one of the man pages in `man/`.

Finally, please squash all the commits into one. Use `git commit --amend` for subsequent changes.

bolinches:
Hi its own sub-package totally fine.

bolinches:
Thanks a lot for all the help

man/tuned-profiles-spectrumscale-ece.7
```diff
@@ -0,0 +1,58 @@
+.\"/* 
+.\" * All rights reserved
+.\" * Copyright (C) 2015-2017 Red Hat, Inc.
+.\" * Authors: Jaroslav Škarvada
```
olysonek, line 4: Not true - you are the author ;).

man/tuned-profiles-spectrumscale-ece.7
```diff
@@ -0,0 +1,58 @@
...
+.BR tuned\-profiles\-atomic (7)
+.BR tuned\-profiles\-oracle (7)
+.BR tuned\-profiles\-realtime (7)
+.BR tuned\-profiles\-nfv\-host (7)
+.BR tuned\-profiles\-nfv\-guest (7)
+.BR tuned\-profiles\-cpu\-partitioning (7)
+.BR tuned\-profiles\-compat (7)
+.SH AUTHOR
+.nf
+Jaroslav Škarvada <jskarvad@redhat.com>
```
olysonek, line 58: You are the author.

bolinches:
Ok I updated the author on man page

olysonek:
We had an internal discussion, and we think the subpackage should be named 'tuned-profiles-ibm-spectrumscale'. What do you think? If you agree, could you add the subpackage to the tuned.spec file?

olysonek:
BTW, to get the profile to RHEL, a Bugzilla ticket will be needed.

man/tuned-profiles-spectrumscale-ece.7
bolinches: Updated, thanks

man/tuned-profiles-spectrumscale-ece.7
bolinches: Updated

bolinches:
Hi

It needs to have the ECE bit as it only applies to the ECE servers. Without going into the details it is a different product that "Spectrum Scale" and has different tunings. It would create confusion.

We might in the future look into a generic Spectrum Scale profile but it is not one fit all as it run on 3 different HW platforms (x86, ppc64/ppc64le, s390x). ECE only is supported on one which makes this unique for ECE setups but not Spectrum Scale

yarda:
> Hi
> 
> It needs to have the ECE bit as it only applies to the Spectrum Scale ECE servers. Without going into the details it is a different product that "Spectrum Scale" and has different tunings. It would create confusion.
> 
> We might in the future look into a generic Spectrum Scale profile but it is not one fit all as it run on 3 different HW platforms (x86_64, ppc64/ppc64le, s390x). ECE only is supported on one which makes this unique for Spectrum Scale ECE setups but not Spectrum Scale

The idea was to house anything Spectrum Scale related under the tuned-profiles-ibm-spectrumscale package (e.g. spectrumscale-ece profile inside it), but if you think it would cause confusion go ahead with the tuned-profiles-ibm-spectrumscale-ece package.

bolinches:
Ahh sorry misunderstood. It is ok if the package is named *spectrumscale* with no ECE as far for now the only profile listed there is spectrumscale-ece

bolinches:
Sorry for latest changes. I believe we have it stable now. Thanks

yarda:
@bolinches could you please squash the commits to one as requested by @olysonek earlier? Also could you please add the new subpackage definition to the .spec file? It should be easy - just copy and paste, if not I can add it myself.

bolinches:
> @bolinches could you please squash the commits to one as requested by @olysonek earlier? Also could you please add the new subpackage definition to the .spec file? It should be easy - just copy and paste, if not I can add it myself.

Hi it is squashed now. About the spec file I would appreciate if you do it. Thanks a lot

yarda:
@bolinches LGTM thanks. Please also open Red Hat bugzilla against Tuned component, becuse it needs to go through relengs. @olysonek could you approve or comment?

bolinches:
> @bolinches LGTM thanks. Please also open Red Hat bugzilla against Tuned component, becuse it needs to go through relengs. @olysonek could you approve or comment?

Yes I plan to wait to see this flow end ... see how it looks and then move to RHEL bugzilla
Thanks

man/tuned-profiles-spectrumscale-ece.7
```diff
@@ -0,0 +1,59 @@
+.\"/* 
+.\" * All rights reserved
+.\" * Copyright (C) 2015-2020 Red Hat, Inc.
```
olysonek, line 3: I thought you are from IBM ;-). Also, I'm not a lawyer but I think the year should be just 2020.

bolinches: I am not sure it is possible to copyright config file on a opensoruced GPL tool but does not matter I add those changes https://github.com/redhat-performance/tuned/commit/7ce1772a7887250e7e038723c04af1c3b0f08a6d

olysonek: I don't see why not. The file doesn't contain anything from Tuned. And so given that you are the author of the file, you (or more likely your employer, per your work contract) are the copyright holder. Which means that you can distribute it in any way you like, even under a non-free license. But by creating this pull request and licensing the file under GPL, you're partially giving up your rights as the copyright holder and you're saying that it can be distributed under the terms of the GPL. And that means that we can incorporate your file in the Tuned project.

Or that's my understanding of it :).

yarda:
Thanks, merging.

PR #265, opened 2020-05-05
network-latency: used unique cmdline suffix

yarda:
related: rhbz#1816168

### DIFF
--- a/profiles/network-latency/tuned.conf
+++ b/profiles/network-latency/tuned.conf
@@ -16,4 +16,4 @@ net.ipv4.tcp_fastopen=3
 kernel.numa_balancing=0
 
 [bootloader]
-cmdline=skew_tick=1
+cmdline_network_latency=skew_tick=1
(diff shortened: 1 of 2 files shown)


yarda:
OK, it raises question: do we need to explicitly sort it? Is there any benefit from the explicit sort? IMHO the underlying options implementation uses OrderedDict, just patching the bootloader plugin to use OrderededDict instead of the dict should keep the order.

yarda:
This is an attempt without the explicit sort. Any reason why the explicit sort is needed?

olysonek:
Indeed, OrderedDict is used now since commit 244917ec41676d7f1af720.

Regular dict was used back when the feature was implemented, which explains why I used the sort.
https://github.com/redhat-performance/tuned/blob/66d550b94d6e7198043ea0c8379fa330da947a81/tuned/profiles/unit.py#L14

In retrospect, I probably should have changed it to OrderedDict. But I think this was my very first patch and I didn't know better :).

Anyway... dropping the sort would be a behaviour change. The "benefit" of the sort is that it basically allows you to override the profile hierarchy. But I'm not sure that's a benefit at all. And having to keep the suffixes ordered is a pain. So personally, I would make the change and drop the sort.

We may want to extend our test coverage, though, to make sure it works correctly.

> just patching the bootloader plugin to use OrderededDict instead of the dict should keep the order

I'm not sure what you mean here. Is there a dict used somewhere? I didn't find it. I think dropping the sort is all that is needed.

olysonek:
Let me just quickly test it.

yarda:
Thanks, merging.

PR #270, opened 2020-05-24
Allow override of detected architecture

yarda:
There should be a way for administrators and testers to force loading of
architecture specific tuning. For example, if there is some specific tuning
for some platforms with specific CPUIDs and the vendor will release new
compatible platform with the new CPUID, there needs to be a way for
administrator to use the tuning on the new platform without waiting for the
Tuned profile update or without the need to customize the profile.

This commit adds two main config options:
uname_string
cpuinfo_string

If unset (commented), which is the default, runtime detection will be used,
i.e. 'uname' will be called and '/proc/cpuinfo' will be read. If set,
content of the variables will be used instead of the runtime detection.

For example by setting:
uname_string = aarch64

It's possible to pretend that Tuned is running on the 64-bit ARM.

### DIFF
--- a/tuned-main.conf
+++ b/tuned-main.conf
@@ -39,3 +39,13 @@ log_file_count = 2
 
 # Log file max size
 log_file_max_size = 1MB
+
+# Preset system uname string for architecture specific tuning.
+# It can be used to force tuning for specific architecture.
+# If commented, "uname" will be called to fill its content.
+# uname_string = x86_64
+
+# Preset system cpuinfo string for architecture specific tuning.
+# It can be used to force tuning for specific architecture.
+# If commented, "/proc/cpuinfo" will be read to fill its content.
+# cpuinfo_string = Intel
--- a/tuned/units/manager.py
+++ b/tuned/units/manager.py
@@ -6,6 +6,7 @@ import tuned.exceptions
 import tuned.logs
 import tuned.plugins.exceptions
 import tuned.consts as consts
+from tuned.utils.global_config import GlobalConfig
 from tuned.utils.commands import commands
 
 log = tuned.logs.get()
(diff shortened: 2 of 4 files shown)


tuned-main.conf
```diff
@@ -39,3 +39,13 @@ log_file_count = 2
 
 # Log file max size
 log_file_max_size = 1MB
+
+# Preset system uname string for architecture specific tuning.
+# It can be used for forcing tuning for specific architecure.
```
olysonek, line 44: s/for forcing/to force/
Typo: architecture

tuned-main.conf
```diff
@@ -39,3 +39,13 @@ log_file_count = 2
...
+# If commented "uname" will be called to fill its content.
```
olysonek, line 45: Add a comma after "If commented".

tuned-main.conf
```diff
@@ -39,3 +39,13 @@ log_file_count = 2
...
+# uname_string = x86_64
+
+# Preset system cpuinfo string for architecture specific tuning.
+# It can be used for forcing tuning for specific architecure.
+# If commented "/proc/cpuinfo" will be read to fill its content.
```
olysonek, line 50: Same thing in this paragraph.

tuned/units/manager.py
```diff
@@ -18,14 +19,15 @@ class Manager(object):
...
-			def_instance_priority, hardware_inventory):
+			def_instance_priority, hardware_inventory, config = None):
 		super(Manager, self).__init__()
 		self._plugins_repository = plugins_repository
 		self._monitors_repository = monitors_repository
 		self._def_instance_priority = def_instance_priority
 		self._hardware_inventory = hardware_inventory
 		self._instances = []
 		self._plugins = []
+		self._config = GlobalConfig() if config is None else config
```
olysonek, line 30: I recently realized that there's a shortcut for this. Use whichever you prefer. (Note that it's not entirely semantically equivalent, because there are other values that evaluate to False, but for our purposes it works.)
```
		self._config = config or GlobalConfig()
```

tuned-main.conf
olysonek: You missed this line.

olysonek: And the "architecure" vs. "architecture" typo on both lines.

yarda:
Thanks, merging.

PR #176, opened 2019-04-07
[WIP] Add a new file access API

olysonek:
Please do not merge this yet. It's only a request for comments and still work-in-progress at this point.

The first commit adds the new API. Its commit message describes the motivation for it and its advantages, and also includes some open questions (these are not that important to get right however, as this is internal API). The other commits make use of the new API.

@yarda @TomasKorbar What do you guys think? Is it ok? Can we make it better?

PR #272, opened 2020-05-28
Add support for a post-loaded profile

olysonek:
This is a minimal viable implementation of a post-loaded profile feature. It was implemented to satisfy the needs of the kernel_settings Ansible role.
https://github.com/linux-system-roles/kernel_settings

This feature allows us to apply a profile in a way that is somewhat separate from the way profiles from the `active_profile` file are applied. Namely, the post-loaded profile is always applied after all the profiles from the `active_profile` (or more precisely, the post-loaded profile is always put to the tail of the profile chain when merging). Also, the post-loaded profile setting is not affected by running `tuned-adm profile <some profile>` - the post-loaded profile remains applied.

The name of the post-loaded profile is stored in `/etc/tuned/post_loaded_profile`.

The lack of support for multiple post-loaded profiles was more or less an arbitrary decision. There are no problems with it, as far as I can see. However, there is currently no use case for multiple post-loaded profiles, so I decided to make things easier for our future selves - it's one less thing to worry about if we need to make changes to the profile loading logic. We can add support for multiple post-loaded profiles anytime.

Resolves: rhbz#1798183

### DIFF
--- a/tuned/utils/commands.py
+++ b/tuned/utils/commands.py
@@ -471,3 +471,25 @@ class commands:
 				f.write(mode + "\n")
 		except (OSError,IOError) as e:
 			raise TunedException("Failed to save profile mode: %s" % e.strerror)
+
+	def get_post_loaded_profile(self):
+		profile_name = ""
+		try:
+			with open(consts.POST_LOADED_PROFILE_FILE, "r") as f:
+				profile_name = f.read().strip()
+		except IOError as e:
+			if e.errno != errno.ENOENT:
+				raise TunedException("Failed to read the active post-loaded profile: %s" % e)
+		except (OSError, EOFError) as e:
+			raise TunedException("Failed to read the active post-loaded profile: %s" % e)
+		if profile_name == "":
+			profile_name = None
+		return profile_name
+
+	def save_post_loaded_profile(self, profile_name):
+		try:
+			with open(consts.POST_LOADED_PROFILE_FILE, "w") as f:
+				if profile_name is not None:
+					f.write(profile_name + "\n")
+		except (OSError,IOError) as e:
+			raise TunedException("Failed to save the active post-loaded profile: %s" % e.strerror)
--- a/Makefile
+++ b/Makefile
@@ -162,6 +162,7 @@ install: install-dirs
 	# None profile in the moment, autodetection will be used
 	echo -n > $(DESTDIR)$(SYSCONFDIR)/tuned/active_profile
 	echo -n > $(DESTDIR)$(SYSCONFDIR)/tuned/profile_mode
+	echo -n > $(DESTDIR)$(SYSCONFDIR)/tuned/post_loaded_profile
 	install -Dpm 0644 bootcmdline $(DESTDIR)$(SYSCONFDIR)/tuned/bootcmdline
 	install -Dpm 0644 modules.conf $(DESTDIR)$(SYSCONFDIR)/modprobe.d/tuned.conf
 
--- a/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/conflicting/tuned.conf
+++ b/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/conflicting/tuned.conf
@@ -1,0 +1,5 @@
+[main]
+summary=Profile conflicting with the parent profile
+
+[sysctl]
+vm.swappiness=10
--- a/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/main.fmf
+++ b/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/main.fmf
@@ -1,0 +1,12 @@
+summary: Test for BZ#1798183 (RFE support post-loaded profile)
+description: |
+    Bug summary: RFE: support post-loaded profile
+    Bugzilla link: https://bugzilla.redhat.com/show_bug.cgi?id=1798183
+contact:
+- Ondřej Lysoněk <olysonek@redhat.com>
+component:
+- tuned
+require:
+- tuned
+duration: 5m
+extra-task: /CoreOS/tuned/Regression/bz1798183-RFE-support-post-loaded-profile
--- a/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/parent-vars/tuned.conf
+++ b/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/parent-vars/tuned.conf
@@ -1,0 +1,8 @@
+[main]
+summary=Parent profile that defines a variable
+
+[variables]
+foo=12
+
+[sysctl]
+vm.swappiness=${foo}
--- a/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/parent/tuned.conf
+++ b/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/parent/tuned.conf
@@ -1,0 +1,5 @@
+[main]
+summary=Parent profile
+
+[sysctl]
+vm.swappiness=20
--- a/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/parent2/tuned.conf
+++ b/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/parent2/tuned.conf
@@ -1,0 +1,5 @@
+[main]
+summary=Different parent profile
+
+[sysctl]
+vm.swappiness=30
--- a/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/post-vars/tuned.conf
+++ b/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/post-vars/tuned.conf
@@ -1,0 +1,5 @@
+[main]
+summary=Post-loaded profile that uses variables from the regular active profile
+
+[sysctl]
+vm.dirty_ratio=${foo}
--- a/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/post/tuned.conf
+++ b/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/post/tuned.conf
@@ -1,0 +1,5 @@
+[main]
+summary=Post-loaded profile
+
+[sysctl]
+vm.dirty_ratio=8
--- a/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/post2/tuned.conf
+++ b/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/post2/tuned.conf
@@ -1,0 +1,5 @@
+[main]
+summary=Second version of the post-loaded profile
+
+[sysctl]
+vm.dirty_ratio=7
--- a/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/runtest.sh
+++ b/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/runtest.sh
@@ -1,0 +1,279 @@
+#!/bin/bash
+# vim: dict+=/usr/share/beakerlib/dictionary.vim cpt=.,w,b,u,t,i,k
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   runtest.sh of /CoreOS/tuned/Regression/bz1798183-RFE-support-post-loaded-profile
+#   Description: Test for BZ#1798183 (RFE support post-loaded profile)
+#   Author: Ondřej Lysoněk <olysonek@redhat.com>
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Copyright (c) 2020 Red Hat, Inc.
+#
+#   This program is free software: you can redistribute it and/or
+#   modify it under the terms of the GNU General Public License as
+#   published by the Free Software Foundation, either version 2 of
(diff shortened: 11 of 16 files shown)


olysonek:
TLDR; the following is a list of requirements for the feature and design decisions that I had to make.

The requirements are based on my understanding of the kernel_settings role. @richm Please let me know if they are not accurate.

Also, please let me know if you disagree with any of my decisions or if there's something else that I should consider.

All the requirements and decisions (except for 8.) are tested by the included `runtest.sh`. For easy reference, the test cases are written in order.

1. *Requirement*: After running 'tuned-adm profile <profile name>', the post-loaded profile must remain applied.
2. *Requirement*: Settings from the post-loaded profile must override settings from the profiles set in the active_profile file if there are conflicts.
3. *Requirement*: 'tuned-adm verify' must not fail just because there are conflicts between the active_profile and the post-loaded profile. (The consequence of this is that the post-loaded profile must be merged with the active_profile - meaning, the post-loaded profile is not applied separately.)
4. What should happen on 'tuned-adm off' when a post-loaded profile is applied?
   a) deactivate and clear the post-loaded profile setting on disk
   b) leave the post-loaded profile applied and keep the setting on disk
   c) add a cmdline option to modify the behaviour. Which behaviour should be the default?
5. (Somewhat related to 4.) What should a starting-up Tuned process do if the active_profile setting on disk is empty and the post-loaded profile setting is non-empty?
   a) complain about illegal configuration
   b) apply the post-loaded profile only
   c) not apply any profile
6. 'tuned-adm active' should show the post-loaded profile. But how should it be displayed?
   a) the post-loaded profile should be shown as if it were a regular profile
   b) the post-loaded profile should get shown separately
   c) the post-loaded profile should be shown in both ways at the same time, e.g.:
```
Current active profile: balanced kernel_settings
Post-loaded profile: kernel_settings
```
7. What should 'tuned -p <profile_name>' do?
   a) clear the post-loaded profile setting on disk, set only <profile_name> as the active profile and apply the profile
   b) keep the post-loaded profile setting, set <profile_name> as the active profile and apply both <profile_name> and the post-loaded profile
   c) keep the post-loaded profile setting set on disk, set <profile name> as the active profile and apply only <profile name>
8. (There's no test case for this, but it's implicitly tested by other test cases combined with the fact that the implementation doesn't modify the tuned-adm code.) When Tuned applies a profile, it sends a "profile_changed" signal to DBus. This signal contains the list of applied profiles. Which profiles should be listed in the in the signal when a post-loaded profile is applied?
   a) List all the applied profiles, including the post-loaded one. This would require changes in tuned-adm and possibly other programs that may consume the signal (e.g. Cockpit). ('tuned-adm profile' consumes the signal so that it knows when the profile is fully applied.)
   b) Always list only the active_profile. This will alter the meaning of the signal. If we choose this option, we may want to implement another signal that would contain all the applied profiles, including the post-loaded one.
9. (Related to 1.) On 'tuned-adm profile <profile_name>, Tuned should
   a) reuse the already loaded post-loaded profile name
   b) reload the post-loaded profile name from disk
10. (Related to 1.) Tuned has a habit of writing to the active_profile file after profiles are applied. Should it always write out both the active_profile and post_loaded_profile, or only the one that changed (e.g. due to 'tuned-adm profile')?
    a) write out both. This means that if we change the post_loaded_profile file on disk and then do 'tuned-adm profile', the post_loaded profile will be overwritten.
    b) write out only the one that changed. This might change the semantics of the files slightly - they no longer reflect the currently applied profiles.
11. Should variables be shared between the 'active_profile' profile and the post-loaded profile?
    a) yes
    b) no
   
My choices and the justification:
4. a) seems to be the best choice to preserve the semantics
5. b) so that the kernel_settings profile can be the only applied
   profile.
6. a) seems to be the best choice to preserve the semantics. We can later implement c) if needed.
7. b) is the easiest to implement. No reason other than that.
8. b) because it doesn't break other programs
9. a) because the requirement is that 'tuned-adm profile' does not change the post-loaded profile. Option a) is more in line with that requirement.
10. a) seems to be the best choice to preserve the semantics of the active_profile file. Also, it's the easiest to implement.
11. I think ideally, the variables should not be shared, because (at least for the use case of kernel_settings) the profile hierarchies are unrelated. However, separating the variables could be tricky to implement, so I choose a), unless someone has a strong opinion on this. We could add a config option later that would modify the behaviour.

olysonek:
There were some syntax errors in my previous comment that caused it to display incorrectly. Sorry about that. Hopefully fixed now.

olysonek:
@richm Please review and test.

olysonek:
>  8. b) because it doesn't break other programs

It seems I've implemented something different. I'll fix it.

olysonek:
> > 8. b) because it doesn't break other programs
> 
> It seems I've implemented something different. I'll fix it.

Never mind. I did actually implement 8. b). I just left some confusing dead code lying around.

I added a test case for 8. b) now.

richm:
I'm running tests/beakerlib/bz1798183-RFE-support-post-loaded-profile on Fedora 31.  It passes until it gets to here:
```
:: [ 15:33:18 ] :: [   LOG    ] :: rlServiceStart: Service tuned started successfully
:: [ 15:33:18 ] :: [   PASS   ] :: Command 'rlServiceStart tuned' (Expected 0, got 0)
:: [ 15:33:18 ] :: [  BEGIN   ] :: Running 'tuned-adm off'
:: [ 15:33:18 ] :: [   PASS   ] :: Command 'tuned-adm off' (Expected 0, got 0)
:: [ 15:33:18 ] :: [   PASS   ] :: Check the output of tuned-adm active (Assert: 'No current active profile.' should equal 'No current active profile.')
:: [ 15:33:18 ] :: [   FAIL   ] :: Check that swappiness has not been changed (Assert: '60' should equal '30')
:: [ 15:33:18 ] :: [   FAIL   ] :: Check that dirty ratio has not been changed (Assert: '20' should equal '30')
::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
::   Duration: 2s
::   Assertions: 5 good, 2 bad
::   RESULT: FAIL (Check that 'tuned-adm off' causes Tuned to clear the post-loaded profile)
```
not sure if this is a false negative - I'm struggling with the fact that the test restarts tuned several times, and I'm hitting the Fedora 31 systemd service restart limit of 5 times in 10s.  Still investigating, but can you confirm the test works on some version of rhel?

olysonek:
@richm Please try the current version. All of the problems you mentioned should be resolved now. If not, please let me know.

I haven't tested on RHEL yet, but it seems to work on CentOS 8.

olysonek:
@richm You can also find the test results here:
https://artifacts-testing-farm.apps.ci.centos.org/05ed77a6-1171-4213-aae8-3862a6780891/ci/execute.log

There are also results for a couple of other tests in there, so search for `bz1798183-RFE-support-post-loaded-profile`.

richm:
> 10  a) write out both. This means that if we change the post_loaded_profile file on disk and then do tuned-adm profile, the post_loaded_profile file will be overwritten.

So how to update the post_loaded_profile safely?  Is it this?

1) shutdown tuned e.g. systemctl stop tuned
2) edit /etc/tuned/post_loaded_profile
3) start tuned

?

olysonek:
> > 10  a) write out both. This means that if we change the post_loaded_profile file on disk and then do tuned-adm profile, the post_loaded_profile file will be overwritten.
> 
> So how to update the post_loaded_profile safely? Is it this?
>     1. shutdown tuned e.g. systemctl stop tuned
>     2. edit /etc/tuned/post_loaded_profile
>     3. start tuned

Either that, or just edit /etc/tuned/post_loaded_profile and then restart Tuned. That's what I'm doing in the tests.

In the future we could add a new tuned-adm command for changing  the post-loaded profile.

olysonek:
> > > 10  a) write out both. This means that if we change the post_loaded_profile file on disk and then do tuned-adm profile, the post_loaded_profile file will be overwritten.
> > 
> > So how to update the post_loaded_profile safely? Is it this?
> > 1. shutdown tuned e.g. systemctl stop tuned
> > 2. edit /etc/tuned/post_loaded_profile
> > 3. start tuned
> 
> Either that, or just edit /etc/tuned/post_loaded_profile and then restart Tuned. That's what I'm doing in the tests.

You could also send the HUP signal to the Tuned process after writing the file. I'll write a test for this.

yarda:
> 
> My choices and the justification:
> 4. a) seems to be the best choice to preserve the semantics
> 5. b) so that the kernel_settings profile can be the only applied
> profile.
> 6. a) seems to be the best choice to preserve the semantics. We can later implement c) if needed.
> 7. b) is the easiest to implement. No reason other than that.
> 8. b) because it doesn't break other programs
> 9. a) because the requirement is that `tuned-adm profile` does not change the post-loaded profile. Option a) is more in line with that requirement.
> 10. a) seems to be the best choice to preserve the semantics of the `active_profile` file. Also, it's the easiest to implement.
> 11. I think ideally, the variables should not be shared, because (at least for the use case of kernel_settings) the profile hierarchies are unrelated. However, separating the variables could be tricky to implement, so I choose a), unless someone has a strong opinion on this. We could add a config option later that would modify the behaviour.
> 
> EDIT: 8. b) now has a test case.

I mostly agree with your choices, except:
6) I think c) is better, because user should be informed that it's some special profile and they can then start searching what the post-loaded profile is about. Otherwise they could be confused and could spent some hard time trying to get rid of it :)
9) I think the post-loaded profile should be reloaded from the disk as all other profiles are, otherwise users could be surprised of this exceptional behaviour.

yarda:
> > My choices and the justification:
> > 4. a) seems to be the best choice to preserve the semantics
> > 5. b) so that the kernel_settings profile can be the only applied
> > profile.
> > 6. a) seems to be the best choice to preserve the semantics. We can later implement c) if needed.
> > 7. b) is the easiest to implement. No reason other than that.
> > 8. b) because it doesn't break other programs
> > 9. a) because the requirement is that `tuned-adm profile` does not change the post-loaded profile. Option a) is more in line with that requirement.
> > 10. a) seems to be the best choice to preserve the semantics of the `active_profile` file. Also, it's the easiest to implement.
> > 11. I think ideally, the variables should not be shared, because (at least for the use case of kernel_settings) the profile hierarchies are unrelated. However, separating the variables could be tricky to implement, so I choose a), unless someone has a strong opinion on this. We could add a config option later that would modify the behaviour.
> > EDIT: 8. b) now has a test case.
> 
> I mostly agree with your choices, except:
> 6) I think c) is better, because user should be informed that it's some special profile and they can then start searching what the post-loaded profile is about. Otherwise they could be confused and could spent some hard time trying to get rid of it :)
> 9) I think the post-loaded profile should be reloaded from the disk as all other profiles are, otherwise users could be surprised of this exceptional behaviour.

9) a) is not blocker for me (if properly documented), but I think 9) b) would simplify the implementation.

tuned/utils/commands.py
```diff
@@ -471,3 +471,25 @@ def save_active_profile(self, profile_name, manual):
 				f.write(mode + "\n")
 		except (OSError,IOError) as e:
 			raise TunedException("Failed to save profile mode: %s" % e.strerror)
+
```
yarda, line 474: Is there any reason for it to be in the 'commands'? I take it as a hack for storage of utilities/helpers which could be useful in other places of the Tuned code/plugins and these two functions seems to be called just from the daemon.py.

olysonek:
>  9) I think the post-loaded profile should be reloaded from the disk as all other profiles are, otherwise users could be surprised of this exceptional behaviour.

Let me clarify that 9) is about the profile *name* (i.e. the contents of /etc/tuned/post_loaded_profile), not the profile *contents* (i.e. the contents of /etc/tuned/foo/tuned.conf). Did you get that right?

Currently, the name of the post-loaded profile is not reloaded, but the contents are reloaded.

The `switch_profile` method of `Controller` doesn't read any profile names from disk either - it just uses the profile names that were provided.

olysonek:
>  6) I think c) is better, because user should be informed that it's some special profile and they can then start searching what the post-loaded profile is about. Otherwise they could be confused and could spent some hard time trying to get rid of it :)

Yes, I agree. I'll implement c) as soon as I can. By implementing a) first, I was just trying to save some time.

yarda:
> > 1. I think the post-loaded profile should be reloaded from the disk as all other profiles are, otherwise users could be surprised of this exceptional behaviour.
> 
> Let me clarify that 9) is about the profile _name_ (i.e. the contents of /etc/tuned/post_loaded_profile), not the profile _contents_ (i.e. the contents of /etc/tuned/foo/tuned.conf). Did you get that right?
> 
> Currently, the name of the post-loaded profile is not reloaded, but the contents are reloaded.
> 
> The `switch_profile` method of `Controller` doesn't read any profile names from disk either - it just uses the profile names that were provided.

Sorry, I missed that. So it's probably OK then.

tuned/utils/commands.py
olysonek: I wanted it to be in the same place as `save_active_profile`, etc., because those methods are related. In the future, I certainly want to move them some place else. My PR #176 already does that - later I'll update that PR to move the new methods too.

olysonek:
> Sorry, I missed that. So it's probably OK then.

No problem. I guess deciding 9) might be a no-brainer, but I wanted to mention it, because I got bit by it when writing my test.

tuned/utils/commands.py
olysonek: Also, in the future they will be called from admin.py too - I'll need them to implement 6) c) and things like `tuned-adm change-post-loaded-profile`.

yarda:
Thanks, LGTM.

yarda:
Thanks, merging.

PR #273, opened 2020-05-28
Added new profile optimize-serial-console

yarda:
Profile which tunes down I/O activity to the serial console by reducing the
printk value. This should make the serial console more responsive.
This profile is intended to be used as an overlay on other
profiles (e.g. throughput-performance profile), example:
```
# tuned-adm profile throughput-performance optimize-serial-console
```

Also minor fixes to the man page quoting.

Resolves: rhbz#1840689


yarda:
Thanks, merging.

PR #274, opened 2020-06-02
Add irqbalance plugin

olysonek:
The irqbalance plugin is a replacement for the bash functions `irqbalance_banned_cpus_setup` and `irqbalance_banned_cpus_clear`. It configures the banned CPUs in /etc/sysconfig/irqbalance. Then it restarts irqbalance (if and only if it was previously running).

The banned CPUs can be specified as a CPU list. For example:
```
[irqbalance]
banned_cpus=2,4,9-13
```

The profiles `cpu-partitioning` and `realtime` have also been updated to use the new plugin.

Resolves: rhbz#1784645

### DIFF
--- a/tuned/plugins/plugin_irqbalance.py
+++ b/tuned/plugins/plugin_irqbalance.py
@@ -1,0 +1,110 @@
+from . import base
+from .decorators import command_custom
+from tuned import consts
+import tuned.logs
+import errno
+import perf
+import re
+
+log = tuned.logs.get()
+
+class IrqbalancePlugin(base.Plugin):
+	"""
+	Plugin for irqbalance settings management.
+	"""
+
+	def __init__(self, *args, **kwargs):
+		super(IrqbalancePlugin, self).__init__(*args, **kwargs)
+		self._cpus = perf.cpu_map()
+
+	def _instance_init(self, instance):
+		instance._has_dynamic_tuning = False
+		instance._has_static_tuning = True
+
+	def _instance_cleanup(self, instance):
+		pass
+
+	@classmethod
+	def _get_config_options(cls):
+		return {
+			"banned_cpus": None,
+		}
+
+	def _read_irqbalance_sysconfig(self):
+		try:
+			with open(consts.IRQBALANCE_SYSCONFIG_FILE, "r") as f:
+				return f.read()
+		except IOError as e:
+			if e.errno == errno.ENOENT:
+				log.warn("irqbalance sysconfig file is missing. Is irqbalance installed?")
+			else:
+				log.error("Failed to read irqbalance sysconfig file: %s" % e)
+			return None
+
+	def _write_irqbalance_sysconfig(self, content):
+		try:
+			with open(consts.IRQBALANCE_SYSCONFIG_FILE, "w") as f:
+				f.write(content)
+			return True
+		except IOError as e:
+			log.error("Failed to write irqbalance sysconfig file: %s" % e)
+			return False
+
+	def _write_banned_cpus(self, sysconfig, banned_cpumask):
+		return sysconfig + "IRQBALANCE_BANNED_CPUS=%s\n" % banned_cpumask
+
+	def _clear_banned_cpus(self, sysconfig):
+		lines = []
+		for line in sysconfig.split("\n"):
+			if not re.match(r"\s*IRQBALANCE_BANNED_CPUS=", line):
+				lines.append(line)
+		return "\n".join(lines)
+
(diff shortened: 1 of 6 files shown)


tuned/plugins/plugin_irqbalance.py
```diff
@@ -0,0 +1,108 @@
...
+
+class IrqbalancePlugin(base.Plugin):
+	"""
+	Plugin for irqbalance settings management.
+	"""
+
+	def __init__(self, *args, **kwargs):
+		super(IrqbalancePlugin, self).__init__(*args, **kwargs)
+		self._cpus = perf.cpu_map()
+		self._irqbalance_sysconfig = "/etc/sysconfig/irqbalance"
```
jmencak, line 17: Make into a constant in a similar way such as `consts.GRUB2_CFG_FILES` for example?

olysonek: It's a good suggestion, thanks. However, given that this is a plugin-specific thing, I think it's cleaner to keep it in the plugin. But I should probably make it into a class variable. @yarda What do you think?

olysonek:
>  Is this really WIP?

I just need to change our profiles to use the plugin and verify something.

I also wanted to implement verification, but I think that can wait, given that it's not currently supported in the scripts either.

olysonek:
Profiles updated.

I just need to make sure it will work on machines with more than 32 cores and then this will be good to go. Currently I'm waiting for HW.

jmencak:
> Profiles updated.
Great!

> 
> I just need to make sure it will work on machines with more than 32 cores and then this will be good to go. Currently I'm waiting for HW.
I've done that in the past with QEMU, simulated a 64 core box, worked fine on Lenovo x1 6th gen. ;)

tuned/plugins/plugin_irqbalance.py
yarda: It could be class variable, but I would initialize it from the consts. It's not optimal, but we currently do it this way, e.g.:
```
...
# modules plugin configuration
MODULES_FILE = "/etc/modprobe.d/tuned.conf"

# systemd plugin configuration
SYSTEMD_SYSTEM_CONF_FILE = "/etc/systemd/system.conf"
...
```
```

**Comment by olysonek** [comment]
> > I just need to make sure it will work on machines with more than 32 cores and then this will be good to go. Currently I'm waiting for HW.
> 
> I've done that in the past with QEMU, simulated a 64 core box, worked fine on Lenovo x1 6th gen. ;)

Good point. I forgot about this possibility.

File: tuned/plugins/plugin_irqbalance.py [hunk 2]
```diff
@@ -0,0 +1,108 @@
...
+			log.error("Failed to write irqbalance sysconfig file: %s" % e)
+			return False
+
+	def _write_banned_cpus(self, sysconfig, banned_cpumask):
+		return sysconfig + "IRQBALANCE_BANNED_CPUS=%s\n" % banned_cpumask
+
+	def _clear_banned_cpus(self, sysconfig):
+		lines = []
+		for line in sysconfig.split("\n"):
+			if not line.startswith("IRQBALANCE_BANNED_CPUS="):
```
yarda, line 57: It's better to use regex or lstrip. It's enviroment variable and it can begin with whitespaces.

tuned/plugins/plugin_irqbalance.py
```diff
@@ -0,0 +1,108 @@
...
+		return "\n".join(lines)
+
+	def _restart_irqbalance(self):
+		# Exit code 5 means unit not found (see 'EXIT_NOTINSTALLED' in
+		# systemd.exec(5))
+		retcode, out = self._cmd.execute(
+			["systemctl", "try-restart", "irqbalance"],
+			no_errors=[5])
+		if retcode != 0:
+			log.warn("Failed to restart irqbalance. It is installed?")
```
yarda, line 68: Typo: "Is it installed?"

tuned/plugins/plugin_irqbalance.py
olysonek: Thanks. I've changed it to use consts now.

olysonek:
Good news, everyone. It seems to work on machines with more than 32 CPUs too (48 to be specific).

I now consider this finished.

jmencak:
> LGTM, thanks. @jmencak do you have more comments or could you approve?

LGTM.  Both cpu-partitioning/realtime are now using the functionality.  Haven't tested again though.

yarda:
Thanks, merging.

PR #275, opened 2020-06-02
scheduler: add support for cgroups

yarda:
This is an alpha version, expect bugs :) It needs more love :)

This commit adds cgroup v1 support to the scheduler plugin. It seems that
cgroup v1 is the default on RHEL-8. Systemd uses it there, so we have to start
with the cgroup v1. On Fedora, systemd has been switched
to the cgroup v2 (tested on f31), but it's still possible to intermix there
with the cgroup v1. We can add cgroup v2 support later (the logic of the
extended syntax will not change).

This commit extends the syntax the following way:
```
[scheduler]
cgroup_init=1
cgroup_mount_point=/cgroup/cpuset
cgroup=group
cgroup.group1=2
cgroup.group2=0,2

group.ksoftirqd=0:f:2:cgroup.group1:ksoftirqd.*

ps_blacklist=ksoftirqd.*;rcuc.*;rcub.*;ktimersoftd.*

isolated_cores=1
```
--
Legend:

'cgroup_init' if set to '1' (or similar boolean) means that Tuned
will create and remove the cgroup mountpoint and all related directories
by itself. If set to '0' (the default) the cgroups needs to be preset
by some different tool or by hand.

'cgroup_mount_point' is where to mount the cgroup FS or where Tuned
expects it to be mounted. If unset '/cgroup/cpuset' is expected.

'cgroup' is the cgroup name used for the 'isolated_cores' functionality.
For example here if the system has 4 CPUs, 'isolated_cores=1' means that
all threads (except of the blacklisted ones by the 'ps_blacklist' regexes)
will be moved to the CPU cores 0,2,3. It will be done the way that cgroup
'group' will be set to use the affinity 0,2-3 (by the 'cpuset.cpus' control
file) and all matching threads will be moved to this cgroup. If 'cgroup'
is unset, classic cpuset affinity will be used to achieve the goal.

'cgroup.CGROUP_NAME' defines affinities for arbitrary cgroups.
Even hierarchic cgroups can be used, but the hieararchy needs to be
specified in the correct order. Also Tuned doesn't do any sanity checks
here (except that it forces the cgroup to be under the mountpoint).
In the example it defines 'group1' with the affinity set to the CPU 2
and 'group2' with the affinity set to CPUs 0, 2.

'group.' its syntax has been extended, so it's now possible to use
e.g. 'cgroup.group1' instead of the hex affinity and the matching
processes will be moved to the 'group1'. It's also possible to use
cgroups which hasn't been defined by the 'cgroup.' option described
above (e.g. cgroups not managed by Tuned).

Resolves: rhbz#1784648

### DIFF
--- a/tuned/consts.py
+++ b/tuned/consts.py
@@ -41,6 +41,13 @@ MACHINE_ID_FILE = "/etc/machine-id"
 KERNEL_UPDATE_HOOK_FILE = "/usr/lib/kernel/install.d/92-tuned.install"
 BLS_ENTRIES_PATH = "/boot/loader/entries"
 
+# scheduler plugin configuration
+# how many times retry to move tasks to parent cgroup on cgroup cleanup
+CGROUP_CLEANUP_TASKS_RETRY = 10
+PROCFS_MOUNT_POINT = "/proc"
+DEF_CGROUP_MOUNT_POINT = "/sys/fs/cgroup/cpuset"
+DEF_CGROUP_MODE = 0o770
+
 # modules plugin configuration
 MODULES_FILE = "/etc/modprobe.d/tuned.conf"
 
--- a/tuned/plugins/plugin_scheduler.py
+++ b/tuned/plugins/plugin_scheduler.py
@@ -15,17 +15,20 @@ import procfs
 import schedutils
 from tuned.utils.commands import commands
 import errno
+import os
+import collections
 
 log = tuned.logs.get()
 
 class SchedulerParams(object):
 	def __init__(self, cmd, cmdline = None, scheduler = None,
-			priority = None, affinity = None):
+			priority = None, affinity = None, cgroup = None):
 		self._cmd = cmd
 		self.cmdline = cmdline
 		self.scheduler = scheduler
 		self.priority = priority
 		self.affinity = affinity
+		self.cgroup = cgroup
 
 	@property
 	def affinity(self):
@@ -98,7 +101,17 @@ class SchedulerPlugin(base.Plugin):
 			self._scheduler_original = {}
 			self._storage.unset(self._scheduler_storage_key)
 
+		self._cgroups_original_affinity = dict()
+
+		# calculated by isolated_cores setter
+		self._affinity = None
+
+		self._cgroup = None
+		self._cgroups = collections.OrderedDict([(self._sanitize_cgroup_path(option[7:]), self._variables.expand(affinity))
+			for option, affinity in instance.options.items() if option[:7] == "cgroup." and len(option) > 7])
+
 		instance._scheduler = instance.options
+
 		for k in instance._scheduler:
 			instance._scheduler[k] = self._variables.expand(instance._scheduler[k])
 		if self._cmd.get_bool(instance._scheduler.get("runtime", 1)) == "0":
@@ -126,10 +139,17 @@ class SchedulerPlugin(base.Plugin):
 	@classmethod
 	def _get_config_options(cls):
 		return {
-			"isolated_cores":       None,
+			"isolated_cores": None,
+			"cgroup_mount_point": consts.DEF_CGROUP_MOUNT_POINT,
+			"cgroup_mount_point_init": False,
+			"cgroup_groups_init": True,
+			"cgroup_for_isolated_cores": None,
 			"ps_whitelist": None,
 			"ps_blacklist": None,
 		}
+
+	def _sanitize_cgroup_path(self, value):
+		return str(value).replace(".", "/") if value is not None else None
 
 	# Raises OSError, IOError
 	def _get_cmdline(self, process):
@@ -273,28 +293,66 @@ class SchedulerPlugin(base.Plugin):
 						% (pid, e))
 		return cont
 
-	def _store_orig_process_affinity(self, pid, affinity):
+	def _is_cgroup_affinity(self, affinity):
+		return str(affinity)[:7] == "cgroup."
+
+	def _store_orig_process_affinity(self, pid, affinity, is_cgroup = False):
 		try:
 			params = self._scheduler_original[pid]
 		except KeyError:
 			params = SchedulerParams(self._cmd)
 			self._scheduler_original[pid] = params
-		if params.affinity is None:
-			params.affinity = affinity
+		if params.affinity is None and params.cgroup is None:
+			if is_cgroup:
+				params.cgroup = affinity
+			else:
+				params.affinity = affinity
+
+	def _get_cgroup_affinity(self, pid):
+		# we cannot use procfs, because it uses comma ',' delimiter which
+		# can be ambiguous
+		for l in self._cmd.read_file("%s/%s/%s" % (consts.PROCFS_MOUNT_POINT, str(pid), "cgroup"), no_error = True).split("\n"):
+			try:
+				cgroup = l.split(":cpuset:")[1][1:]
+				return cgroup if cgroup != "" else "/"
+			except IndexError:
+				pass
+		return "/"
+
+	# it can be arbitrary cgroup even cgroup we didn't set, but it needs to be
+	# under "cgroup_mount_point"
+	def _set_cgroup(self, pid, cgroup):
+		cgroup = self._sanitize_cgroup_path(cgroup)
+		path = self._cgroup_mount_point
+		if cgroup != "/":
+			path = "%s/%s" % (path, cgroup)
+		self._cmd.write_to_file("%s/tasks" % path, str(pid), no_error = True)
+
+	def _parse_cgroup_affinity(self, cgroup):
+		# "cgroup.CGROUP"
+		cgroup = cgroup[7:]
+		# this should be faster than string comparison
+		is_cgroup = not isinstance(cgroup, list) and len(cgroup) > 0
+		return is_cgroup, cgroup
 
 	def _tune_process_affinity(self, pid, affinity, intersect = False):
 		cont = True
 		if affinity is None:
 			return cont
 		try:
-			prev_affinity = self._get_affinity(pid)
-			if intersect:
-				affinity = self._get_intersect_affinity(
-						prev_affinity, affinity,
-						affinity)
-			self._set_affinity(pid, affinity)
+			(is_cgroup, cgroup) = self._parse_cgroup_affinity(affinity)
+			if is_cgroup:
+				prev_affinity = self._get_cgroup_affinity(pid)
+				self._set_cgroup(pid, cgroup)
+			else:
+				prev_affinity = self._get_affinity(pid)
+				if intersect:
+					affinity = self._get_intersect_affinity(
+							prev_affinity, affinity,
+							affinity)
+				self._set_affinity(pid, affinity)
 			self._store_orig_process_affinity(pid,
-					prev_affinity)
+					prev_affinity, is_cgroup)
 		except (SystemError, OSError) as e:
 			if hasattr(e, "errno") and e.errno == errno.ESRCH:
 				log.debug("Failed to read affinity of PID %d, the task vanished."
@@ -335,6 +393,8 @@ class SchedulerPlugin(base.Plugin):
 	def _convert_affinity(self, str_affinity):
 		if str_affinity == "*":
 			affinity = None
+		elif self._is_cgroup_affinity(str_affinity):
+			affinity = str_affinity
 		else:
 			affinity = self._cmd.hex2cpulist(str_affinity)
 			if not affinity:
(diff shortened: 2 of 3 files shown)


tuned/plugins/plugin_scheduler.py
```diff
@@ -273,28 +292,65 @@ def _tune_process_rt(self, pid, sched, prio):
...
 		if affinity is None:
 			return cont
 		try:
-			prev_affinity = self._get_affinity(pid)
-			if intersect:
-				affinity = self._get_intersect_affinity(
-						prev_affinity, affinity,
-						affinity)
-			self._set_affinity(pid, affinity)
+			(is_cgroup, cgroup) = self._parse_cgroup_affinity(affinity)
```
olysonek, line 341: If I'm reading the code right, then `affinity` can be a list of CPUs. And if that list is has more than 7 elements, then this will return a false positive.

tuned/plugins/plugin_scheduler.py
```diff
@@ -350,8 +408,91 @@ def _convert_sched_cfg(self, vals):
 		affinity = self._convert_affinity(affinity)
 		return (rule_prio, scheduler, priority, affinity, regex)
 
+	def _cgroup_initialize_one(self, cgroup):
+		path = "%s/%s" % (self._cgroup_mount_point, cgroup)
+		try:
+			os.mkdir(path, consts.DEF_CGROUP_MODE)
+		except OSError as e:
```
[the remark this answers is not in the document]
yarda: Updated to print the underlying error.

tuned/plugins/plugin_scheduler.py
```diff
@@ -350,8 +408,91 @@ def _convert_sched_cfg(self, vals):
...
+		if (not self._cmd.write_to_file("%s/%s" % (path, "cpuset.cpus"), \
+				self._cmd.read_file("%s/%s" % (self._cgroup_mount_point, "cpuset.cpus"), no_error = True), \
+				no_error = True)):
+					log.error("Unable to initialize 'cpuset.cpus' for cgroup '%s'" % path)
+
+	def _cgroup_initialize(self):
+		log.debug("Initializing cgroups settings")
+		try:
+			os.makedirs(self._cgroup_mount_point, consts.DEF_CGROUP_MODE)
+		except OSError as e:
```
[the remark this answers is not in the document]
yarda: Also updated.

tuned/plugins/plugin_scheduler.py
```diff
@@ -350,8 +408,91 @@ def _convert_sched_cfg(self, vals):
...
+			log.error("Unable to mount '%s'" % self._cgroup_mount_point)
+		if self._cgroup is not None and not self._cgroup in self._cgroups:
+			self._cgroup_initialize_one(self._cgroup)
+		for cg in self._cgroups:
+			self._cgroup_initialize_one(cg)
+
+	def _remove_dir(self, cgroup):
+		try:
+			os.rmdir(cgroup)
+		except OSError as e:
```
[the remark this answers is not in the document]
yarda: Also updated.

tuned/consts.py
```diff
@@ -41,6 +41,13 @@
 KERNEL_UPDATE_HOOK_FILE = "/usr/lib/kernel/install.d/92-tuned.install"
 BLS_ENTRIES_PATH = "/boot/loader/entries"
 
+# scheduler plugin configuration
+# how many times retry to move tasks to parent cgroup on cgroup cleanup
+CGROUP_CLEANUP_TASKS_RETRY = 10
+PROCFS_MOUNT_POINT = "/proc"
+DEF_CGROUP_MOUNT_POINT = "/cgroup/cpuset"
```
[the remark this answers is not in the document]
yarda: Well, I wasn't sure here. It's not FHS compliant, but it was mentioned in the older kernel docs. The /sys/fs/cgroup/cpuset would work great with cgroup v1 (e.g. on RHEL-8), because /sys/fs/cgroup is tmpfs mount point there. But it wouldn't work so great with the cgroup v2 enabled systemd (e.g. on Fedora when intermixing with cgroup v2). There /sys/fs/cgroup/ is cgroup2 FS, so the following would happen:
- cpuset group would be created after the mkdir call and all the special files would be auto populated by the kernel
- then we would mount over it the cgroup v1 FS, i.e. we would loose the previously created v2 group
In the end it would work, but it is not nice.

However, we are primary targeting RHEL-8 with this feature so I will probably change the default to integrate easily with the RHEL-8.

tuned/plugins/plugin_scheduler.py
```diff
@@ -98,7 +101,17 @@ def _instance_init(self, instance):
 			self._scheduler_original = {}
 			self._storage.unset(self._scheduler_storage_key)
 
+		self._cgroups_original_affinity = dict()
+
+		# calculated by isolated_cores setter
+		self._affinity = None
+
+		self._cgroup = None
```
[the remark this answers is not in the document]
yarda: Well, it's in fact the opposite, cgroup that will hold tasks on the non-isolated cores. I agree that the current name is not good, but at the moment I am unable to come up with anything better.

tuned/plugins/plugin_scheduler.py
yarda: Good catch, hopefully fixed.

tuned/consts.py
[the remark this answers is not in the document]
yarda: On RHEL-8 /sys/fs/cgroup is read-only, so "cgroup_init" has to be set to False, i.e. with the default settings we rely on systemd to preset it.

yarda:
I updated the interface a bit:
'cgroup_mount_point' is where to mount the cgroup FS or where Tuned expects it to be mounted. If unset '/sys/fs/cgroup/cpuset' is expected.

'cgroup_groups_init' if set to '1' (the default) it means that Tuned will create (and remove) all cgroups defined with the 'cgroup*' options. If set to '0' (or similar boolean) the cgroups need to be preset by some different tool or by hand.

'cgroup_mount_point_init' if set to '1' (or similar boolean) it means that Tuned will create (and remove) the cgroup mountpoint. It implies 'cgroup_groups_init = 1'. If set to '0' (the default) the cgroups mount point needs to be preset by some different tool or by hand.

tuned/plugins/plugin_scheduler.py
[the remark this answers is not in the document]
olysonek: I think even `non_isolated_cores_cgroup` is better - it's wordy, but at least it's descriptive.

AFAIK, the non-isolated cores/tasks are sometimes referred to as "housekeeping". So another option might be `housekeeping_cgroup`.

tuned/plugins/plugin_scheduler.py
```diff
@@ -350,8 +410,103 @@ def _convert_sched_cfg(self, vals):
...
+			self._cgroup_set_affinity_one(cg[0], cg[1], backup = True)
+
+	def _cgroup_restore_affinity(self):
+		log.debug("Restoring cgroups affinities")
+		for cg in self._cgroups_original_affinity.items():
+			self._cgroup_set_affinity_one(cg[0], cg[1])
+
 	def _instance_apply_static(self, instance):
+		# need to get "cgroup_mount_point_init", "cgroup_mount_point", "cgroup_groups_init",
+		# "cgroup", and initialize mount point and cgroups before super class init
```
olysonek, line 493: Did you mean "before calling the _instance_apply_static method of the super class"? I understand "init" to mean `__init__`.

yarda: Yep, typo.

tuned/plugins/plugin_scheduler.py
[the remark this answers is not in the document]
yarda: Also proposing `cgroup_for_isolated_cores`, it's not accurate, but it could be more intuitive - i.e. that it's bind with the isolated_cores option.

tuned/plugins/plugin_scheduler.py
```diff
@@ -350,8 +410,103 @@ def _convert_sched_cfg(self, vals):
 		affinity = self._convert_affinity(affinity)
 		return (rule_prio, scheduler, priority, affinity, regex)
 
+	def _cgroup_create_group(self, cgroup):
+		path = "%s/%s" % (self._cgroup_mount_point, cgroup)
+		try:
+			os.mkdir(path, consts.DEF_CGROUP_MODE)
+		except OSError as e:
+			log.error("Unable to create cgroup '%s': %s" % (path, e))
+		if (not self._cmd.write_to_file("%s/%s" % (path, "cpuset.mems"),
```
olysonek, line 419: Why are `cpuset.mems` and `cpuset.cpus` initialized here? I'm not that familiar with cpusets, but I'd expect the new cpusets to automatically inherit these parameters from the parent cpuset. Which seems like a desireable behaviour.

yarda: IMHO with the cgroup v1 you need to initialize it.

olysonek: It seems you're right. If I don't initialize it, then writing to the `tasks` file fails with "No space left on device". That's just crazy...

tuned/plugins/plugin_scheduler.py
```diff
@@ -350,8 +410,103 @@ def _convert_sched_cfg(self, vals):
...
+	def _cgroup_set_affinity_one(self, cgroup, affinity, backup = False):
+		log.debug("Setting cgroup '%s' affinity to '%s'" % (cgroup, affinity))
+		path = "%s/%s/%s" % (self._cgroup_mount_point, cgroup, "cpuset.cpus")
+		if backup:
+			orig_affinity = self._cmd.read_file(path, err_ret = "ERR", no_error = True)
+			if orig_affinity != "ERR":
+				self._cgroups_original_affinity[cgroup] = orig_affinity
+			else:
+				log.warn("Unable to save affinity of cgroup '%s'" % cgroup)
+		if not self._cmd.write_to_file(path, affinity, no_error = True):
```
olysonek, line 476: Should we change the affinity if we failed to read the original affinity? In general I think we shouldn't change anything that we can't change back. There's an example of this in `_tune_process_affinity` (the "Refusing to set..." error).

yarda: I am not sure, from the performance point of view I would prefer to set it (or attempt to set it), because I would care about the performance, not about the possibility to roll-back. But I agree we should be consistent in the behaviour. Also this is a bit different than just the plain affinity, because in case of removal of the cgroup we move the task to the root cgroup. But I dont' have problem to change it.

tuned/plugins/plugin_scheduler.py
```diff
@@ -350,8 +410,103 @@ def _convert_sched_cfg(self, vals):
...
+	def _remove_dir(self, cgroup):
+		try:
+			os.rmdir(cgroup)
+		except OSError as e:
+			log.error("Unable to remove directory '%s': %s" % (cgroup, e))
+
+	def _cgroup_finalize_groups(self):
+		if self._cgroup is not None and not self._cgroup in self._cgroups:
+			self._remove_dir("%s/%s" % (self._cgroup_mount_point, self._cgroup))
+		for cg in self._cgroups:
```
olysonek, line 453: I think the directories need to be removed in reversed order. If you have `cgroup.parent` and `cgroup.parent/child`, then removing parent will fail because child is not yet removed at that point.

yarda: Reversed.

tuned/plugins/plugin_scheduler.py
olysonek: And if the failure is not really expected, I think `log.error` is more appropriate than `log.warn`.

yarda: > Well, it's not really _expected_ for the read operation to fail, is it? If it does fail, I think it either means that there's a bug in Tuned (in which case it's safer to do nothing), or there's some deeper problem in the system. I think a failure of the read operation is mostly hypothetical, so we should prefer consistency over hypothetical performance gains.
>
It's the read of kernel special file, not regular file, which is handled by the cgroup kernel driver. I am not familiar with it's implementation, so the failure can mean nearly anything.

> > Also this is a bit different than just the plain affinity, because in case of removal of the cgroup we move the task to the root cgroup. But I dont' have problem to change it.
> 
> Well, in the case where the cgroups are removed, there's not really a point in doing the backup in the first place. (Or at least that's my current understanding.) So in that case, it would be better not to do the backup at all.
> 
The cgroup remove operation could fail and the cgroup could also exists before Tuned attempted to create it, in such case it's safer to backup.

> What we should be talking about here is the case where the cgroups are _not_ removed, i.e. they are user provided. In that case, we're not really talking about restoring affinity of the _tasks_, but about restoring affinity of the _cgroups_. And we should treat the cgroups as any other thing that deserves rollback.

As I wrote earlier I am for consistency in behaviour, so I am not against your proposal.

tuned/plugins/plugin_scheduler.py
olysonek: I'd like to reopen this thread...

Is it really necessary that `cpuset.cpus` is initialized *here*, given that it's re-initialized a bit later in `_cgroup_set_affinity` (called from `_instance_apply_static`)?

I think the initialization is causing real problems. I'm testing with the following profile:
```
[scheduler]
cgroup_mount_point=/sys/fs/cgroup/cpuset
cgroup_mount_point_init=0
cgroup_groups_init=1
cgroup.group1=2
cgroup.group1/child=2
```
With this profile, Tuned gives the following error:
```
2020-06-05 21:35:40,683 ERROR    tuned.plugins.plugin_scheduler: Unable to set affinity '2' for cgroup 'group1'
```
I believe that what is happening is the following:
1. `group1/cpuset.cpus` is set to 0-7 (which is all of my CPU cores)
2. `group1/child/cpuset.cpus` is set to 0-7
3. `group1/cpuset.cpus` is set to 2, but this fails, because there's a child cgroup with affinity 0-7, which is not a subset of {2}.

olysonek: Note that if I drop the initialization, then there's an error during rollback. Not sure what that's about yet.

yarda: > Note that if I drop the initialization, then there's an error during rollback. Not sure what that's about yet.

Probably because it reads as empty file before initialized, and it's impossible to roll it back to this state.

yarda: Btw the cgroup v1 implementation has many inconsistencies, it's even possible to get it easily into the "non-allowed" state. You will be probably really unhappy with its behavior :) That's why the more polished (and more complex) v2 implementation was proposed.

yarda: Maybe we could drop the cpuset.cpus init and do not rollback if there is empty value read.

tuned/plugins/plugin_scheduler.py
olysonek: > > Well, it's not really _expected_ for the read operation to fail, is it? If it does fail, I think it either means that there's a bug in Tuned (in which case it's safer to do nothing), or there's some deeper problem in the system. I think a failure of the read operation is mostly hypothetical, so we should prefer consistency over hypothetical performance gains.
> 
> It's the read of kernel special file, not regular file, which is handled by the cgroup kernel driver. I am not familiar with it's implementation, so the failure can mean nearly anything.

Fair enough. However it seems like a very basic operation that I really would not expect to fail.

> > > Also this is a bit different than just the plain affinity, because in case of removal of the cgroup we move the task to the root cgroup. But I dont' have problem to change it.
> > 
> > 
> > Well, in the case where the cgroups are removed, there's not really a point in doing the backup in the first place. (Or at least that's my current understanding.) So in that case, it would be better not to do the backup at all.
> 
> The cgroup remove operation could fail and the cgroup could also exists before Tuned attempted to create it, in such case it's safer to backup.

I'd suggest that if creating the cgroup fails because the cgroup is already there, then we should abort everything and not touch those cgroups at all :). Because that's clearly either a bug or misconfiguration, so it's safer not to do anything.

So no cgroup touching => no backup needed :).

tuned/plugins/plugin_scheduler.py
olysonek: > Probably because it reads as empty file before initialized, and it's impossible to roll it back to this state.

That might be it.

So that's exactly the case where we're creating the cgroups and, as I'm suggesting in the other thread, no backup/rollback is needed.

olysonek: > Maybe we could drop the cpuset.cpus init and do not rollback if there is empty value read.

I'm guessing that would skip the rollback in exactly those cases where it makes sense. It wouldn't skip the backup, but that's not a blocker for me right now.

So I think that could be done.

tuned/plugins/plugin_scheduler.py
yarda: This wouldn't be good for recovery, e.g. recovery from deamon crash etc. In such case just the restart (or double restart) fixes the problem.

olysonek: For those cases, there probably should be a `force` option to make Tuned ignore the directory creation error. Or perhaps a way to tell Tuned to cleanup the leftover directories.

tuned/plugins/plugin_scheduler.py
yarda: PRed the change. Unfortunately, I wasn't able to test it, because I lost the test environment.

tuned/plugins/plugin_scheduler.py
yarda: This has to be engine / tuned-adm wide.

olysonek: To make it easy to use (i.e. without modifying the profile), yes, I think it would have to be a global option. But I think it's the right solution.

olysonek: Note that I'm not saying that you should now go off and implement the force/cleanup option. We can do that later :).

I guess the only thing I'd like to do right now is to decide the general direction where the code is going to go.

tuned/plugins/plugin_scheduler.py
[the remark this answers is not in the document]
olysonek: `cgroup_for_isolated_cores` sounds good.

yarda:
Thanks, merging.

PR #276, opened 2020-06-03
Arch specific tuning for Marvell ThunderX and AMD

yarda:
This adds architecture specific tuning for the Marvell ThunderX and AMD.

It works but I can see few problems with this approach:
1. sysctl are reapplied multiple times - it should be harmless, but it's inefficient.
2. if conflicting settings are used in the arch specific instance and in the base instance, both settings will be applied and the last settings will win - this is not optimal, but it shouldn't cause trouble for unloading, because the units are unloaded in the opposite order as they were loaded.
3. if someone wants to override the arch specific tuning she or he has to override the arch specific instance or has to increase the base instance priority to change the processing order (similarly for the case if the arch specific tuning was defined the first).

I think 1 and 2 are fixable, but fix for 3 would probably require redesign and redefinition of the way how Tuned behaves. However, I think it's not blocker and the behaviour could be described in the documentation.

### DIFF
--- a/profiles/throughput-performance/tuned.conf
+++ b/profiles/throughput-performance/tuned.conf
@@ -7,6 +7,7 @@ summary=Broadly applicable tuning that provides excellent performance across a v
 
 [variables]
 thunderx_cpuinfo_regex=CPU part\s+:\s+(0x0?516)|(0x0?af)|(0x0?a[0-3])|(0x0?b8)\b
+amd_cpuinfo_regex=model name\s+:.*\bAMD\b
 
 [cpu]
 governor=performance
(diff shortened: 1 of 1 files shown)


olysonek:
There's also the problem that if there are conflicting settings in the units, then verification will fail.

All of these problems would go away if these were overlay profiles, but I guess that's not what we want, because then we wouldn't have autodetection.

Let me think about this for a bit. But I don't think we can do better currently.

yarda:
Added more known IDs.

yarda:
Thanks, merging.

PR #279, opened 2020-06-08
[For v2.15] Make tuned-adm indicate which profile is post-loaded

olysonek:
Currently, the post-loaded profile is shown only among regular active profiles. For example:

Current active profile: balanced post

This is not entirely user-friendly, because the user may not know why the post-loaded profile is there, given that they did not specify it in a 'tuned-adm profile' command and given that it's not present in
/etc/tuned/active_profile.

To give the user a hint about this and also to make it easy to see exactly which profile is post-loaded, we make 'tuned-adm active' show the following:

Current active profile: balanced post
Current post-loaded profile: post

This PR additionally fixes a bug with `tuned-adm profile_info`.

### DIFF
--- a/com.redhat.tuned.policy
+++ b/com.redhat.tuned.policy
@@ -27,6 +27,16 @@
     </defaults>
   </action>
 
+  <action id="com.redhat.tuned.post_loaded_profile">
+    <description>Show active post-loaded profile</description>
+    <message>Authentication is required to show active post-loaded profile</message>
+    <defaults>
+      <allow_any>yes</allow_any>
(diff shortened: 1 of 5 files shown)


olysonek:
I don't think we need this in v2.14. I think it can wait for v2.15.

olysonek:
> I don't think we need this in v2.14. I think it can wait for v2.15.

But of course feel free to merge it sooner if you like.

yarda:
Thanks I think it can go to 2.14. It could break some Beaker tests which use hardcoded responses, but it can be easily fixed.

PR #283, opened 2020-06-30
throughput-performance: fix performance regression on AMD

yarda:
It turned out that disablement of the numa_balancing could result in
upto 20% performance drop on some loads.

Related: rhbz#1746957


yarda:
Thanks, merging.

PR #284, opened 2020-07-03
scheduler: fix isolated_cores to work with cgroups

yarda:
It seems cpuset.cpus needs to be initialized before writing to tasks.

Related: rhbz#1784648


yarda:
Thanks, merging.

PR #285, opened 2020-07-03
cpu-partitioning: do not hardcode path to mktemp

yarda:
E.g. Ubuntu uses /bin/mktemp, while e.g. Fedora uses /usr/bin/mktemp.
There is probably no need to hardcode the path, so rely on the system path.


yarda:
Thanks, merging.

PR #291, opened 2020-07-29
realtime-virtual-host profile: exclude ovs-vswitchd pmds

christophefontaine:
Like the cpu-partitioning profile, exclude all 'pmd' threads
to avoid repinning ovs-vswitchd pmds.

Resolves: rhbz#1861767


olysonek:
LGTM. Perhaps this should be done for realtime-virtual-guest too?

christophefontaine:
> LGTM. Perhaps this should be done for realtime-virtual-guest too?
I don't know, as this targets ovs-dpdk pmds: the thread naming scheme for other dpdk based applications may be different, so I don't think it should be applied as-is for realtime-virtual-guest.

yarda:
Thanks for contribution, merging.

PR #293, opened 2020-08-21
Added plugin 'service'

yarda:
The service plugin can handle sysvinit, sysv-rc, openrc, systemd.
Syntax is the following:

[service]
service.sendmail = start,enable,file:${i:PROFILE_DIR}/tuned-sendmail.conf

'start' means to start the service, 'stop' to stop, 'enable' to enable,
'disable' to disable, file:FILE is the overlay configuration file
that will be installed for the service. Multiple directives can be
separated by comma ',' or semicolon ';'. If the directives are
conflicting, the last one will be used. The internal variable
${i:PROFILE_DIR} points to the directory from which the profile is
loaded.

The service plugin supports configuration overlays only for the systemd,
for other init systems this directive is ignored. The configuration
overlay files are copied to the /etc/systemd/system/SERVICE_NAME.service.d/
directories. Upon unloading if the directory is empty, it's removed.

With systemd, the 'start' directive is implemented by 'restart' in order
to allow loading of the service configuration file overlay.

With init systems other than systemd, it works only with the current runlevel.

Resolves: rhbz#1869991

### DIFF
--- a/tuned/plugins/plugin_service.py
+++ b/tuned/plugins/plugin_service.py
@@ -1,0 +1,283 @@
+from . import base
+import collections
+import tuned.consts as consts
+from .decorators import *
+import os
+import re
+import tuned.logs
+from tuned.utils.commands import commands
+
+log = tuned.logs.get()
+cmd = commands()
+
+class Service():
+	def __init__(self, start = None, enable = None, cfg_file = None, runlevel = None):
+		self.enable = enable
+		self.start = start
+		self.cfg_file = cfg_file
+		self.runlevel = runlevel
+
+class InitHandler():
+	def runlevel_get(self):
+		(retcode, out) = cmd.execute(["runlevel"])
+		return out.split()[-1] if retcode == 0 else None
+
+	def daemon_reload(self):
+		cmd.execute(["telinit", "q"])
+
+	def cfg_install(self, name, cfg_file):
+		pass
+
+	def cfg_uninstall(self, name, cfg_file):
+		pass
+
+	def cfg_verify(self, name, cfg_file):
+		return None
+
+# no enable/disable
+class SysVBasicHandler(InitHandler):
+	def start(self, name):
+		cmd.execute(["service", name, "start"])
+
+	def stop(self, name):
+		cmd.execute(["service", name, "stop"])
+
+	def enable(self, name, runlevel):
+		raise NotImplementedError()
+
+	def disable(self, name, runlevel):
+		raise NotImplementedError()
+
+	def is_running(self, name):
+		(retcode, out) = cmd.execute(["service", name, "status"], no_errors = [0])
+		return retcode == 0
+
+	def is_enabled(self, name, runlevel):
+		raise NotImplementedError()
+
+class SysVHandler(SysVBasicHandler):
+	def enable(self, name, runlevel):
+		cmd.execute(["chkconfig", "--level", runlevel, name, "on"])
+
+	def disable(self, name, runlevel):
+		cmd.execute(["chkconfig", "--level", runlevel, name, "off"])
+
+	def is_enabled(self, name, runlevel):
+		(retcode, out) = cmd.execute(["chkconfig", "--list", name])
+		return out.split("%s:" % str(runlevel))[1][:2] == "on" if retcode == 0 else None
+
+class SysVRCHandler(SysVBasicHandler):
+	def enable(self, name, runlevel):
+		cmd.execute(["sysv-rc-conf", "--level", runlevel, name, "on"])
+
+	def disable(self, name, runlevel):
+		cmd.execute(["sysv-rc-conf", "--level", runlevel, name, "off"])
+
+	def is_enabled(self, name, runlevel):
+		(retcode, out) = cmd.execute(["sysv-rc-conf", "--list", name])
+		return out.split("%s:" % str(runlevel))[1][:2] == "on" if retcode == 0 else None
+
+class OpenRCHandler(InitHandler):
+	def runlevel_get(self):
+		(retcode, out) = cmd.execute(["rc-status", "-r"])
+		return out.strip() if retcode == 0 else None
+
+	def start(self, name):
+		cmd.execute(["rc-service", name, "start"])
+
+	def stop(self, name):
+		cmd.execute(["rc-service", name, "stop"])
+
+	def enable(self, name, runlevel):
+		cmd.execute(["rc-update", "add", name, runlevel])
+
+	def disable(self, name, runlevel):
+		cmd.execute(["rc-update", "del", name, runlevel])
+
+	def is_running(self, name):
+		(retcode, out) = cmd.execute(["rc-service", name, "status"], no_errors = [0])
+		return retcode == 0
+
+	def is_enabled(self, name, runlevel):
+		(retcode, out) = cmd.execute(["rc-update", "show", runlevel])
+		return bool(re.search(r"\b" + re.escape(name) + r"\b", out))
+
+class SystemdHandler(InitHandler):
+	# runlevel not used
+	def runlevel_get(self):
+		return ""
+
+	def start(self, name):
+		cmd.execute(["systemctl", "restart", name])
+
+	def stop(self, name):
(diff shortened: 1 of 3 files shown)


yarda:
> First of all, thank you! This is great! I didn't go through the code much, but I did test the `systemd` init system part of this new `service` plugin and it does what's "on the tin". The only suggestion I have at this point is how should the "unapply" of settings be handled. Should a service be restarted in case it was running to go back to the old settings once the overlay file is gone? If so, should the restart be optional or automatic? I'll play with this PR more later on and see if I have any more suggestions. Many thanks!

I think with the systemd the service is restarted even on unapply (in case it should be running). With other init systems it's only started, but it shouldn't be problem, because configuration overlay is supported only with the systemd.

yarda:
> Just tried to run this inside a container. One problem I'm seeing is the runlevel detection. Inside a container, I'm getting runlevel "unknown". I'll try to dig more into this and see if I can make the runlevel detection possible. Failing that, is the runlevel detection really needed?
> 
> Edit: I was able to work around this by providing a custom runlevel script inside the container.

In fact the runlevel is not used with the systemd handler, so I bypassed the check when systemd is in use.

jmencak:
> > First of all, thank you! This is great! I didn't go through the code much, but I did test the `systemd` init system part of this new `service` plugin and it does what's "on the tin". The only suggestion I have at this point is how should the "unapply" of settings be handled. Should a service be restarted in case it was running to go back to the old settings once the overlay file is gone? If so, should the restart be optional or automatic? I'll play with this PR more later on and see if I have any more suggestions. Many thanks!
> 
> I think with the systemd the service is restarted even on unapply (in case it should be running). With other init systems it's only started, but it shouldn't be problem, because configuration overlay is supported only with the systemd.

Actually, I don't think the unapply/rollback is happening at the moment.  So the question is whether it should -- and I believe so.

Consider scenario 1.  The service is stopped/disabled.  Then the Tuned starts/enables it.  What should be the rollback?  Can/Should we stop/disable the service in that case?  

Scenario 2.  The service is running.  Tuned starts/enables the service but with different overlay file.  Should Tuned restart the service (without the overlay) on the unapply/rollback?

There's probably quite a few scenarios.  However, I believe that the Tuned daemon should probably try to unapply/rollback to the previous state as much as possible.  To that end, it probably needs to keep track of what was the state of the service prior to running the `[service]` plugin.  Can this be done?

yarda:
> Actually, I don't think the unapply/rollback is happening at the moment. So the question is whether it should -- and I believe so.
It should rollback in the _instance_unapply_static.

> Consider scenario 1. The service is stopped/disabled. Then the Tuned starts/enables it. What should be the rollback? Can/Should we stop/disable the service in that case?
The service was stopped/disabled, so I think it should stop/disable the service. It should work this way now.

> Scenario 2. The service is running. Tuned starts/enables the service but with different overlay file. Should Tuned restart the service (without the overlay) on the unapply/rollback?
Yes, exactly. It should also do it, but it seems I introduced bug, because I restart the service with the overlay, which is wrong. I will fix it.

tuned/plugins/plugin_service.py
```diff
@@ -0,0 +1,275 @@
...
+		if cfg_file is None:
+			return None
+		path = "%s/%s" % (consts.SERVICE_SYSTEMD_CFG_PATH % name, os.path.basename(cfg_file))
+		if not os.path.exists(cfg_file):
+			log.error("Unable to find service '%s' configuration '%s'" % (name, cfg_file))
+			return False
+		if not os.path.exists(path):
+			log.error("Service '%s' configuration not installed in '%s'" % (name, path))
+			return False
+		md5sum1 = cmd.md5sum(cfg_file)
```
MarSik, line 159: I am a bit unsure about this. MD5?

MarSik:
In general it looks good and we saw it working, what can we do to get this merged? I see some builds failed.

tuned/plugins/plugin_service.py
yarda: Yep, it's very simple check that the file wasn't modified, it's unlikely that somebody will play with the md5 collisions on the service configuration overlay file just for the profile verification to fail :D. But no problem to change to e.g. sha-256 or whatever is supported by hashlib.

yarda:
> In general it looks good and we saw it working, what can we do to get this merged? I see some builds failed.

I would like to see approval from Tuned developer, @olysonek could you please take a look?

olysonek:
Hi, I'll get to it on Monday.

jmencak:
I wonder if it would make sense to add some info messages like the bootloader plugin does for example with:

```
2020-08-28 16:01:32,583 INFO     tuned.plugins.plugin_bootloader: installing additional boot command line parameters to grub2
```

This might help troubleshooting this *new* functionality a bit even though these messages are probably much closer to debugging.

olysonek (commented):
Looks good to me overall. With the `is_enabled` issue resolved, consider this approved.

jmencak:
One thing that will definitely need to be fixes is when the overlay configuration file is changed and then the Tuned daemon reloaded, you'll get something like this:
```
Warning: The unit file, source configuration file or drop-ins of XYZ.service changed on disk. Run 'systemctl daemon-reload' to reload units.
```
New env variable settings will not be reflected when the Tuned daemon is reloaded (e.g. by sending it a HUP signal).  Hitting this issue right now.

jmencak:
> One thing that will definitely need to be fixed is when the overlay configuration file is changed and then the Tuned daemon reloaded, you'll get something like this:
> 
> ```
> Warning: The unit file, source configuration file or drop-ins of XYZ.service changed on disk. Run 'systemctl daemon-reload' to reload units.
> ```
> 
> New env variable settings will not be reflected when the Tuned daemon is reloaded (e.g. by sending it a HUP signal). Hitting this issue right now.

I think this can be easily fixed by adding something like this: 
```
        def daemon_reload(self):
                cmd.execute(["systemctl", "daemon-reload"])
```

to `SystemdHandler` and then calling it from `cfg_install()` and `cfg_uninstall()`.  Can we have this fixed, Jardo?  Thanks!
/cc @MarSik

tuned/plugins/plugin_service.py
```diff
@@ -0,0 +1,275 @@
...
+	def disable(self, name, runlevel):
+		cmd.execute(["systemctl", "disable", name])
+
+	def is_running(self, name):
+		(retcode, out) = cmd.execute(["systemctl", "is-active", name], no_errors = [0])
+		return retcode == 0
+
+	def is_enabled(self, name, runlevel):
+		(retcode, out) = cmd.execute(["systemctl", "is-enabled", name], no_errors = [0])
+		return retcode == 0
```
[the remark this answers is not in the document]
olysonek: Looking at it again, it turns out "static" means the unit cannot be enabled. Not sure about the other enable statuses that result in a zero exit code. Still, I think it's worth looking into if you haven't done so.

With a "static" service, the following can happen. Let's say we have a profile as follows
```
[service]
service.a=start
```
where a.service is a static service. On rollback of this profile, Tuned will unnecessarily/wrongly attempt to enable a.service.
```
2020-09-01 04:12:45,395 DEBUG    tuned.utils.commands: Executing ['systemctl', 'enable', 'a'].
```
This is not exactly serious and it doesn't cause errors, but still worth fixing in my opinion.

yarda:
- added log.info when installing/uninstalling the service configuration overlay file
- is_enabled returns True if service is "enabled", False if service is "disabled", None otherwise
- I haven't used 'systemctl daemon-reload' workaround, but more portable 'telinit q'

jmencak:
>     * added log.info when installing/uninstalling the service configuration overlay file
> 
>     * is_enabled returns True if service is "enabled", False if service is "disabled", None otherwise
> 
>     * I haven't used 'systemctl daemon-reload' workaround, but more portable 'telinit q'

Thanks, I confirm the `telinit q` also works.

olysonek (approved):
Note that I'm unable to verify some of the non-systemd assumptions, but it looks good to me.

yarda:
Thanks, merging.

PR #298, opened 2020-10-01
Synchronize documentation files from RHEL 8

mrksu:
The manual in this repo and the RHEL 8 documentation files have slightly diverged over time.

With this PR, I'm performing a manual conflict resolution, merging changes between the two. We also have a similar merge request in RHEL 8 docs to incorporate the changes from this repo.

Most of the edits here are caused by us changing our preferred sub-heading syntax. But there's also a couple of less trivial edits, such as adding the _Setting the disk scheduler using Tuned_ module or documenting the `cpu-partitioning` profile.

After this PR is merged, we can be quicker to synchronize the two documentation sets and accept changes thanks to a synchronization tool that I've made. The tool is internal to Red Hat now because it interacts with our internal documentation repo but I can share details with other Red Hatters if you're interested.


yarda:
Thanks, LGTM, merging.

PR #306, opened 2020-11-10
scheduler: added default_irq_smp_affinity option

yarda:
The option 'default_irq_smp_affinity' affects what will be written to the
/proc/irq/default_smp_affinity. The option supports the following values:

calc:
  Content of the /proc/irq/default_smp_affinity will be calculated from
  the 'isolated_cores'. Non-isolated cores are calculated as an inversion
  of the 'isolated_cores'. Then the intersection of the non-isolated cores
  and the previous content of the /proc/irq/default_smp_affinity is
  written to the /proc/irq/default_smp_affinity. If the intersection is
  empty set, then just the non-isolated cores are written to the
  /proc/irq/default_smp_affinity. This behavior is the default if
  option 'default_irq_smp_affinity' is not specified.

ignore:
  The /proc/irq/default_smp_affinity is not touched by Tuned.

cpulist like e.g. '1,3-4':
  The cpulist is unpacked and directly written to the
  /proc/irq/default_smp_affinity

Examples:

[scheduler]
isolated_cores = 1, 2
default_irq_smp_affinity = calc

[scheduler]
isolated_cores = 1, 2
default_irq_smp_affinity = ignore

[scheduler]
isolated_cores = 1, 3
default_irq_smp_affinity = 2, 4

Resolves: rhbz#1896348

### DIFF
--- a/doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
+++ b/doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
@@ -84,6 +84,25 @@ It sets Aggressive Link Power Management (ALPM) to the value specified by the [o
 `mounts`::
 Enables or disables barriers for mounts according to the Boolean value of the [option]`disable_barriers` option. 
 
+`scheduler`::
+Allows tuning of scheduling priorities, processes/threads/IRQs affinities, and CPU cores isolation.
++
+The `default_irq_smp_affinity` parameter controls the values *Tuned* writes to `/proc/irq/default_smp_affinity`.
+The following values are supported:
++
+--
+`calc`::
+Content of `/proc/irq/default_smp_affinity` will be calculated from the `isolated_cores` parameter.
+Non-isolated cores are calculated as an inversion of the `isolated_cores`. Then the intersection of the non-isolated cores
+and the previous content of `/proc/irq/default_smp_affinity` is written to `/proc/irq/default_smp_affinity`.
+If the intersection is an empty set, then just the non-isolated cores are written to `/proc/irq/default_smp_affinity`.
+This behavior is the default if the parameter `default_irq_smp_affinity` is omitted.
+`ignore`::
+*Tuned* will not touch `/proc/irq/default_smp_affinity`.
+cpulist such as `1,3-4`::
+The cpulist is unpacked and written directly to `/proc/irq/default_smp_affinity`.
+--
+
 `script`::
 Executes an external script or binary when the profile is loaded or unloaded. You can choose an arbitrary executable.
 +
--- a/tuned/plugins/plugin_scheduler.py
+++ b/tuned/plugins/plugin_scheduler.py
@@ -147,6 +147,7 @@ class SchedulerPlugin(base.Plugin):
 			"cgroup_for_isolated_cores": None,
 			"ps_whitelist": None,
 			"ps_blacklist": None,
+			"default_irq_smp_affinity": "calc",
 		}
 
 	def _sanitize_cgroup_path(self, value):
@@ -707,6 +708,17 @@ class SchedulerPlugin(base.Plugin):
 		if enabling and value is not None:
 			self._ps_blacklist = "|".join(["(%s)" % v for v in re.split(r"(?<!\\);", str(value))])
 
+	@command_custom("default_irq_smp_affinity", per_device = False)
+	def _default_irq_smp_affinity(self, enabling, value, verify, ignore_missing):
+		# currently unsupported
+		if verify:
+			return None
+		if enabling and value is not None:
+			if value in ["calc", "ignore"]:
+				self._default_irq_smp_affinity_value = value
+			else:
+				self._default_irq_smp_affinity_value = self._cmd.cpulist_unpack(value)
+
 	# Raises OSError
 	# Raises SystemError with old (pre-0.4) python-schedutils
 	# instead of OSError
@@ -846,9 +858,13 @@ class SchedulerPlugin(base.Plugin):
 		# default affinity
 		prev_affinity_hex = self._cmd.read_file("/proc/irq/default_smp_affinity")
 		prev_affinity = self._cmd.hex2cpulist(prev_affinity_hex)
-		_affinity = self._get_intersect_affinity(prev_affinity, affinity, affinity)
-		self._set_default_irq_affinity(_affinity)
-		irq_original.default = prev_affinity
+		if self._default_irq_smp_affinity_value == "calc":
+			_affinity = self._get_intersect_affinity(prev_affinity, affinity, affinity)
+		elif self._default_irq_smp_affinity_value != "ignore":
+			_affinity = self._default_irq_smp_affinity_value
+		if self._default_irq_smp_affinity_value != "ignore":
+			self._set_default_irq_affinity(_affinity)
+			irq_original.default = prev_affinity
 		self._storage.set(self._irq_storage_key, irq_original)
 
 	def _restore_all_irq_affinity(self):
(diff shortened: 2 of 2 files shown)


marcel-apf (approved):
Looks good to me, thanks!

jmencak:
I know the comment above and commit message could be considered a sort of a documentation, but really, it is not.  Can we start a practice of adding documentation to every new feature such as this directly in the doc/manual/... ?

yarda:
> I know the comment above and commit message could be considered a sort of a documentation, but really, it is not. Can we start a practice of adding documentation to every new feature such as this directly in the doc/manual/... ?

It's good idea, I will try to transform the commit message into the asciidoc text.

jmencak:
I've tested the default behaviour is unchanged + calc/ignore/cpulist seems to work too.  Rollback also works.  After the docs are added, happy to "lgtm".

yarda:
An attempt to document this change.

After more thinking about it, I think this is not the best way to go. We should probably write such documentation into the code (e.g. by using docstrings) and then automatically generate structured reference guide from it. But it will require more man power to kick this in.

yarda:
We can keep the whole asciidoc documentation (and further extend it), but I think that the chapter regarding the plugins and their options/parameters should be auto generated. This will allow easy maintenance.

doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
```diff
@@ -84,6 +84,22 @@ It sets Aggressive Link Power Management (ALPM) to the value specified by the [o
 `mounts`::
 Enables or disables barriers for mounts according to the Boolean value of the [option]`disable_barriers` option. 
 
+`scheduler`::
+Allows tuning of scheduling priorities, processes/threads/IRQs affinities, and CPU cores isolation.
++
+The `default_irq_smp_affinity` parameter controls what *Tuned* will write to the `/proc/irq/default_smp_affinity`.
```
jmencak, line 90: Suggestion:
s/what *Tuned* will write/the values *Tuned* writes/
s/to the/to/
Even though the file is "unique" I believe 'the' should be dropped as it is is "name".  A document person should review this.

yarda:
Maybe there are still some 'wording' problems in the documentation added by this commit, waiting for feedback.

doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
```diff
@@ -84,6 +84,22 @@ It sets Aggressive Link Power Management (ALPM) to the value specified by the [o
...
+The following values are supported:
++
+--
+** `calc` content of the `/proc/irq/default_smp_affinity` will be calculated from the `isolated_cores` parameter.
```
jmencak, line 94: s/of the/of/  (everywhere where you refer to *the* file :)

doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
```diff
@@ -84,6 +84,22 @@ It sets Aggressive Link Power Management (ALPM) to the value specified by the [o
...
+Non-isolated cores are calculated as an inversion of the `isolated_cores`. Then the intersection of the non-isolated cores
```
jmencak, line 95: s/of the `isolated_cores`/of `isolated_cores`/  I don't see isolated_cores mentioned anywhere, therefore I'd drop it here.

doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
```diff
@@ -84,6 +84,22 @@ It sets Aggressive Link Power Management (ALPM) to the value specified by the [o
...
+and the previous content of the `/proc/irq/default_smp_affinity` is written to the `/proc/irq/default_smp_affinity`.
+If the intersection is empty set, then just the non-isolated cores are written to the `/proc/irq/default_smp_affinity`.
```
jmencak, line 97: s/is empty/is an empty/

doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
```diff
@@ -84,6 +84,22 @@ It sets Aggressive Link Power Management (ALPM) to the value specified by the [o
...
+This behavior is the default if the parameter `default_irq_smp_affinity` is omitted.
+** `ignore` *Tuned* will not touch `/proc/irq/default_smp_affinity`.
+** cpulist like e.g. `1,3-4` the cpulist is unpacked and directly written to the `/proc/irq/default_smp_affinity`.
```
jmencak, line 100: s/cpulist like e.g./for a cpulist such as/

jmencak (commented):
Thanks for the docs!  Looks good, minor nits a documentation writer or a native English speaker should confirm.

jmencak:
@courtneypacheco , could you please go over the docs?  Thank you!

doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
yarda: It's mentioned on the previous line. I think the parameter `isolated_cores` should be described in its own paragraph preceding this text, but it's out of the context of this commit, so I didn't add it.

doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
yarda: It's list describing 'calc', 'ignore', and cpulist parameters and it's introduced by the sentence: 'The following values are supported:', that's why I didn't want to write 'for a cpulist' but just 'cpulist'.

yarda: I reformatted the text to use asciidoc description, but I am not sure whether it looks better, nor whether the syntax I used is correct :)

yarda:
I addressed most of the issues except those where I put comment.

doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
```diff
@@ -84,6 +84,25 @@ It sets Aggressive Link Power Management (ALPM) to the value specified by the [o
...
+`calc`::
+Content of `/proc/irq/default_smp_affinity` will be calculated from the `isolated_cores` parameter.
+Non-isolated cores are calculated as an inversion of the `isolated_cores`. Then the intersection of the non-isolated cores
+and the previous content of `/proc/irq/default_smp_affinity` is written to `/proc/irq/default_smp_affinity`.
+If the intersection is an empty set, then just the non-isolated cores are written to `/proc/irq/default_smp_affinity`.
+This behavior is the default if the parameter `default_irq_smp_affinity` is omitted.
+`ignore`::
+*Tuned* will not touch `/proc/irq/default_smp_affinity`.
+cpulist such as `1,3-4`::
+The cpulist is unpacked and directly written to `/proc/irq/default_smp_affinity`.
```
jmencak, line 103: s/directly written/written directly/  ?

jmencak (approved):
Just a minor nit.  Looks good to me otherwise, thank you.  I've asked Courtney to review this as a native speaker of English.  You may want to give her a day or two before this merges.  Thank you. /cc @courtneypacheco

doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
yarda: I don't know :) but I changed it.

jmencak:
The rendered asciidoc looks OK to me. Still /lgtm

jmencak:
It is unlikely we are getting a native-speaker review of the docs within a reasonable timeframe.  I believe this is looks good and is good to go, @yarda

yarda:
OK, thanks for info. I also think there is no point in blocking the merge due to it. We can update the text anytime later.

PR #307, opened 2020-11-19
Don't modify boom managed BLS snippets

bmr-cymru:
Currently the 92-tuned.install kernel-install plugin modifies all
BLS snippets that match the KERNEL_INSTALL_MACHINE_ID passed to
the script.

This makes sense for the system managed entries but it conflicts
with entries that are managed by the boom boot manager: boom will
propagate these options if they are present when the entry is set
up, but the files should not be modified outside boom's control
since this will cause the boom boot identifier to change.

Boom boot entries include the boot_id in the file name pattern:

  %{machine_id}-%{boot_id}-%{version}.conf

Compared to the system managed entries which use:

  %{machine_id}-%{version}.conf

Modify 92-tuned.install to skip this change for files that match
the boom naming convention.

Rather than rely on the presence of '-' (which may appear in some
kernel builds, especially custom or debug kernels), use a regular
expression to match the 7-character or greater SHA1 hash that
appears in boom BLS snippet file names and the remaining structure
of the file name.

A paired change is made to the tuned.spec %postun script to apply
the same scope to the logic that reverts these changes on removal.

This gives the correct behaviour for me of continuing to add the
$tuned_... references to the system managed entries where needed,
but without causing unexpected changes to the boom boot entries.
The spec file also correctly reverts the change for the same set
of entries when %postun is executed.

Signed-off-by: Bryn M. Reeves <bmr@redhat.com>

### DIFF
--- a/92-tuned.install
+++ b/92-tuned.install
@@ -25,10 +25,16 @@ ARCH=`uname -m`
 
 pushd "$LOADER_ENTRIES" &> /dev/null
 for f in `basename "$MACHINE_ID"`-*.conf; do
-  if [ -f "$f" -a "${f: -12}" != "-rescue.conf" ]; then
-    grep -q '^\s*options\s\+.*\$tuned_params' "$f" || sed -i '/^\s*options\s\+/ s/\(.*\)/\1 \$tuned_params/' "$f"
-    grep -q '^\s*initrd\s\+.*\$tuned_initrd' "$f" || sed -i '/^\s*initrd\s\+/ s/\(.*\)/\1 \$tuned_initrd/' "$f"
+  # Skip non-files and rescue entries
+  if [ ! -f "$f" -o "${f: -12}" == "-rescue.conf" ]; then
+    continue
   fi
+  # Skip boom managed entries
+  if [[ "$f" =~ \w*-[0-9a-f]{7,}-.*-.*.conf ]]; then
+    continue
+  fi
+  grep -q '^\s*options\s\+.*\$tuned_params' "$f" || sed -i '/^\s*options\s\+/ s/\(.*\)/\1 \$tuned_params/' "$f"
+  grep -q '^\s*initrd\s\+.*\$tuned_initrd' "$f" || sed -i '/^\s*initrd\s\+/ s/\(.*\)/\1 \$tuned_initrd/' "$f"
 done
 popd &> /dev/null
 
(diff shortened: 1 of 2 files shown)


yarda:
LGTM, but it would be great if boom could clearly mark the BLS entries it manages, e.g. by some suffix (e.g. 'boom') or by using some special tag inside the BLS entry.

bmr-cymru:
Thanks! An additional check would be to look for the string `OsProfile` in a comment inside the BLS snippet, e.g.:

```
# cat /boot/loader/entries/cfef654da87742bf851cf388da0cb87d-8f1010b-4.18.0-193.19.1.el8_2.x86_64.conf
#OsIdentifier: 1ba00433fd7ba24532ec42c3cdf9f457821e6e02
title RHEL Snapshot before upgrade 8.3 -10/11/2020
machine-id cfef654da87742bf851cf388da0cb87d
version 4.18.0-193.19.1.el8_2.x86_64
linux /vmlinuz-4.18.0-193.19.1.el8_2.x86_64
initrd /initramfs-4.18.0-193.19.1.el8_2.x86_64.img $tuned_initrd
options root=/dev/rootvg/root_snap_before_upgrade ro rd.lvm.lv=rootvg/root_snap_before_upgrade $tuned_params
```

We use this as a heuristic to set the OsProfile identifier on distros where the uname version strings cannot be easily mapped to distribution versions. We only rely on it on non-Red Hat distros, but it's guaranteed to be written into every snippet that the tool creates. I didn't implement this check as the naming pattern is sufficient to distinguish all known BLS implementations today but I agree it would be good to tighten it in future.

Thinking forward I would really like to see an update to the BLS specification to make an optional "uniquifying" field explicitly allowed and to add guidance for its formatting. Right now, if you follow the "should" recommendations in the spec to the letter it is only possible to have a single BLS boot entry for each (machine_id, version) tuple. I'd like to propose an optional 3rd element to the naming perhaps with a "name_id" structure (although that could also be optional). In that model our file names would look like:

```
cfef654da87742bf851cf388da0cb87d-boom_8f1010b-4.18.0-193.19.1.el8_2.x86_64.conf
```

I've held off doing this until there's buy-in from others using the standard (and because it makes already long file names even longer) but it's a fairly straightforward change for us to make and to support entries using both the existing and new conventions.

yarda:
OK, thanks for info, please let me know or provide PR in case it's changed.

PR #309, opened 2020-11-24
scheduler: add perf_process_fork option to enable processing of fork

yarda:
Scheduler plugin uses perf event loop to catch newly created processes.
By default it listens to perf.RECORD_COMM and perf.RECORD_EXIT events.

By setting `perf_process_fork` parameter to `true` it will also listen
to perf.RECORD_FORK events, i.e. it will explicitly process
child processes created by the fork system call. Usually, child processes
inherit affinity from their parents, thus the `scheduler` plugin doesn't
need to explicitly process them, that's why the `perf_process_fork`
parameter is by default set to `false` and child processes aren't
explicitly processed. This default setting may also save some CPU time.

Resolves: rhbz#1894610

### DIFF
--- a/doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
+++ b/doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
@@ -87,6 +87,13 @@ Enables or disables barriers for mounts according to the Boolean value of the [o
 `scheduler`::
 Allows tuning of scheduling priorities, processes/threads/IRQs affinities, and CPU cores isolation.
 +
+The scheduler plugin uses perf event loop to catch newly created processes. By default it listens to perf.RECORD_COMM and
+perf.RECORD_EXIT events. By setting `perf_process_fork` parameter to `true`, perf.RECORD_FORK events will be also listened to.
+In other words, child processes created by the fork() system call will be processed. Since child processes inherit CPU affinity
+from their parents, the `scheduler` plugin usually does not need to explicitly process these events. As processing perf events
+can pose a significant CPU overhead, the `perf_process_fork` parameter is set to `false` by default and child processes
+are not processed by the scheduler plugin.
++
 The `default_irq_smp_affinity` parameter controls the values *Tuned* writes to `/proc/irq/default_smp_affinity`.
 The following values are supported:
 +
--- a/tuned/plugins/plugin_scheduler.py
+++ b/tuned/plugins/plugin_scheduler.py
@@ -148,6 +148,7 @@ class SchedulerPlugin(base.Plugin):
 			"ps_whitelist": None,
 			"ps_blacklist": None,
 			"default_irq_smp_affinity": "calc",
+			"perf_process_fork": "false",
 		}
 
 	def _sanitize_cgroup_path(self, value):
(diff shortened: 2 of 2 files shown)


jmencak:
Thank you for the PR, Jardo.  I think that before merging this it would be really useful to assess the real impact of listening for `perf.RECORD_FORK` (and other) events as this is a user-facing change and we may not want to add yet another configuration option that may not be needed.  Has anyone measured this?  Thoughts?

yarda:
> Thank you for the PR, Jardo. I think that before merging this it would be really useful to assess the real impact of listening for `perf.RECORD_FORK` (and other) events as this is a user-facing change and we may not want to add yet another configuration option that may not be needed. Has anyone measured this? Thoughts?

Currently I don't have real benchmark data regarding impact of the fork itself (because it's not easy to get some). But I received report from Joe Mario running libmicro benchmark from which it was clear that the original perf code (without the fork handler) consumes cca. 30-40 % of the CPU time under the benchmark heavy load. Then I spent cca. week trying to optimize the perf loop and realized that it seems there is no much room for optimization. So if it is problem for some high performance oriented customers we recommend them to disable the perf loop and use just the systemd affinity or cgroups which the plugin also supports.

But I ran simple test. I kept my workstation running for 15 minutes doing my normal daily developer tasks and process the perf events through the instrumented Tuned code. Over the 15 minutes Tuned processed 635 COMM events and 388 FORK events. By doing simple math it seems the overhead of the perf loop could increase 1.6 x in such cases when the FORK events will be processed which is really not good (e.g. very rough math: 40 % x 1.6 = 64 % overhead). And we are still talking about normal mostly idle workstation, with the high loading benchmarks the FORK/COMM ratio could be much worse.

By taking into account that the most of the child processes inherits affinity setting from their parents we usually don't need to explicitly process the FORK event. That why I made it configurable.

doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
```diff
@@ -87,6 +87,12 @@ Enables or disables barriers for mounts according to the Boolean value of the [o
 `scheduler`::
 Allows tuning of scheduling priorities, processes/threads/IRQs affinities, and CPU cores isolation.
 +
+It uses perf event loop to catch newly created processes. By default it listens to perf.RECORD_COMM and perf.RECORD_EXIT events.
+By setting `perf_process_fork` parameter to `true` it will also listen to perf.RECORD_FORK events, i.e. it will explicitly process
+child processes created by the fork system call. Usually, child processes inherit affinity from their parents, thus the
+`scheduler` plugin doesn't need to explicitly process them, that's why the `perf_process_fork` parameter is by default set
+to `false` and child processes aren't explicitly processed. This default setting may also save some CPU time.
```
jmencak, line 94: By setting `perf_process_fork` parameter to `true`, perf.RECORD_FORK events will also be listened to.  In other words, child processes created by the fork() system call will be processed. Since child processes inherit CPU affinity from their parents, the `scheduler` plugin usually does not need to explicitly process these events.  As processing perf events can pose a significant CPU overhead, the `perf_process_fork` parameter is set to `false` by default and child processes are not processed by the scheduler plugin.

jmencak (commented):
Thanks for the explanation and the PR, Jardo.  Just a few suggestions wrt. documentation, pick and choose as you feel appropriate.

yarda:
@jmencak thanks, hopefully I addressed all documentation issues.

jmencak:
/lgtm
Only one nit in the docs, contractions such as `aren't` and similar are usually considered informal and probably should not appear in the docs.

yarda:
Fixed, I was copy&pasting too fast :)

jmencak:
Thank you.
/lgtm

yarda:
Thanks, merging.

PR #312, opened 2020-12-03
Reference the correct manpage

7h3w1zz:
Replaces reference to the non-existent tuned-conf with the correct
manpage (tuned.conf)

Signed-off-by: Jacob Collins <jaco3collins@gmail.com>


yarda:
Thanks, LGTM, merged.

PR #314, opened 2020-12-09
Support for adjusting netdev queues

niteshnarayanlal:
Following patch-set enables tuned to adjust the queue count for
each network device via tuned configuration file. This functionality is
implemented by integrating the ethtool channels control feature with tuned.

This support is essentially required for the real-time setup that has
several isolated but very few housekeeping CPUs. In such environments, it
has been observed that the housekeeping CPUs often run out of vectors when
the IRQs are moved from isolated to housekeeping CPUs. This primarily
happens due to a large number of queues (hence IRQs) creation for each
network device based on all available online CPUs.

With this support added tuned can be configured to adjust the queue count
based on the housekeeping CPUs.

Nitesh Narayan Lal (2):
  plugin_net: Implement setting 'channels' parameters
  realtime: added support for netdev_queue_count

 profiles/realtime/realtime-variables.conf     |  8 +++
 profiles/realtime/tuned.conf                  |  3 ++
 tuned/plugins/plugin_net.py                   | 50 +++++++++++++++++--
 .../functions/function_get_queue_count.py     | 23 +++++++++
 4 files changed, 80 insertions(+), 4 deletions(-)
 create mode 100644 tuned/profiles/functions/function_get_queue_count.py

### DIFF
--- a/tuned/plugins/plugin_net.py
+++ b/tuned/plugins/plugin_net.py
@@ -123,6 +123,13 @@ class NetTuningPlugin(base.Plugin):
 			"tx": None }
 
 	@classmethod
+	def _get_config_options_channels(cls):
+		return { "rx": None,
+			"tx": None,
+			"other": None,
+			"combined": None }
+
+	@classmethod
 	def _get_config_options(cls):
 		return {
 			"dynamic": True,
@@ -132,6 +139,7 @@ class NetTuningPlugin(base.Plugin):
 			"coalesce": None,
 			"pause": None,
 			"ring": None,
+			"channels": None,
 		}
 
 	def _init_stats_and_idle(self, instance, device):
@@ -282,7 +290,8 @@ class NetTuningPlugin(base.Plugin):
 		params = set(d.keys())
 		supported_getter = { "coalesce": self._get_config_options_coalesce, \
 				"pause": self._get_config_options_pause, \
-				"ring": self._get_config_options_ring }
+				"ring": self._get_config_options_ring, \
+				"channels": self._get_config_options_channels }
 		supported = set(supported_getter[context]().keys())
 		if not params.issubset(supported):
 			log.error("unknown %s parameter(s): %s" % (context, str(params - supported)))
@@ -313,6 +322,29 @@ class NetTuningPlugin(base.Plugin):
 		l = [x for x in [re.split(r":\s*", x) for x in l] if len(x) == 2]
 		return dict(l)
 
+	# parse output of ethtool -l
+	def _parse_channels_parameters(self, s):
+		a = re.split(r"^Current hardware settings:$", s, flags=re.MULTILINE)
+		s = a[1]
+		s = self._cmd.multiple_re_replace(\
+				{"RX": "rx",
+				"TX": "tx",
+				"Other": "other",
+				"Combined": "combined"}, s)
+		l = s.split("\n")
+		l = [x for x in l if x != '']
+		l = [x for x in [re.split(r":\s*", x) for x in l] if len(x) == 2]
+		return dict(l)
+
+	def _replace_channels_parameters(self, context, params_list, dev_params):
+		mod_params_list = []
+		if "combined" in params_list:
+			mod_params_list.extend(["rx", params_list[1], "tx", params_list[1]])
+		else:
+			cnt = str(max(int(params_list[1]), int(params_list[3])))
+			mod_params_list.extend(["combined", cnt])
+		return dict(list(zip(mod_params_list[::2], mod_params_list[1::2])))
+
 	def _check_device_support(self, context, parameters, device, dev_params):
 		"""Filter unsupported parameters and log warnings about it
 
@@ -337,7 +369,8 @@ class NetTuningPlugin(base.Plugin):
 			parameters.pop(param, None)
 
 	def _get_device_parameters(self, context, device):
-		context2opt = { "coalesce": "-c", "features": "-k", "pause": "-a", "ring": "-g" }
+		context2opt = { "coalesce": "-c", "features": "-k", "pause": "-a", "ring": "-g", \
+                        "channels": "-l"}
 		opt = context2opt[context]
 		ret, value = self._cmd.execute(["ethtool", opt, device])
 		if ret != 0 or len(value) == 0:
@@ -345,7 +378,8 @@ class NetTuningPlugin(base.Plugin):
 		context2parser = { "coalesce": self._parse_device_parameters, \
 				"features": self._parse_device_parameters, \
 				"pause": self._parse_pause_parameters, \
-				"ring": self._parse_ring_parameters }
+				"ring": self._parse_ring_parameters, \
+				"channels": self._parse_channels_parameters }
 		parser = context2parser[context]
 		d = parser(value)
 		if context == "coalesce" and not self._check_parameters(context, d):
@@ -362,10 +396,14 @@ class NetTuningPlugin(base.Plugin):
 		# check if device supports parameters and filter out unsupported ones
 		if dev_params:
 			self._check_device_support(context, d, device, dev_params)
+			# replace the channel parameters based on the device support
+			if context == "channels" and int(dev_params[next(iter(d))]) == 0:
+				d = self._replace_channels_parameters(context, self._cmd.dict2list(d), dev_params)
 
 		if not sim and len(d) != 0:
 			log.debug("setting %s: %s" % (context, str(d)))
-			context2opt = { "coalesce": "-C", "features": "-K", "pause": "-A", "ring": "-G" }
+			context2opt = { "coalesce": "-C", "features": "-K", "pause": "-A", "ring": "-G", \
+                                "channels": "-L"}
 			opt = context2opt[context]
 			# ignore ethtool return code 80, it means parameter is already set
 			self._cmd.execute(["ethtool", opt, device] + self._cmd.dict2list(d), no_errors = [80])
@@ -422,3 +460,7 @@ class NetTuningPlugin(base.Plugin):
 	@command_custom("ring", per_device = True)
 	def _ring(self, start, value, device, verify, ignore_missing):
 		return self._custom_parameters("ring", start, value, device, verify)
+
+	@command_custom("channels", per_device = True)
+	def _channels(self, start, value, device, verify, ignore_missing):
+		return self._custom_parameters("channels", start, value, device, verify)
--- a/tuned/profiles/functions/function_get_queue_count.py
+++ b/tuned/profiles/functions/function_get_queue_count.py
@@ -1,0 +1,23 @@
+import tuned.logs
+from . import base
+from tuned.utils.commands import commands
+
+log = tuned.logs.get()
+
+class get_queue_count(base.Function):
+	"""
+	Checks whether the user has specified a queue count for net devices. If
+        not, return the number of housekeeping CPUs.
+	"""
+	def __init__(self):
+		# 1 argument
+		super(get_queue_count, self).__init__("get_queue_count", 1, 1)
+
+	def execute(self, args):
+		if not super(get_queue_count, self).execute(args):
+			return None
+		if args[0].isdigit():
+			return args[0]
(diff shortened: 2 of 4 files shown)


niteshnarayanlal:
Please let me know if there are comments/suggestions, I will be happy to address them.

tuned/profiles/functions/function_get_queue_count.py
```diff
@@ -0,0 +1,23 @@
...
+	def __init__(self):
+		# 1 argument
+		super(get_queue_count, self).__init__("get_queue_count", 1, 1)
+
+	def execute(self, args):
+		if not super(get_queue_count, self).execute(args):
+			return None
+		if args[0].isdigit():
+			return args[0]
+		(ret, out) = self._cmd.execute(["nproc"])
```
marcel-apf, line 21: I would use the  /sys/devices/system/cpu/online file, nproc may be bound by the current process limits. nproc --all can be used, but better to use the sysfs, I think.
tuned already has some cpulist_online class, maybe will help.

tuned/plugins/plugin_net.py
```diff
@@ -337,15 +369,17 @@ def _check_device_support(self, context, parameters, device, dev_params):
...
 		opt = context2opt[context]
 		ret, value = self._cmd.execute(["ethtool", opt, device])
 		if ret != 0 or len(value) == 0:
 			return None
 		context2parser = { "coalesce": self._parse_device_parameters, \
 				"features": self._parse_device_parameters, \
 				"pause": self._parse_pause_parameters, \
-				"ring": self._parse_ring_parameters }
+				"ring": self._parse_ring_parameters, \
+				"channels": self._parse_channels_parameters }
```
marcel-apf, line 382: A few questions so I can better understand the implications:
What happens if a device needs less queues then specified in the tuned profile?
What happens if the number of queues in the tuned profile is less than minimum queues needed by the device to work properly?
What happens if the number of IRQ vectors per CPU gets exhausted because of too many SR-IOV functions or other devices?
Is this the right place to check/ensure the above?

marcel-apf: Maybe a per device settings would also make sense? Or the setting is per device and I missed that?

tuned/profiles/functions/function_get_queue_count.py
niteshnarayanlal: "/sys/devices/system/cpu/online" includes all the online CPUs (isolated+HK), what we want is to get only the number of HK CPUs.

tuned/plugins/plugin_net.py
niteshnarayanlal: > A few questions so I can better understand the implications:
> What happens if a device needs less queues then specified in the tuned profile?

Tuned will still go ahead and allocate the queues as per the specification in the tuned profile.

The general trend from the device perspective is to have 1 queue per available CPU and that's what we are following here. Tuned should not automatically reduce anything to a minimum as that will impact the performance.

However, if a user wants more queues then tuned should not prevent that either.

> What happens if the number of queues in the tuned profile is less than minimum queues needed by the device to work properly?

This kind of configuration should automatically fail.
The same is the case if you specify a number that is greater than the max allowed queues for the device, tuned/ethtool will simply return an error without any change.

> What happens if the number of IRQ vectors per CPU gets exhausted because of too many SR-IOV functions or other devices?

Do you mean something like a block device? if so then yes that is a control we would like to integrate into tuned as well. However, that can not be achieved from ethtool, so first we have to find a tool that can do that, and then we will have to integrate its support in tuned.
As the first step, this patch-set is only taking care of the network devices.

> Is this the right place to check/ensure the above?

So, one change that I am planning to do in the next version or as another patch set is to terminate tuned execution if tuned's scheduler plugin fails due to ENOSPC (suggested by Marcelo). That way the user will atleast know that some device (eg block) queues need to be adjusted manually since tuned can only take care of the network device queues at the moment.

I also hope that in the future we can have something more concrete from the kernel side to prevent these kinds of situations.

Please let me know if you have any questions/comments.

niteshnarayanlal: > Maybe a per device settings would also make sense? Or the setting is per device and I missed that?

Per device setting that can be controlled by the user? 
The idea is to let tuned configure the device (network for now) queues generically based on HK CPUs for RT setup similar to what kernel does base on the online CPUs at the time of device initialization. 

If the queues for each device need to be adjusted individually then the user can still do that through ethtool directly.

MarSik: The issue here is that not some workloads use kernel networking (as opposed to DPDK), so we want to make sure that it is still possible to configure the queues based on workload knowledge to use enough cpus for interrupt handling.

MarSik: And the per device idea is something similar. There are cases where some NICs are used for DPDK (no queues or interrupts needed really) and some others are used for less important work (management, reporting, aggregated data..). It might be ok to reduce the mgmt NICs queues to housekeeping cores too as the traffic is usually light. But it is an use case that needs to be taken into account.

niteshnarayanlal: > The issue here is that not some workloads use kernel networking (as opposed to DPDK), so we want to make sure that it is still possible to configure the queues based on workload knowledge to use enough cpus for interrupt handling.

May I know who is responsible for controlling/adjusting the queues for these workloads right now based on the workload requirement? I would assume this is still done via userspace using something like ethtool?

MarSik: Yep, ethtool in userspace.

What we are after here is some kind of tuned variable that would let the user override the defaults if needed. I know it is tricky to do this per NIC though..

niteshnarayanlal: > Yep, ethtool in userspace.
> 
> What we are after here is some kind of tuned variable that would let the user override the defaults if needed. I know it is tricky to do this per NIC though..

It should be doable but will require certain changes in plugin_net.py.
Right now, all ethtool commands are integrated in a way that the changes are applied to all the networking devices.

A very crude way that I can think of for doing this is by introducing another variable such as netdev_name.
With this we will have a few options to configure for eg in realtime-virtual-host.conf:
netdev_name=all or netdev_name=eth12 or netdev_name=eth12,eth13,...

realtime's tuned.conf will change to:
[net]
channels=${netdev_name} combined ${f:get_queue_count:${netdev_queue_count}}

The usability of netdev_queue_count will still remain as it is proposed in this patch-set.
What do you think?

MarSik: That might work indeed

niteshnarayanlal: Thanks for confirming.
@yarda any initial thoughts or objections on making this change?

I can start working on this and send another version of the patch.
The only detail here is that I may have to either set "netdev_name=all" as the default setting in all the configurations that inherit the realtime profile or else I can just add a comment suggesting to set this in the realtime-virtual-host.conf, and other files similar to what we have for isolated_cores.

yarda:
Sorry for delay. I added few minor comments. Regarding the new option syntax it's OK for me. Please note that you can setup multiple instances of the net plugin and each instance can even handle multiple devices and you can use regexes in devices_udev_regex specification, e.g.:

```
[net_eth1all]
type=net
devices_udev_regex=eth1.*
channels=combined 2
```

niteshnarayanlal:
> Sorry for delay. I added few minor comments. Regarding the new option syntax it's OK for me. 

Sure no worries. Somehow, I am not able to see the comments, have you made them on github or on the mailing list?

> Please note that you can setup multiple instances of the net plugin and each instance can even handle multiple devices and > you can use regexes in `devices_udev_regex` specification, e.g.:
> 
> ```
> [net_eth1all]
> type=net
> devices_udev_regex=eth1.*
> channels=combined 2
> ```
> 

Ah, I have not used this option myself but I wonder if Martin and Marcel have used this and if this could resolve their concern.
@MarSik ?
 
> I am also going to do some testing of the patch.

That would be great thank you.

yarda:
> Sure no worries. Somehow, I am not able to see the comments, have you made them on github or on the mailing list?

Ah, sorry, my bad, the review was stuck in the 'pending' state and I havent noticed. You should see inlince comments now (just only two comments at the moment).

tuned/profiles/functions/function_get_queue_count.py
```diff
@@ -0,0 +1,23 @@
+import tuned.logs
+from . import base
+from tuned.utils.commands import commands
+
+log = tuned.logs.get()
+
+class get_queue_count(base.Function):
```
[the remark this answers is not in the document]
niteshnarayanlal: Sure, check_net_queue_count() does make more sense.

tuned/plugins/plugin_net.py
```diff
@@ -362,10 +396,14 @@ def _set_device_parameters(self, context, value, device, sim,
...
 			self._check_device_support(context, d, device, dev_params)
+			# replace the channel parameters based on the device support
+			if context == "channels" and int(dev_params[next(iter(d))]) == 0:
+				d = self._replace_channels_parameters(context, self._cmd.dict2list(d), dev_params)
 
 		if not sim and len(d) != 0:
 			log.debug("setting %s: %s" % (context, str(d)))
-			context2opt = { "coalesce": "-C", "features": "-K", "pause": "-A", "ring": "-G" }
+			context2opt = { "coalesce": "-C", "features": "-K", "pause": "-A", "ring": "-G", \
+                                "channels": "-L"}
```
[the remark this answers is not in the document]
niteshnarayanlal: > > Sure no worries. Somehow, I am not able to see the comments, have you made them on github or on the mailing list?
> 
> Ah, sorry, my bad, the review was stuck in the 'pending' state and I havent noticed. You should see the inline comments now (just only two comments at the moment).

No worries.

https://github.com/redhat-performance/tuned/pull/314#discussion_r562243955
I think this is exactly what is required for OCP. So, once Martin also confirms I will create another PR by just incorporating your comments.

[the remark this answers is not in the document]
niteshnarayanlal: > Inconsitent whitespaces, tabs/spaces
@yarda probably a newbie question:
For tuned, I think it is recommended to use tabs over space, isn't it (similar to the kernel)?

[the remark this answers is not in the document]
yarda: Yes, the project uses tabs. Your PR seems good regarding this, just this one line uses spaces.

niteshnarayanlal:
So, I have just tested the per-device configuration based on Jaroslav's suggestion and there is no need for any additional control to do that. Following is an example that I tried at my end:

/usr/lib/tuned/realtime/tuned.conf

..
[net_eth123]
type=net
devices=eth11,eth12
channels=combined 8

[net_eth1]
type=net
devices=eno1
channels=combined 16
...

After applying the profile->

[root@virtlab719 ~]# ethtool -l eno1
Channel parameters for eno1:
Pre-set maximums:
RX:		0
TX:		0
Other:		1
Combined:	80
Current hardware settings:
RX:		0
TX:		0
Other:		1
Combined:	16

[root@virtlab719 ~]# ethtool -l eth11
Channel parameters for eth11:
Pre-set maximums:
RX:		0
TX:		0
Other:		1
Combined:	80
Current hardware settings:
RX:		0
TX:		0
Other:		1
Combined:	8

[root@virtlab719 ~]# ethtool -l eth12
Channel parameters for eth12:
Pre-set maximums:
RX:		0
TX:		0
Other:		1
Combined:	80
Current hardware settings:
RX:		0
TX:		0
Other:		1
Combined:	8

yarda:
> So, I have just tested the per-device configuration based on Jaroslav's suggestion and there is no need for any additional control to do that. Following is an example that I tried at my end:
> 
I will just add that the instance name in the braces, e.g. `[net_eth123]` can be any unique string. It's just to name the instance and you can see this instance name in the logs. Also for the device specification you can use negation, e.g. `devices=!eth0` for all devices except eth0.

niteshnarayanlal:
I have created another PR (https://github.com/redhat-performance/tuned/pull/320) after incorporating the comments, closing this one.

PR #315, opened 2020-12-15
Adding a docstring to the sysctl plugin for the purpose of autogenerating an asciidoc

courtneypacheco:
Added a docstring to the sysctl plugin that describes what the plugin does, what its parameters are, and how it should be used.

### DIFF
--- a/tuned/plugins/plugin_sysctl.py
+++ b/tuned/plugins/plugin_sysctl.py
@@ -16,7 +16,12 @@ SYSCTL_CONFIG_DIRS = [ "/run/sysctl.d",
 
 class SysctlPlugin(base.Plugin):
 	"""
-	Plugin for applying custom sysctl options.
+	This plugin is used for applying custom `sysctl` settings and should only be used to change 
+        system settings that are not covered by other plugins available in *Tuned*. If the settings
+        are covered by other *Tuned* plugins, please use those plugins instead.
+        
+        The syntax for this plugin is `name=value`, where _name_ is the same as the name provided by
+        the `sysctl` utility.
 	"""
 
 	def __init__(self, *args, **kwargs):


yarda:
@courtneypacheco thanks for the PR.

Please be consistent with tab and spaces, so please update this PR (you can use e.g. vim or emacs or mc / mcedit to see the spacing). Update the PR by e.g. `git commit --amend` in the branch for this PR and then force push by e.g. `git push -f`.

Also I think it would be good to have some simple example here, so please add it. You can take inspiration from existing Tuned profiles. Please tag the examples consistently (e.g. by using chapter named Example in the asciidoc). We will later create tools for parsing it.

Please note that this is proof of concept work without tight specification for the formatting/tooling etc. I will start merging these PRs in January so if you come up with some different formatting during your work on different plugins, feel free to update your older PRs to be consistent.

tuned/plugins/plugin_sysctl.py
```diff
@@ -16,7 +16,12 @@
 
 class SysctlPlugin(base.Plugin):
 	"""
-	Plugin for applying custom sysctl options.
+	This plugin is used for applying custom `sysctl` settings and should only be used to change 
+        system settings that are not covered by other plugins available in *Tuned*. If the settings
+        are covered by other *Tuned* plugins, please use those plugins instead.
+        
+        The syntax for this plugin is `name=value`, where _name_ is the same as the name provided by
```
jmencak, line 23: Perhaps say `variable=value` as you're referring to the `sysctl` tool unless `name=value` is an established convention in the Tuned docs?  See `man sysctl`.  You can find some good wording  like "modify kernel parameters at runtime" in the manual page too, instead of saying applying custom `sysctl` settings.  Would saying that this modifies kernel parameters be more precise and specific?

courtneypacheco: Good point, Jiri. I will fix that. I was using the old docstring info, but you're right that the `sysctl` documentation does say `variable=value`.

jmencak:
Obsoleted by: https://github.com/redhat-performance/tuned/pull/337

PR #316, opened 2020-12-26
explicitly use /bin/bash for tuned scripts

evgeni:
`/bin/sh` is bash on Red Hat/Fedora, but other distributions (especially Debian/Ubuntu based ones) have other (POSIX compatible) shells there. Let's explicitly use `/bin/bash` for things that use bash features (the `functions` file uses some, for example).


yarda:
LGTM, sorry for delay.

PR #319, opened 2021-01-15
Lower the threading.Event's wait() timeout.

jmencak:
The current 3600s `wait()` timeout in the controller's loop is too high.  There are known instances of signals not being acted upon.  One known example is `SIGTERM` not being acted upon during system shutdown.  Lowering the timeout to 5s allows for this signal to be acted upon right after the timeout.

### DIFF
--- a/tuned/daemon/controller.py
+++ b/tuned/daemon/controller.py
@@ -61,7 +61,7 @@ class Controller(tuned.exports.interfaces.ExportableInterface):
 		if daemon:
 			self._terminate.clear()
 			# we have to pass some timeout, otherwise signals will not work
-			while not self._cmd.wait(self._terminate, 3600):
+			while not self._cmd.wait(self._terminate, 10):
 				pass
 
 		log.info("terminating controller")


yarda:
I don't have problem with it. But why 5 s? Could we go with 10 s? IIRC last time when I did some measurements I saw tuned in powertop with 5 s timeouts, but not with 10 s. And also 10 s is timeout TuneD also uses in other threads, which means that with 5 s we will increase the number of wakeups twice, but with 10 s the number of wakeups will become constant.

jmencak:
> I don't have problem with it. But why 5 s? Could we go with 10 s? IIRC last time when I did some measurements I saw tuned in powertop with 5 s timeouts, but not with 10 s. And also 10 s is timeout TuneD also uses in other threads, which means that with 5 s we will increase the number of wakeups twice, but with 10 s the number of wakeups will become constant.

Changed to 10s now.  I was planning to actually investigate the impact of this, but still didn't get round to it.  Let's keep this at 10 then and lower if we see no impact.

Thanks!

yarda:
LGTM, merging.

PR #323, opened 2021-02-18
Changes to scheduler plugin to support os interface instead of schedutils in newer python versions.

CZerta:
schedutils library becomes obsolete in newer versions, so this change allows to use interface in os module where is presented with possibility of usage of schedutils where is not.

### DIFF
--- a/tuned/plugins/plugin_scheduler.py
+++ b/tuned/plugins/plugin_scheduler.py
@@ -12,12 +12,16 @@ import perf
 import select
 import tuned.consts as consts
 import procfs
-import schedutils
 from tuned.utils.commands import commands
 import errno
 import os
 import collections
 import math
+# Check existence of scheduler API in os module
+try:
+	os.SCHED_FIFO
+except AttributeError:
+	import schedutils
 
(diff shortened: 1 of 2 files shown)


yarda:
@CZerta thanks for the PR. Just small idea, could we have two classes one for handling schedutils and the other handling os? Then just assign the object from the right class to the handler. IMHO this could be faster than the 'if' approach with 'if' in each method. I have concerns about the speed, because there is a push for optimization of the scheduler plugin (e.g. https://bugzilla.redhat.com/show_bug.cgi?id=1921738), so we shouldn't add more overhead than necessary.

tuned/plugins/plugin_scheduler.py
```diff
@@ -52,20 +56,93 @@ def __init__(self):
 		# IRQs that don't support changing CPU affinity:
 		self.unchangeable = []
 
+
```
yarda, line 59: Spurious new line.

yarda (approved):
Thanks, please fix the spurious whitespace, otherwise LGTM.

PR #324, opened 2021-02-19
Add make as BuildRequires for new fedora and rhel

CZerta:
make has been removed from BuildRoot and is needed to be add as BuildRequires for new systems.
bz#1900006

### DIFF
--- a/tuned.spec
+++ b/tuned.spec
@@ -59,6 +59,7 @@ BuildRequires: asciidoctor
 Requires(post): systemd, virt-what
 Requires(preun): systemd
 Requires(postun): systemd
+BuildRequires: make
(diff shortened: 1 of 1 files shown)


yarda:
Thanks for the PR. Could you add it unconditionally? It was always the requirement, but it was preinstalled or pulled in by some other dep, thus the requirement should be there and even in older build roots where it is preinstalled it will not cause any harm.

yarda:
Could you squash it to just one commit an force push?

yarda:
Could you move the BuildRequires up? E.g. to line 62 not to mix Requires/BuildRequires. I know the spec is a bit mess at the moment, but let's do not make it worse.

yarda:
Thanks, LGTM.

PR #331, opened 2021-03-17
Prevent unnecessary ksm* error log messages.

jmencak:
Check the presence of `ksm` and `ksmtuned` units before (un)masking them.

Prevents messages in the logs such as:
```
2021-02-24 14:46:52,399 ERROR    tuned.plugins.plugin_script: script '/usr/lib/tuned/cpu-partitioning/script.sh' error output: 'Unit ksm.service does not exist, proceeding anyway.
```

### DIFF
--- a/profiles/functions
+++ b/profiles/functions
@@ -559,6 +559,8 @@ disable_ksm()
 		if ! touch $KSM_MASK_FILE; then
 			die "failed to create $KSM_MASK_FILE"
 		fi
+		# Do not run any systemctl commands if $KSM_SERVICES units do not exist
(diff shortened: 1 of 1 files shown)


jmencak:
@yarda , do these little changes make sense?  I also tried not to re-introduce [rhbz#1622239](https://bugzilla.redhat.com/show_bug.cgi?id=1622239).

yarda (approved):
LGTM, the CI error is already fixed - according to the limitation how packit works it would require PR rebase in order to pass, which I think is not necessary.

PR #333, opened 2021-03-24
udev: explicitly sort udev properties

yarda:
Although udev properties are sorted by libudev, according
to libudev/systemd upstream this behavior cannot be relied
upon. Sort them explicitly for consistent behavior.

Resolves: rhbz#1939970

### DIFF
--- a/tuned/hardware/device_matcher_udev.py
+++ b/tuned/hardware/device_matcher_udev.py
@@ -18,7 +18,7 @@ class DeviceMatcherUdev(device_matcher.DeviceMatcher):
 		except AttributeError:
 			items = device.items()
 
-		for key, val in list(items):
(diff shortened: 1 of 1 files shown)


yanirq:
@yarda   this means the items are sorted by the key first right ? 
(if so then lgtm)

CZerta (approved):
Seems OK

yarda:
> @yarda this means the items are sorted by the key first right ?
> (if so then lgtm)

AFAIK it sorts according to the keys and if the keys are the same then it sorts according to the content. Technically, we sort string tuples, not dicts, i.e. (("key1", "val1"), ("key2", "val2"), ...), so classic python tuple comparison takes place in the sorted(), and the result should be exactly what we need.

PR #334, opened 2021-03-28
Project renamed to TuneD

yarda:
The 'D' is now capital to clarify pronunciation.

DBus service name is kept as '/Tuned' not to break
backward compatibility.

This will also need update of downstream tests
which rely on the specific output strings containing
the name.


jmencak:
This looks reasonable.  Another future cleanup I propose is replace 'plugin' -> 'plug-in' as that seems to be the term the official docs seem to use.

yarda:
I fixed two Beaker tests:
```
Regression/profile-selection-in-inactive-tuned
Sanity/bz1068699-GSS-7-0-tuned-needs-better-error-message-when
```

I wasn't able to find more affected Beaker tests, but there is still small probability that some test(s) avoided my grep queries :).

CZerta (commented):
To be really thorough, what about:
./tuned/plugins/plugin_systemd.py:100:			log.info("removing '%s' systemd tuning previously added by Tuned" % consts.SYSTEMD_CPUAFFINITY_VAR)
./tuned/plugins/plugin_systemd.py:104:	# convert cpulist from systemd syntax to Tuned syntax and unpack it
(or is is meant as DBus service?)

Rest seems ok

yarda:
> To be really thorough, what about:
> ./tuned/plugins/plugin_systemd.py100	log.info("removing '%s' systemd tuning previously added by Tuned" % consts.SYSTEMD_CPUAFFINITY_VAR)
> ./tuned/plugins/plugin_systemd.py:104: # convert cpulist from systemd syntax to Tuned syntax and unpack it
> (or is is meant as DBus service?)
> 
> Rest seems ok

Thanks, I updated it.

CZerta (approved):
Seems ok

yarda:
Finally merging, let's see what will happen :)

PR #335, opened 2021-04-07
mssql: update profile to be in sync with MS recommendations

yarda:
Resolves: rhbz#1942733

### DIFF
--- a/profiles/mssql/tuned.conf
+++ b/profiles/mssql/tuned.conf
@@ -3,15 +3,29 @@
 #
 
 [main]
-summary=Optimize for MS SQL Server
+summary=Optimize for Microsoft SQL Server
 include=throughput-performance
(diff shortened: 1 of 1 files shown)


CZerta (changes requested):
Missing line "kernel.sched_migration_cost_ns=500000" from BZ? Also different summary, but it's probably not important.

yarda:
> Missing line "kernel.sched_migration_cost_ns=500000" from BZ? Also different summary, but it's probably not important.

Good catch. It was set only on the AMD, but it should be set everywhere, I will update the PR.

yarda:
I also changed the summary, using 'Microsoft' instead of the 'MS' seems better to me.

CZerta (approved):
Seems ok

PR #336, opened 2021-04-07
plugin_net: Fix check for channels supported mode

niteshnarayanlal:
Commit 0ab1f8cc3f87a adds a check which replaces the channels mode parameters
based on the supported mode. With the older kernels, the un-supported mode
was indicated by 0, however recently that has been replaced by a string 'n/a'.
Due to this the check, in plugin_net that verifies if a replacement is
required or not fails.

Fix the check to incorporate this new string that indicates the un-supported
queue mode.

Resolves: rhbz#1943291

Signed-off-by: Nitesh Narayan Lal <nitesh@redhat.com>


niteshnarayanlal:
@yarda Hi, any comments here?

yarda:
@niteshnarayanlal LGTM, thanks.

PR #339, opened 2021-04-12
syspurpose is provides by subscription-manager on RHEL-9

yarda:

### DIFF
--- a/tuned.spec
+++ b/tuned.spec
@@ -94,8 +94,13 @@ Recommends: hdparm
 Recommends: kernel-tools
 Recommends: kmod
 %endif
+# syspurpose
+%if 0%{?rhel} > 8
+Requires: subscription-manager
+%else
 %if 0%{?rhel} > 7
 Requires: python3-syspurpose
+%endif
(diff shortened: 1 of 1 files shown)


CZerta (approved):
Seems ok

nullr0ute:
What does syspurpose do? We don't ship subscription-manager in RHEL 4 Edge and this now pulls in the entire subscription-manager/dnf stack which we don't use at all for edge.

nullr0ute:
Looking at the code this module is optional so can we at least have it as a recommends as opposed to a hard requires?

yarda:
> What does syspurpose do? We don't ship subscription-manager in RHEL 4 Edge and this now pulls in the entire subscription-manager/dnf stack which we don't use at all for edge.

It's used for autodetection whether machine is desktop/workstation and then the default TuneD profile is selected accordingly.

yarda:
> Looking at the code this module is optional so can we at least have it as a recommends as opposed to a hard requires?

Recommends should work. I will open PR for it.

PR #341, opened 2021-04-29
profiles: improve realtime verify functions

pjps:
tuned(8) realtime-virtual-{host|guest} profiles set Kernel and
KVM module parameters via script.sh plugin. These parameters are
to be verified in a verify() function, invoked by tuned-adm verify
command.

* This patch updates verify() functions to validate KVM module
  parameters.

* It moves kernel parameters to the tuned.conf file under the
  [sysfs] & [sysctl] plugin sections. And removes call to
  disable_ksm function, no longer required.

### DIFF
--- a/profiles/realtime-virtual-guest/script.sh
+++ b/profiles/realtime-virtual-guest/script.sh
@@ -2,13 +2,8 @@
 
 . /usr/lib/tuned/functions
 
-KTIMER_LOCKLESS_FILE=/sys/kernel/ktimer_lockless_check
-
 start() {
-    if [ -f $KTIMER_LOCKLESS_FILE ]; then
-        echo 1 > $KTIMER_LOCKLESS_FILE
-    fi
-    return "$?"
+    return 0
 }
 
 stop() {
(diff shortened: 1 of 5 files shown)


yarda:
Thanks for contribution.

I have question about the `/sys/module/kvm/parameters/kvmclock_periodic_sync`. Could it return '0' in the past (e.g. with some old kernel)? If yes we should also cover it for backward compatibility.

Also please add bugzilla reference to the commit message i.e.:
Resolves: rhbz#1947858

pjps:
Hello Jaroslav,

> Could it return '0' in the past (e.g. with some old kernel)? If yes we should also cover it for backward compatibility.
> Resolves: rhbz#1947858

Made the changes to check both zero(0) or 'Y/N' values.

Thank you.

yarda:
@pjps thanks, but I don't think we need multiple commits. Could you please squash it to one commit and force push?

pjps:
> @pjps thanks, but I don't think we need multiple commits. Could you please squash it to one commit and force push?

I've created a new PR from its own branch -> https://github.com/redhat-performance/tuned/pull/345

Hope it works. I need to clean-up my master branch to remove those automatic merge commits.

Thank you.

pjps:
Closing this PR here. Thank you.

PR #342, opened 2021-05-10
RHBZ#1956387 tuned-adm: Add --verbose option to verify command

pjps:
'tuned-adm verify' command validates system settings against the active tuned(8) profiles. But it does not show detailed validation results to the user. User is left to sift through the tuned(8) log file to figure it out.

Adding '--verbose' option to the tuned-adm command addresses this issue. It enables tuned(8) daemon to collect and return
detailed validation results back to the tuned-adm program. tuned-adm displays these results on the console.
    
It enables users to confirm that active tuned(8) profile settings are effective on the system.

Fixes: RHBZ#1956387

### DIFF
--- a/profiles/realtime-virtual-guest/script.sh
+++ b/profiles/realtime-virtual-guest/script.sh
@@ -2,13 +2,8 @@
 
 . /usr/lib/tuned/functions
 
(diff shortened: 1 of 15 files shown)


yarda:
@pjps could you please squash it to one commit and rebase it on the master? If it requires #341 as prerequisity, please wait until #341 is merged and then rebase on the master.

PR #345, opened 2021-05-13
RHBZ#1947858 profiles: improve realtime verify functions

pjps:
tuned(8) realtime-virtual-{host|guest} profiles set Kernel and
KVM module parameters via script.sh plugin. These parameters are
to be verified in a verify() function, invoked by tuned-adm verify
command.

* This patch updates verify() functions to validate KVM module
  parameters.

* It moves kernel parameters to the tuned.conf file under the
  [sysfs] & [sysctl] plugin sections. And removes call to
  disable_ksm function, no longer required.

Fixes: RHBZ#1947858

### DIFF
--- a/profiles/realtime-virtual-host/script.sh
+++ b/profiles/realtime-virtual-host/script.sh
@@ -2,36 +2,39 @@
 
 . /usr/lib/tuned/functions
 
-KTIMER_LOCKLESS_FILE=/sys/kernel/ktimer_lockless_check
-
 start() {
     setup_kvm_mod_low_latency
-
-    disable_ksm
-
-    if [ -f $KTIMER_LOCKLESS_FILE ]; then
-        echo 1 > $KTIMER_LOCKLESS_FILE
-    fi
-
     return 0
 }
 
 stop() {
     if [ "$1" = "full_rollback" ]; then
         teardown_kvm_mod_low_latency
-        enable_ksm
     fi
     return "$?"
 }
 
 verify() {
+    retval=0
+
+    # set via /etc/modprobe.d/kvm.conf and /etc/modprobe.d/kvm.rt.tuned.conf
     if [ -f /sys/module/kvm/parameters/kvmclock_periodic_sync ]; then
-        test "$(cat /sys/module/kvm/parameters/kvmclock_periodic_sync)" = 0
-        retval=$?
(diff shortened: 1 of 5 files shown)


profiles/realtime-virtual-host/script.sh
```diff
@@ -2,37 +2,39 @@
...
 stop() {
     if [ "$1" = "full_rollback" ]; then
         teardown_kvm_mod_low_latency
-        enable_ksm
     fi
     return "$?"
 }
 
 verify() {
+    # set via /etc/modprobe.d/kvm.conf and /etc/modprobe.d/kvm.rt.tuned.conf
```
yarda, line 18: I think retval should be explicitly set to '0' here, because otherwise you could return it on line 38 expanded to empty string, which will have the same effect as returning '0', but it may look suspicious.

profiles/realtime-virtual-host/script.sh
```diff
@@ -2,37 +2,39 @@
...
     if [ -f /sys/module/kvm/parameters/kvmclock_periodic_sync ]; then
-        test "$(cat /sys/module/kvm/parameters/kvmclock_periodic_sync)" = 0
-        retval=$?
+        kps=$(cat /sys/module/kvm/parameters/kvmclock_periodic_sync)
+        if [ "$kps" != "N" -a "$kps" != "0" ]; then
```
yarda, line 21: Can it influence the result of the verification? I.e. is the fail if one of the checks (this and ple_gap) fails? In such case you could also set the retval here (and set it later on line 29 only on failure not to mask this check).

profiles/realtime/script.sh
```diff
@@ -11,8 +11,12 @@ stop() {
 }
 
 verify() {
-    tuna -c "$TUNED_isolated_cores" -P
-    return "$?"
+    retval=0
+    if [ ! -z "$TUNED_isolated_cores" ]; then
```
yarda, line 15: with bash I think you could simplify it to"
```
if [ "$TUNED_isolated_cores" ]; then
```

pjps:
Hello @yarda,

I've made changes as suggested above. Please see if they are okay.

Thank you.

profiles/realtime-virtual-host/script.sh
pjps: done.

yarda:
Thanks, merging.

PR #346, opened 2021-05-13
RHBZ#1956387 tuned-adm: add --verbose option to verify command

pjps:
'tuned-adm verify' command validates system settings against the
active tuned(8) profiles. But it does not show detailed validation
results to the user. User is left to sift through the tuned(8) log
file to figure it out.

Adding '--verbose' option to the tuned-adm command addresses
this issue. It enables tuned(8) daemon to collect and return
detailed validation results back to the tuned-adm program.
tuned-adm displays these results on the console.

It enables users to confirm that active tuned(8) profile settings
are effective on the system.

Fixes: RHBZ#1956387

### DIFF
--- a/profiles/cpu-partitioning/00-tuned-pre-udev.sh
+++ b/profiles/cpu-partitioning/00-tuned-pre-udev.sh
@@ -4,7 +4,7 @@ type getargs >/dev/null 2>&1 || . /lib/dracut-lib.sh
 
 cpumask="$(getargs tuned.non_isolcpus)"
 
-file=/sys/devices/virtual/workqueue/cpumask
+files=$(echo /sys/devices/virtual/workqueue{/,/*/}cpumask)
 
 log()
 {
@@ -12,8 +12,12 @@ log()
 }
 
 if [ -n "$cpumask" ]; then
-  log "setting workqueue CPU mask to $cpumask"
-  if ! echo $cpumask > $file 2>/dev/null; then
-    log "ERROR: could not write workqueue CPU mask"
-  fi
+  log "setting workqueues CPU mask to $cpumask"
+  for f in $files; do
+    if [ -f $f ]; then
+      if ! echo $cpumask > $f 2>/dev/null; then
+        log "ERROR: could not write workqueue CPU mask '$cpumask' to '$f'"
+      fi
+    fi
+  done
 fi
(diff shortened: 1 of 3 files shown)


yarda (approved):
Thanks for contribution. The change LGTM. The only concern I currently have is the API change - verify_profile* methods now also returning string. I checked Cockpit Project (which is known consumer of the API) and it doesn't call these methods. Adding v2 D-Bus interface or new methods seems as unnecessary complication to me (taking into account that the 'verify' functionality is usually not much used by regular users). AFAIK we have never claimed the API to be stable so the change is probably OK.

It's bigger change, so also adding jzerdik to review it in case I missed anything. In the meantime I will run tier1 tests on the scratch build to be 100% sure there is no regression.

tuned/plugins/plugin_modules.py
```diff
@@ -85,9 +87,14 @@ def _instance_verify_static(self, instance, ignore_missing, devices):
 			mpath = "/sys/module/%s" % module
 			if not os.path.exists(mpath):
 				ret = False
-				log.error(consts.STR_VERIFY_PROFILE_FAIL % "module '%s' is not loaded" % module)
+				emsg = consts.STR_VERIFY_PROFILE_FAIL % "module '%s' is not loaded" % module
+				log.error(emsg)
+				vlog.add_log(instance.name, {module: {"message": emsg, "result": ret}})
 			else:
-				log.info(consts.STR_VERIFY_PROFILE_OK % "module '%s' is loaded" % module)
+				ret = True
```
CZerta, line 94: Should this line be here? If something already failed and this will be last run through cycle, it would overwrite False result.

tuned/verifylog.py
```diff
@@ -0,0 +1,108 @@
...
+		return self.vLog
+
+	def put_log(self):
+		vlog = self.vLog
+		self.vLog = {}
+		return vlog
+
+	def _display_device(self, log):
+		ret = True
+		for i in range(len(log)):
```
CZerta, line 39: I would rewrite these functions more pythonic/readable. Var i is used just for index, so it can be "for l in log" or smth.

tuned/verifylog.py
```diff
@@ -0,0 +1,108 @@
...
+			for cmd in list(log[i].keys()):
+				print(" %s=%s" % (cmd, log[i][cmd]["value"]))
+				for dev in log[i][cmd]["devices"]:
+					for k in dev.keys():
```
CZerta, line 43: Similar to above, why not use "for k, v in dev.items()"? Also similar in other functions down there.

yarda (changes requested):
Dropping approval, because we agreed off channel on some design changes that will be compatible with the 3rd party plugins.

PR #349, opened 2021-05-17
recommend: do not ignore chassis_type if there is no dmidecode

yarda:
Previously, if there wasn't dmidecode all 'chassis_type' matches
were ignored. This is not correct, because we cannot enforce
chassis type. Now in such cases the chassis_type is matched
against empty string.

I.e. previously the following matched in case of no dmidecode:
something_that_matched
chassis_type=.*(Notebook|Laptop|Portable).*

Now it doesn't match, but the previous behavior can be still
emulated by e.g.:
something_that_matched
chassis_type=(.*(Notebook|Laptop|Portable).*)|^$

Resolves: rhbz#1959889

### DIFF
--- a/tuned/utils/profile_recommender.py
+++ b/tuned/utils/profile_recommender.py
@@ -91,11 +91,8 @@ class ProfileRecommender:
 					elif option == "chassis_type":
 						chassis_type = self._get_chassis_type()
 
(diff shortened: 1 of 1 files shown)


CZerta (approved):
Seems ok

PR #350, opened 2021-05-24
RHBZ#1956387 tuned-adm: add --verbose option to verify command

pjps:
'tuned-adm verify' command validates system settings against the
active tuned(8) profiles. But it does not show detailed validation
results to the user. User is left to sift through the tuned(8) log
file to figure it out.

Adding '--verbose' option to the tuned-adm command addresses
this issue. It enables tuned(8) daemon to collect and return
detailed validation results back to the tuned-adm program.
tuned-adm displays these results on the console.

It enables users to confirm that active tuned(8) profile settings
are effective on the system.

Fixes: RHBZ#1956387

### DIFF
--- a/ci.fmf
+++ b/ci.fmf
@@ -1,4 +1,4 @@
 discover:
     how: fmf
 execute:
-    how: beakerlib
+    how: tmt
--- a/tests/beakerlib/Program-tuned-tried-to-access-dev-mem-between/main.fmf
+++ b/tests/beakerlib/Program-tuned-tried-to-access-dev-mem-between/main.fmf
@@ -7,3 +7,4 @@ duration: 20m
 relevancy: |
     distro = rhel-4, rhel-5, rhel-6: False
 summary: Test for BZ#1688371 (Program tuned tried to access /dev/mem between)
+framework: beakerlib
--- a/tests/beakerlib/Tuned-takes-too-long-to-reload-start-when-ulimit/main.fmf
+++ b/tests/beakerlib/Tuned-takes-too-long-to-reload-start-when-ulimit/main.fmf
@@ -7,3 +7,4 @@ duration: 20m
 relevancy: |
     distro = rhel-4, rhel-5, rhel-6: False
 summary: Test for BZ#1663412 (Tuned takes too long to reload/start when \"ulimit)
+framework: beakerlib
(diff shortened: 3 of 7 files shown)


tuned/plugins/plugin_modules.py
```diff
@@ -85,9 +87,14 @@ def _instance_verify_static(self, instance, ignore_missing, devices):
...
 			if not os.path.exists(mpath):
 				ret = False
-				log.error(consts.STR_VERIFY_PROFILE_FAIL % "module '%s' is not loaded" % module)
+				emsg = consts.STR_VERIFY_PROFILE_FAIL % "module '%s' is not loaded" % module
+				log.error(emsg)
+				vlog.add_log(instance.name, module,
+                                                    emsg, result=ret)
 			else:
-				log.info(consts.STR_VERIFY_PROFILE_OK % "module '%s' is loaded" % module)
+				ret = True
```
CZerta, line 95: I still have problem with this. IE. instance._modules has keys = (module_a, module_b, module_c), module_a and module_b fails and module_c pass. In current implementation ret is False, because it's not overwrite to True again. In new implementation as module_c is processed last, ret is True and it changes result of this function.

pjps: Fixed and repushed the commit.

Thank you.

tuned/admin/admin.py
```diff
@@ -354,24 +355,36 @@ def _action_recommend_profile(self):
 		print(self._profile_recommender.recommend())
 		return True
 
-	def _action_dbus_verify_profile(self, ignore_missing):
+	def _action_dbus_verify_profile(self, ignore_missing, verbose=False):
+		profile_name = self._get_active_profile()
+		if (not profile_name):
```
yarda, line 360: This introduces regression in the following situation:
- no active profile set
- post-load profile set to some profile

Reproducer:
```
# systemctl stop tuned
# echo balanced > /etc/tuned/post_loaded_profile
# > /etc/tuned/active_profile
# echo manual > /etc/tuned/profile_mode
# systemctl start tuned
# tuned-adm verify
```
Note: test coverage for such scenario would be handy, but it's not related to this PR
```

yarda (changes requested):

PR #352, opened 2021-06-02
RHBZ#1817827 profiles: add isolate_managed_irq & netdev_queue_count variables

pjps:
For realtime KVM host and guests it is important to use housekeeping
CPUs for processing interrupts (IRQs) and network packets transmission.
So that isolated CPUs focus on the -realtime tasks assigned to them.

The 'isolate_managed_irq' and 'netdev_queue_count' variables help
to configure kernels towards this end. Users can define these
variables via tuned(8) configuration files.

Fixes: RHBZ#1817827

### DIFF
--- a/profiles/realtime-virtual-guest/realtime-virtual-guest-variables.conf
+++ b/profiles/realtime-virtual-guest/realtime-virtual-guest-variables.conf
@@ -1,4 +1,23 @@
+#
+# Variable settings below override the definitions from the
+# /etc/tuned/realtime-variables.conf file.
+#
 # Examples:
 # isolated_cores=2,4-7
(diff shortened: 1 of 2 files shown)


yarda (changes requested):
I think you should mention in the comments that these settings override settings from the `/etc/tuned/realtime-variables.conf`, e.g. please add comment to the top of the files.

pjps:
> I think you should mention in the comments that these settings override settings from the `/etc/tuned/realtime-variables.conf`, e.g. please add comment to the top of the files.

Yes, done.

yarda:
Thanks merging.

PR #354, opened 2021-06-03
network-throughput: Remove explicit udp_mem amount

superjamie:
Explicitly setting a value for udp_mem does not make sense.

An explicit value does not take into account how much RAM the user's system
has, so setting values here like "12G 16G 64G" works for few specific use
cases but makes no sense for others.

This actually could cause problems if applied on a system with <16G RAM
and UDP protocol usage is allowed to grow as large as it likes, causing
kernel memory pressure in other areas.

Remove these values for now. If still required, a better way to calculate
a larger value based on system RAM can be implemented later.

Signed-off-by: Jamie Bainbridge <jamie.bainbridge@gmail.com>


yarda:
Thanks for contribution. I handed it to the team dealing with network performance tuning.

yarda:
Approved by perf team, merging, thanks.

PR #356, opened 2021-06-07
RHBZ#1820626 profiles: enable isolate_managed_irq by default

pjps:
Enable 'isolate_managed_irq' parameter by default. It is important
for -realtime systems. When it is enabled, kernel directs interrupts
to the housekeeping CPUs and thus reduces latencies for the isolated
ones.

Fixes: RHBZ#1820626


yarda:
Thanks, LGTM. According to the previous discussion it seems if kernel doesn't support it, it can result in degraded performance, but it shouldn't prevent the kernel from booting.

PR #357, opened 2021-06-09
profiles: fix loading multiple profiles if there are repeating profiles

yarda:
The original idea behind this patch is credited to
Jiří Mencák <jmencak@redhat.com>.

Resolves: rhbz#1825882


yarda:
I am not sure whether this simplified profile loading mechanism could bring some unexpected behavior e.g. in combination with the 'replace=true', but it's definitely improvement over what we had.

CZerta (approved):
Seems ok, definitely improvement

jmencak:
Thank you Jaroslav for an improved fix and Honzo for the review!  I did some extra testing and I'm unable to find/think of any side effects at this point.

PR #360, opened 2021-06-20
Fix conditional order for setting net device param

yanirq:
The conditional for setting a net device parameter has
a faulty order where it can potentially try to first cast a string to int
in a wrong manner and only then check if the value is a specific integer value.

current order:
1. check if a net device has a combined channel containing 0
2. check if a net device has a combined channel containing 'n/a'

fixed order:
robust check to see if device parameter is either 0 or 'n/a'

Signed-off-by: Yanir Quinn <yquinn@redhat.com>

### DIFF
--- a/tuned/plugins/plugin_net.py
+++ b/tuned/plugins/plugin_net.py
@@ -397,7 +397,7 @@ class NetTuningPlugin(base.Plugin):
 		if dev_params:
 			self._check_device_support(context, d, device, dev_params)
 			# replace the channel parameters based on the device support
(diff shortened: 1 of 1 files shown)


CZerta (approved):
Seems ok, definitely reasonable

yarda (changes requested):
Although unlikely it could still fail with traceback if there is different string returned than 'n/a'. Could you rewrite it to something more robust and more simple like e.g.:
```
if context == "channels" and str(dev_params[next(iter(d))]) in ["n/a", "0"]:
```

yarda:
@yanirq, could you please merge the commits to one and force push?

yarda (approved):
LGTM, thanks.

PR #361, opened 2021-06-22
bootloader: support for systems that don't have grub installed

adriaan42:
This makes some changes to address systems that don't support changing
of grub configurations, e.g. embedded systems, where the cmdline is set in
other ways, but we still want to process it and verify against the tuning profile.

* Introduce new option "skip_grub_config". When this is set, no changes to grub
  config are attempted. cmdline options are still processed, and the result is
  used to verify the current cmdline.
* Improve reporting on expected/found/missing cmdline arguments.
* Add "no_error = True" to the call that reads /etc/default/grub in _bls_enabled().

### DIFF
--- a/tuned/plugins/plugin_bootloader.py
+++ b/tuned/plugins/plugin_bootloader.py
@@ -31,6 +31,7 @@ class BootloaderPlugin(base.Plugin):
 		instance._has_static_tuning = True
 		# controls grub2_cfg rewrites in _instance_post_static
 		self.update_grub2_cfg = False
+		self._skip_grub_config_val = False
 		self._initrd_remove_dir = False
 		self._initrd_dst_img_val = None
 		self._cmdline_val = ""
@@ -52,6 +53,7 @@ class BootloaderPlugin(base.Plugin):
 			"initrd_add_dir": None,
 			"initrd_remove_dir": None,
 			"cmdline": None,
+			"skip_grub_config": None,
 		}
 
 	@staticmethod
@@ -184,7 +186,7 @@ class BootloaderPlugin(base.Plugin):
 		return cfg_files
 
 	def _bls_enabled(self):
-		grub2_default_env = self._cmd.read_file(consts.GRUB2_DEFAULT_ENV_FILE)
+		grub2_default_env = self._cmd.read_file(consts.GRUB2_DEFAULT_ENV_FILE, no_error = True)
 		if len(grub2_default_env) <= 0:
 			log.info("cannot read '%s'" % consts.GRUB2_DEFAULT_ENV_FILE)
 			return False
@@ -220,7 +222,7 @@ class BootloaderPlugin(base.Plugin):
 		self._patch_bootcmdline({consts.BOOT_CMDLINE_TUNED_VAR: "", consts.BOOT_CMDLINE_KARGS_DELETED_VAR: ""})
 
 	def _instance_unapply_static(self, instance, full_rollback = False):
-		if full_rollback:
+		if full_rollback and not self._skip_grub_config_val:
 			if self._rpm_ostree:
 				log.info("removing rpm-ostree tuning previously added by Tuned")
 				self._remove_rpm_ostree_tuning()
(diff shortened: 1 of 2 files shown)


tuned/plugins/plugin_bootloader.py
```diff
@@ -355,20 +357,46 @@ def _cmdline(self, enabling, value, verify, ignore_missing):
...
+			if len(missing_set) == 0:
 				log.info(consts.STR_VERIFY_PROFILE_VALUE_OK % ("cmdline", str(value_set)))
 				return True
 			else:
-				log.error(consts.STR_VERIFY_PROFILE_VALUE_FAIL % ("cmdline", str(cmdline_intersect), str(value_set)))
+				cmdline_dict = {v.split("=", 1)[0]: v for v in cmdline_set}
+				for m in missing_set:
+					arg = m.split("=", 1)[0]
+					if not arg in cmdline_dict:
+						log.error(consts.STR_VERIFY_PROFILE_VALUE_FAIL % ("cmdline arg "+arg, "<missing>", m))
```
yarda, line 369: Please output arg value in quotes to be consistent with other logging, i.e.:
`cmdline arg 'ARGUMENT'` instead of `cmdline arg ARGUMENT`

tuned/plugins/plugin_bootloader.py
```diff
@@ -355,20 +357,46 @@ def _cmdline(self, enabling, value, verify, ignore_missing):
...
+					else:
+						log.error(consts.STR_VERIFY_PROFILE_VALUE_FAIL % ("cmdline arg "+arg, cmdline_dict[arg], m))
```
yarda, line 371: The same quoting as in the previous log.error.

tuned/plugins/plugin_bootloader.py
```diff
@@ -355,20 +357,46 @@ def _cmdline(self, enabling, value, verify, ignore_missing):
...
 			log.info("installing additional boot command line parameters to grub2")
 			self.update_grub2_cfg = True
 			self._cmdline_val = v
 
+	@command_custom("skip_grub_config", per_device = False, priority = 10)
+	def _skip_grub_config(self, enabling, value, verify, ignore_missing):
+		if verify:
+			return None
+		if enabling and value is not None:
+			if value.lower() in "yes true t 1".split():
```
yarda, line 385: Please use `self._cmd.get_bool(value)` to be consistent with bools.

tuned/plugins/plugin_bootloader.py
```diff
@@ -355,20 +357,46 @@ def _cmdline(self, enabling, value, verify, ignore_missing):
...
+				log.info("skipping any modification of grub config")
+				self._skip_grub_config_val = True
+
 	def _instance_post_static(self, instance, enabling):
-		if enabling and self.update_grub2_cfg:
+		if enabling and self._skip_grub_config_val:
```
yarda, line 390: Lines 390 - 399: do we need it? I.e. with the `skip_grub_config = True` should we care what's already there? There is 'verification' for the check. Or if we care, shouldn't we throw exception (e.g. InvalidProfileException) to stop loading of the profile? I am just thinking, this comment is not request for change.

yarda (changes requested):
Thanks for contribution. There are few minor things worth changing to be consistent with the current code (see the inline comments). And I am not sure whether we should run verification upon loading of the profile.

Please don't care about the CI failure for the Rawhide. It's unrelated and it should be fixed by the https://github.com/redhat-performance/tuned/pull/365

yarda:
Could you please rebase and fix conflicts?

tuned/plugins/plugin_bootloader.py
adriaan42: I added the quotes, but I'm not sure the result is what you intended... it now has nested quotes:
```
ERROR    tuned.plugins.plugin_bootloader: verify: failed: 'cmdline arg 'intel_pstate'' = '<missing>', expected 'intel_pstate=disable'
```
Examples from other plugins:
```
DEBUG    tuned.plugins.plugin_scheduler: Read SMP affinity of IRQ '39': '[0, 1, 2]'
ERROR    tuned.plugins.plugin_scheduler: verify: failed: 'SMP affinity of IRQ 39' = '[0, 1, 2]', expected '[0, 1, 3]'
[...]
DEBUG    tuned.plugins.plugin_sysctl: Value of sysctl parameter 'kernel.timer_migration' is '0'
INFO     tuned.plugins.base: verify: passed: 'kernel.timer_migration' = ''
```

We could simply leave out the "cmdline arg" and make it:
```
ERROR    tuned.plugins.plugin_bootloader: verify: failed: 'intel_pstate' = '<missing>', expected 'intel_pstate=disable'
```

tuned/plugins/plugin_bootloader.py
adriaan42: Yes, I was wondering the same. I don't think profile loading should fail because of a wrong cmdline. But I would be ok with not checking during profile load, and require an explicit `tuned-adm verify` instead.

adriaan42:
> Could you please rebase and fix conflicts?

I have rebased, and see that the new rpm_ostree changes intersect in some if/else statements (`_instance_unapply_static` and `_instance_post_static`). Could you double-check that I did not break the logic here?

CZerta:
@adriaan42 It looks fine in regard to rpm-ostree changes.

tuned/plugins/plugin_bootloader.py
yarda: I don't have strong opinion on it, I think we could merge it as it is. @CZerta what's your opinion?

tuned/plugins/plugin_bootloader.py
yarda: You are right, please drop the nested quotes, the following could work, but I have problem that it's not consistent, i.e.  `'intel_pstate' = '<missing>'` vs `'intel_pstate=disable'`:
```
ERROR    tuned.plugins.plugin_bootloader: verify: failed: cmdline arg 'intel_pstate' = '<missing>', expected 'intel_pstate=disable'
```
Could you rewrite it with the one level quotes and consistent output, e.g.:
```
ERROR    tuned.plugins.plugin_bootloader: verify: failed: cmdline arg 'intel_pstate' = '<missing>', expected 'intel_pstate' = 'disable'
```
or:
```
ERROR    tuned.plugins.plugin_bootloader: verify: failed: cmdline arg 'intel_pstate = <missing>', expected 'intel_pstate = disable'
```
Or similarly?

adriaan42: The trouble comes with using the format string `consts.STR_VERIFY_PROFILE_VALUE_FAIL`, which is a bit limiting here.

If it is ok that I don't use it, then I would like to distinguish the two cases "missing" and "set to something else", and make the output:
```
ERROR    tuned.plugins.plugin_bootloader: verify: failed: cmdline arg 'intel_pstate' is missing, expected 'intel_pstate=disable'
```
and
```
ERROR    tuned.plugins.plugin_bootloader: verify: failed: cmdline arg 'intel_pstate=something_else', expected 'intel_pstate=disable'
```

yarda: Could you use the `consts.STR_VERIFY_PROFILE_VALUE_FAIL` for the case where it's different and introduce new const, e.g. `consts.STR_VERIFY_PROFILE_VALUE_FAIL_MISSING` for the missing case?

adriaan42: The existing `STR_VERIFY_PROFILE_VALUE_FAIL` still seems a bit awkward (it would give `'cmdline arg intel_pstate' = 'something_else'`, and I think it would be better to show the string exactly as found in the cmdline). So I'd rather introduce two new format strings (`STR_VERIFY_PROFILE_CMDLINE_FAIL` and `STR_VERIFY_PROFILE_CMDLINE_FAIL_MISSING`?) to produce the messages as shown in my previous comment.

tuned/plugins/plugin_bootloader.py
CZerta: I don't have strong opinion too, but probably wouldn't verify it either. Without verification it can theoretically be used in example on systems without "/proc/cmdline" file and won't log error.

adriaan42: I have removed the verification during profile loading.

tuned/plugins/plugin_bootloader.py
adriaan42: I pushed an update with two new format strings.

yarda:
LGTM, merging, thanks for contribution. And sorry for delay, it was public holiday here.

PR #364, opened 2021-06-28
realtime-virtual-*: Set irqaffinity through kernel command line

vianpl:
I/O APIC, one of intel's interrupt controllers, can only change the
affinity on an IRQ line during an EOI (end of interrupt)[1]. In other
words, on boot, although tuned will set a preferred IRQ affinity, it
might not be applied until the next time the interrupt is triggered.
This might be right away, or hours later while oslat or a
latency-sensitive workload is running.

To mitigate this set the 'irqaffinity='kernel command-line[2] option to
match the non-isolated CPU mask when running the realtime-virtual-*
profiles. This way we'll get the right affinity regardless of the
situation stated above.

[1] For more reference see kernel's ioapic_ack_level() function and the
    IRQD_SETAFFINITY_PENDING flag.

[2] Actually, this is why irqaffinity was introduced in the first place,
    see kernel commit fbf198030e0b0.

Resolves: rhbz#1974820

### DIFF
--- a/profiles/realtime-virtual-guest/tuned.conf
+++ b/profiles/realtime-virtual-guest/tuned.conf
@@ -16,6 +16,7 @@ assert1=${f:assertion_non_equal:isolated_cores are set:${isolated_cores}:${isola
 
 isolated_cores_expanded=${f:cpulist_unpack:${isolated_cores}}
 isolated_cores_online_expanded=${f:cpulist_online:${isolated_cores}}
+non_isolated_cores=${f:cpulist_invert:${isolated_cores}}
 
 # Fail if isolated_cores contains CPUs which are not online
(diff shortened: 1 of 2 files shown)


yarda (changes requested):
The [2] link in the commit message is not referenced (I think it should be referenced from the second paragraph instead of [1]). Otherwise the commit LGTM (from the TuneD perspective).

vianpl:
> The [2] link in the commit message is not referenced (I think it should be referenced from the second paragraph instead of [1]). 

Yes, my bad.

> Otherwise the commit LGTM (from the TuneD perspective).

Thanks for the review!

yarda:
LGTM, merging, thanks.

PR #365, opened 2021-06-28
scheduler: abstract sched_ and numa_ knobs

yarda:
New kernels (5.13 and newer) moved some sched_ and numa_ knobs from
the sysctl to the debugfs, thus add and abstract these knobs under the
scheduler plugin. With help of this abstraction it will write
the tuning to the correct place according to the kernel used.

Example:
[scheduler]
sched_migration_cost_ns = 500000

Will work on the old kernel the same way as:
[sysctl]
kernel.sched_migration_cost_ns = 500000

I.e. it will write '500000' to the:
/proc/sys/kernel/sched_migration_cost_ns

And on the new kernel it will write '500000' to the:
/sys/kernel/debug/sched/migration_cost_ns

Also updated TuneD profiles.

Resolves: rhbz#1952687

### DIFF
--- a/tuned/plugins/plugin_scheduler.py
+++ b/tuned/plugins/plugin_scheduler.py
@@ -149,6 +149,10 @@ class SchedulerPlugin(base.Plugin):
 			self._daemon = global_cfg.get_bool(consts.CFG_DAEMON, consts.CFG_DEF_DAEMON)
 			self._sleep_interval = int(global_cfg.get(consts.CFG_SLEEP_INTERVAL, consts.CFG_DEF_SLEEP_INTERVAL))
 		self._cmd = commands()
+		# helper variable utilized for showing hint only once that the error may be caused by Secure Boot
+		self._secure_boot_hint = None
+		# paths cache for sched_ and numa_ tunings
+		self._sched_knob_paths_cache = {}
 		# default is to whitelist all and blacklist none
 		self._ps_whitelist = ".*"
 		self._ps_blacklist = ""
@@ -252,6 +256,16 @@ class SchedulerPlugin(base.Plugin):
 			"default_irq_smp_affinity": "calc",
 			"perf_mmap_pages": None,
 			"perf_process_fork": "false",
+			"sched_min_granularity_ns": None,
+			"sched_latency_ns": None,
+			"sched_wakeup_granularity_ns": None,
+			"sched_tunable_scaling": None,
+			"sched_migration_cost_ns": None,
+			"sched_nr_migrate": None,
+			"numa_balancing_scan_delay_ms": None,
+			"numa_balancing_scan_period_min_ms": None,
+			"numa_balancing_scan_period_max_ms": None,
+			"numa_balancing_scan_size_mb": None
 		}
 
 	def _sanitize_cgroup_path(self, value):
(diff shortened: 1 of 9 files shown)


yarda:
As a bonus the rawhide tests should now pass (if not broken by something else in the meantime :)

yarda:
It seems it will not work in the kernel_lockdown, i.e. with the EFI secure boot enabled. In this mode accessing the debugfs is restricted. I will have to drop the tuning which are touching the debugfs for the Rawhide test to pass.

tuned/plugins/plugin_scheduler.py
```diff
@@ -1065,3 +1075,99 @@ def _isolated_cores(self, enabling, value, verify, ignore_missing):
...
+	def _get_sched_nr_migrate(self):
+		return self._get_sched_knob("sched", "nr_migrate")
+
+	@command_set("sched_nr_migrate")
+	def _set_sched_nr_migrate(self, value, sim):
+		return self._set_sched_knob("sched", "nr_migrate", value, sim)
+
+	@command_get("numa_balancing_scan_delay_ms")
+	def _get_numa_balancing_scan_delay_ms(self):
+		return self._get_sched_knob("numa", "balancing_scan_delay_ms")
```
CZerta, line 1145: This will create path "/sys/kernel/debug/numa/balancing_scan_delay_ms", but path in BZ (and also on system I tried) is "/sys/kernel/debug/sched/numa_balancing/scan_delay_ms". Same for other numa*.

CZerta (changes requested):

tuned/plugins/plugin_scheduler.py
yarda: Good catch.

yarda:
Thanks for review.

yarda:
I also did the following changes:
* Inverted the priority for checking the sysctl/debugfs support. Now the sysctl is checked first. It's because debugfs may not work with the Secure Boot or kernel_lockdown, so some people could have patches to re-enable sysctl, thus check it first.
* Added "higher level" error logging for the read/write - it's for users not being confused from where the error is coming from.
* Added one time logged hint about Secure Boot and kernel_lockdown - this could save people some time with debugging.

tuned/plugins/plugin_scheduler.py
```diff
@@ -1065,3 +1077,112 @@ def _isolated_cores(self, enabling, value, verify, ignore_missing):
 			# Restoring processes' affinity is done in
 			# _instance_unapply_static()
 			self._restore_all_irq_affinity()
+
+	def _get_sched_knob_path(self, prefix, namespace, knob):
+		path = "/proc/sys/kernel/%s_%s" % (namespace, knob)
```
CZerta, line 1082: Maybe some optimization idea. Cannot we store this path or the other* on self, so we does not need to check for existence of file every time? (Probably depends on possibility that some of the files can be on old location and some on new)
\* - in format "/sys/kernel/debug/%s/%s" % (prefix+namespace, knob)

Rest LGTM

CZerta (approved):

tuned/plugins/plugin_scheduler.py
yarda: I was thinking about caching, but I dropped this idea in the PoC, because it's only few paths which will be checked just during the profile load and the cache uses some RSS. 

I added the cache in the latest PR version.

yarda:
I added one related commit which drops the sched_ tuning from several TuneD profiles. See the associated bugzilla for details. I would like to get it to RHEL-8 as well (if approved). If not, we will use distro patch to revert this specific commit in RHEL-8.

yarda:
And as a bonus all tests now pass.

CZerta (approved):
LGTM

yarda:
Thanks, merging.

PR #368, opened 2021-07-15
scheduler: new option cgroup_ps_blacklist

yarda:
This option allows skipping processes belonging to the blacklisted
cgroups. It matches the regular expression against items from the
/proc/PID/cgroups. Items/lines from the /proc/PID/cgroups are separated
by commas ','. Each item consists of the:
hierarchy-ID:controller-list:cgroup-path

Example of the content on which the regular expression is run:
10:hugetlb:/,9:perf_event:/,8:blkio:/

For cgroups v2 the hierarchy-ID is 0 and the controller-list is ''.
For details see man cgroups.7. The only difference from the man
cgroups.7 is that it uses commas for separation of the items instead
of the new lines.

Multiple regular expressions can be separated by the semicolon ';'.

Examples:
[scheduler]
isolated_cores=1
cgroup_ps_blacklist=\b/daemons\b

It will move all processes away from the core 1 except processes which
belongs to the cgroup '/daemons'. The '\b' is regular expression
metacharacter that matches word boundary (i.e. it matches only
'/daemons', not e.g. '/daemonset' or '/group/daemons'). In this example
we do not care about the hierarchy-ID and the controller-list.

[scheduler]
isolated_cores=1
cgroup_ps_blacklist=\b8:blkio:/\b

In this example it skips processes belonging to the cgroup '/',
with hierarchy-ID 8 and controller-list blkio.

[scheduler]
isolated_cores=1
cgroup_ps_blacklist=\b/daemons\b;\b/test\b

In this example two regular expressions are used which tries to match
'/daemons' and '/test' cgroup-path. If either matches (i.e. the OR operator),
the process is skipped (i.e. not moved away from the core 1).

Resolves: rhbz#1980715

### DIFF
--- a/tuned/plugins/plugin_scheduler.py
+++ b/tuned/plugins/plugin_scheduler.py
@@ -156,6 +156,7 @@ class SchedulerPlugin(base.Plugin):
 		# default is to whitelist all and blacklist none
 		self._ps_whitelist = ".*"
 		self._ps_blacklist = ""
+		self._cgroup_ps_blacklist_re = ""
 		self._cpus = perf.cpu_map()
 		self._scheduler_storage_key = self._storage_key(
 				command_name = "scheduler")
@@ -251,6 +252,7 @@ class SchedulerPlugin(base.Plugin):
 			"cgroup_mount_point_init": False,
 			"cgroup_groups_init": True,
 			"cgroup_for_isolated_cores": None,
+			"cgroup_ps_blacklist": None,
 			"ps_whitelist": None,
 			"ps_blacklist": None,
 			"default_irq_smp_affinity": "calc",
@@ -811,6 +813,14 @@ class SchedulerPlugin(base.Plugin):
 							elif event.type == perf.RECORD_EXIT:
 								self._remove_pid(instance, int(event.tid))
 
+	@command_custom("cgroup_ps_blacklist", per_device = False)
+	def _cgroup_ps_blacklist(self, enabling, value, verify, ignore_missing):
+		# currently unsupported
+		if verify:
+			return None
+		if enabling and value is not None:
+			self._cgroup_ps_blacklist_re = "|".join(["(%s)" % v for v in re.split(r"(?<!\\);", str(value))])
+
 	@command_custom("ps_whitelist", per_device = False)
 	def _ps_whitelist(self, enabling, value, verify, ignore_missing):
 		# currently unsupported
@@ -886,6 +896,9 @@ class SchedulerPlugin(base.Plugin):
 		if self._ps_blacklist != "":
 			psl = [v for v in psl if re.search(self._ps_blacklist,
 					self._get_stat_comm(v)) is None]
+		if self._cgroup_ps_blacklist_re != "":
+			psl = [v for v in psl if re.search(self._cgroup_ps_blacklist_re,
+					self._get_stat_cgroup(v)) is None]
 		psd = dict([(v.pid, v) for v in psl])
 		for pid in psd:
 			try:
@@ -910,6 +923,12 @@ class SchedulerPlugin(base.Plugin):
 				self._set_all_obj_affinity(
 						psd[pid]["threads"].values(),
 						affinity, True)
+
+	def _get_stat_cgroup(self, o):
+		try:
+			return o["cgroups"]
+		except (OSError, IOError, KeyError):
+			return ""
 
 	def _get_stat_comm(self, o):
 		try:


jmencak:
Many thanks for this PR, @yarda !  I'll be reviewing and testing this today.  Do we also want something like `cgroup_ps_whitelist` to be consistent with `ps_whitelist`?

yarda:
> Many thanks for this PR, @yarda ! I'll be reviewing and testing this today. Do we also want something like `cgroup_ps_whitelist` to be consistent with `ps_whitelist`?

I think it is needed only in case there is a need for features/options to be orthogonal. The whitelist feature will have some overhead (especially if we will add later support for the runtime processing through the perf). With the blacklist there is just the check if the regex is not empty, process it, otherwise skip. I.e. nearly no overhead if the feature is not used. With the whitelist it will have to match the regex all the time (or we could add some optimization to skip the check in the case the regex is '.*'). And if you want to pin just some cgroups to some cpuset, you can do it directly through the cgroups (or through the cgroups feature of TuneD).

jmencak:
I've tested this PR and it works for both cgroup v1 and v2.  It also solves my usecase and I can confirm it solves rhbz#1980715. The `cgroup_ps_whitelist` would be nice-to-have, however, at the moment I do not see any real need for implementing it at this point.  The only concern I had was the separation cgroup hierarchies from the /proc/PID/cgroups by '',", as the same separator is used by the controller-list (hierarchy-ID:controller-list:cgroup-path, e.g. 5:cpuacct,cpu,cpuset:/daemons in cgroup v1) and that might make regular expression matching a bit trickier.  Nevertheless, I believe these are minor concerns and unless @CZerta has any objections, I believe this PR is good to go.  Thanks @yarda !

CZerta:
@jmencak Did you try regexps with '\b' symbol? It seams it does not work for me, rest LGTM

jmencak:
> @jmencak Did you try regexps with '\b' symbol? It seams it does not work for me, rest LGTM

Yeah, the example `cgroup_ps_blacklist=\b/daemons\b;\b/test\b` is a bit misleading and needs fixed.  I think you want something like:
`cgroup_ps_blacklist=\bdaemons\b;\btest\b` instead (/cc @yarda).  Then it works.

CZerta (approved):
Silly me. LGTM

MarSik:
Folks? I see this was approved 7 days ago, how can we unblock the merge?

jmencak:
From my perspective we can merge this after @yarda fixes the commit message as the examples with '\b' don't quite work.  See [my docs of this functionality](https://github.com/redhat-performance/tuned/pull/337/commits/ce23791eb1be350d2fc983d28dc5e0143951a908) that hopefully should work.

yarda:
> Folks? I see this was approved 7 days ago, how can we unblock the merge?

We had feature freeze between 2.16.0 RC and stable (which was released yesterday). AFAIK @jmencak is cherry picking patches he is interested in, so it shouldn't block him.

jmencak:
> > Folks? I see this was approved 7 days ago, how can we unblock the merge?
> 
> We had feature freeze between 2.16.0 RC and stable (which was released yesterday). AFAIK @jmencak is cherry picking patches he is interested in, so it shouldn't block him.

Well, yes and no.  I cherry-pick stuff that *merged* upstream as until the stuff merges there's actually no guaranteed it will merge. :)

yarda:
Rebased and updated the commit message.

jmencak:
> Rebased and updated the commit message.

Thanks for the effort, Jardo, but I think it is still not quite right.  `\b` matches word boundaries, so `/\b` in will not work in `cgroup_ps_blacklist=\b8:blkio:/\b`.  Example:

```
$ echo '10:hugetlb:/,9:perf_event:/,8:blkio:/,7:st_else:/' | grep -P '\b8:blkio:/\b'
$ echo '10:hugetlb:/,9:perf_event:/,8:blkio:/,7:st_else:/' | grep -P '\b8:blkio:/,'
10:hugetlb:/,9:perf_event:/,8:blkio:/,7:st_else:/
```

Also, I believe the description in the PR should also be probably adjusted so that people blindly copy&pasting from PRs are not bit by this.

yarda:
> > Rebased and updated the commit message.
> 
> Thanks for the effort, Jardo, but I think it is still not quite right. `\b` matches word boundaries, so `/\b` in will not work in `cgroup_ps_blacklist=\b8:blkio:/\b`. Example:
> 
> ```
> $ echo '10:hugetlb:/,9:perf_event:/,8:blkio:/,7:st_else:/' | grep -P '\b8:blkio:/\b'
> $ echo '10:hugetlb:/,9:perf_event:/,8:blkio:/,7:st_else:/' | grep -P '\b8:blkio:/,'
> 10:hugetlb:/,9:perf_event:/,8:blkio:/,7:st_else:/
> ```
> 
> Also, I believe the description in the PR should also be probably adjusted so that people blindly copy&pasting from PRs are not bit by this.

Good catch, check now.

jmencak:
This will work, many thanks!
/lgtm

PR #369, opened 2021-07-19
scheduler: fix construction of the process name regex

adriaan42:
To optimize process name matching, the regexes of all groups are combined
into one (r1)|(r2)|..., and after matching the capture is used to determine
which of the original regexes matched. This means that the individual
group regexes must not contain any captures. The current solution replaces
all "(" and ")" with "\\(" and "\\)".

Thus, if we want a group to match "thread(1|2|3)", this will turn into
"thread\\(1|2|3\\)", a regex that matches the three strings "thread(1",
"2" and "3)", which is definitely not what is intended.

This is fixed by instead replacing any "(" by "(?:", making the group
non-capturing. Replacement is not performed if the "(" is preceded by "\\"
or follwed by "?".

### DIFF
--- a/tuned/plugins/plugin_scheduler.py
+++ b/tuned/plugins/plugin_scheduler.py
@@ -660,8 +660,9 @@ class SchedulerPlugin(base.Plugin):
 			sched = dict([(pid, (cmd, option, scheduler, priority, affinity, regex))
 					for pid, cmd in processes])
 			sched_all.update(sched)
-			regex = str(regex).replace("(", r"\(")
-			regex = regex.replace(")", r"\)")
+			# make any contained regexes non-capturing: replace "(" with "(?:",
+			# unless the "(" is preceded by "\" or followed by "?"
+			regex = re.sub(r"(?<!\\)\((?!\?)", "(?:", str(regex))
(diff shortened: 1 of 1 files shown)


CZerta:
Probably good idea, but it can cause problems for people that count on current behavior. As now '(' in regex gets escaped and with this change it gets replaced even without ending bracket, so not it just change current meaning, but it also can cause invalidity of combined regex

adriaan42:
Actually, the validity of the regex is checked a few lines above with an `re.compile()`, so there can't be any unmatched, unescaped brackets.

The current behavior would in fact break if there are any (correctly) escaped `\(` or `\)` in the original regex, because it turns them into `\\(` and `\\)`.

CZerta:
> Actually, the validity of the regex is checked a few lines above with an `re.compile()`, so there can't be any unmatched, unescaped brackets.
> 
> The current behavior would in fact break if there are any (correctly) escaped `\(` or `\)` in the original regex, because it turns them into `\\(` and `\\)`.

Oh, you're right

CZerta (approved):
Seems OK

yarda:
LGTM, merging. Thanks for contribution.

PR #370, opened 2021-07-19
scheduler: always set/restore the default irq affinity first

adriaan42:
The default affinity is the one used for all future irqs. It needs
to be set first, otherwise there is a potential race condition,
where a newly registered irq gets the old default affinity,
and is not touched by the setting of individual affinities.

### DIFF
--- a/tuned/plugins/plugin_scheduler.py
+++ b/tuned/plugins/plugin_scheduler.py
@@ -965,6 +965,19 @@ class SchedulerPlugin(base.Plugin):
 
 	def _set_all_irq_affinity(self, affinity):
 		irq_original = IRQAffinities()
+
+		# default affinity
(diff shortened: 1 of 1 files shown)


CZerta:
Maybe I understand it completely wrong, but doesn't this change cause race condition of it's own?

1) You set new default
2) You get irqs with "irqs = procfs.interrupts()"
3) During "for irq in irqs.keys():" cycle comes new irq A that has new default, but is not saved to "irq_original"

As you restore this irq A keeps new default affinity as it's not in "irq_original".

I don't know what situation is worse in real live situations, so maybe this comment is "out of touch"

adriaan42:
@CZerta Yes, you are right, this introduces a race that can break the ability to rollback changes... I only focused on really making sure that no irq can hit one of the isolated CPUs.

I think both situations are highly unlikely to occur, so let's close and keep the current behavior.

PR #371, opened 2021-07-19
feat(scheduler): match thread names in addition to process names

adriaan42:
This is a first try at matching thread names (comm) in addition to
the process cmdline for changing scheduling policies and priorities
with the `scheduler` plugin.

It adds an optional second regex field to the end of `group.*`
options (defaults to ".*"). The first regex is to match the
process cmdline, the second one is used to match the thread's
comm. If both match, scheduling options are applied.

Note that you can have multiple `group.*` options with the
same process regex and different thread regexes. You can even have
one rule to match a specific thread name, and another with no thread
regex (matching all threads). In this case you need to assign rule
priorities (higher number -> higher priority).


tuned/plugins/plugin_scheduler.py
```diff
@@ -530,7 +531,15 @@ def _convert_sched_cfg(self, vals):
 		(scheduler, priority) = self._convert_sched_params(
 				scheduler, priority)
 		affinity = self._convert_affinity(affinity)
-		return (rule_prio, scheduler, priority, affinity, regex)
+		# split regex into two, one to match the cmdline, and one the comm
+		# if there is only one regex for the cmdline, then the comm is matched by ".*"
+		regexes = regex.split(":", 1)
```
CZerta, line 536: I would add possibility to escape colonn in regex if somebody would need ':' in process regex for some reason

CZerta (changes requested):
Except the comment, rest LGTM

tuned/plugins/plugin_scheduler.py
adriaan42: I have pushed an update that allows escaping ('\:')

CZerta (approved):
Seems OK

tuned/plugins/plugin_scheduler.py
```diff
@@ -642,29 +652,39 @@ def _instance_apply_static(self, instance):
 		sched_cfg = [(option, str(value).split(":", 4)) for option, value in instance._scheduler.items()]
 		buf = [(option, self._convert_sched_cfg(vals))
 				for option, vals in sched_cfg
-				if re.match(r"group\.", option)
+				if re.match(r"^group\.", option)
```
yarda, line 655: Why this change? IMHO the re.match should already match from the beginning of the string.

adriaan42: You're right. I'll remove the change.

yarda:
@adriaan42 please extend the ADOC documentation of the class SchedulerPlugin in the plugin_scheduler.py to describe the new feature.

PR #372, opened 2021-07-19
profiles: improve regexes to match kernel threads

adriaan42:
* add "^\[" to the start of the regexes, so they only match
    the intended kernel threads, and not processes that have
    the name of a kthread in their cmdline.
  * remove unneccessary ".*" at start/end of regexes


CZerta (approved):
LGTM

PR #373, opened 2021-07-21
bootloader: make skip_grub_config consistent with initrd_remove_dir

jmencak:
Currently, `skip_grub_config` and `initrd_remove_dir` options in the
bootloader plug-in are inconsistent.  This commit fixes this issue.

The `skip_grub_config=VALUE` does not change grub configuration if VALUE
is true.  However, cmdline options are still processed, and the result
is used to verify the current cmdline.  Only 'y', 'yes', 't', 'true' and
'1' (case insensitive) are accepted as true values for this option.
Other values are interpreted as false.

Example:
```
[bootloader]
skip_grub_config=True
cmdline=+systemd.cpu_affinity=1
```

### DIFF
--- a/tuned/plugins/plugin_bootloader.py
+++ b/tuned/plugins/plugin_bootloader.py
@@ -510,7 +510,7 @@ class BootloaderPlugin(base.Plugin):
 		if verify:
 			return None
 		if enabling and value is not None:
-			if self._cmd.get_bool(value):
+			if self._cmd.get_bool(value) == "1":
 				log.info("skipping any modification of grub config")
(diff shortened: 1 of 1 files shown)


jmencak:
@adriaan42 PTAL

adriaan42:
@jmencak my mistake. Your fix is absolutely correct!

jmencak:
Thank you for the review, Adriaan, much appreciated!
/assign @yarda

yarda:
@jmencak good catch thanks, I will merge it as a fix for the RC.

yarda (approved):

PR #375, opened 2021-07-30
Removing dependency on python-configobj.

CZerta:
Resolves: rhbz#1936386

### DIFF
--- a/recommend.conf
+++ b/recommend.conf
@@ -29,7 +29,7 @@
 # Limitation:
 # Each profile can be specified only once, because there cannot be
 # multiple sections in the configuration file with the same name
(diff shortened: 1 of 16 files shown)


yarda (commented):
Thanks for the PR. Generally it looks good. I am now running regression tests. Could you please:
- Update both GUI an non-GUI global config (tuned-main.conf) parsers to support checking of all currently supported options?
- It would be great to have unit tests for global config, profiles and variables, could you please add some?

yarda:
LGTM.

I ran several tests including tier1 and tier2 beaker tests and haven't spotted any regression.

PR #376, opened 2021-08-02
Add OpenShift profiles.

jmencak:

### DIFF
--- a/profiles/openshift-control-plane/tuned.conf
+++ b/profiles/openshift-control-plane/tuned.conf
@@ -1,0 +1,23 @@
+#
+# tuned configuration
+#
+
+[main]
+summary=Optimize systems running OpenShift control plane
+include=openshift
+
+[scheduler]
+# ktune sysctl settings, maximizing i/o throughput
+#
+# Minimal preemption granularity for CPU-bound tasks:
+# (default: 1 msec#  (1 + ilog(ncpus)), units: nanoseconds)
+sched_min_granularity_ns=10000000
+# The total time the scheduler will consider a migrated process
+# "cache hot" and thus less likely to be re-migrated
+# (system default is 500000, i.e. 0.5 ms)
+sched_migration_cost_ns=5000000
+# SCHED_OTHER wake-up granularity.
+#
+# Preemption granularity when tasks wake up.  Lower the value to
+# improve wake-up latency and throughput for latency critical tasks.
+sched_wakeup_granularity_ns=4000000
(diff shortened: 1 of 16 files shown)


jmencak:
/cc @yanirq 

TODO: Needs testing to see if the 
```
[scheduler]
cgroup_ps_blacklist=/kubepods\.slice/
```

works fine when inheriting from child profiles.

jmencak:
> TODO: Needs testing to see if the
> 
> ```
> [scheduler]
> cgroup_ps_blacklist=/kubepods\.slice/
> ```
> 
> works fine when inheriting from child profiles.

This works just fine.  Cancelling WiP
/cc @yarda

profiles/openshift-control-plane/tuned.conf
```diff
@@ -0,0 +1,23 @@
...
+sched_min_granularity_ns=10000000
+# The total time the scheduler will consider a migrated process
+# "cache hot" and thus less likely to be re-migrated
+# (system default is 500000, i.e. 0.5 ms)
+sched_migration_cost_ns=5000000
+# SCHED_OTHER wake-up granularity.
+#
+# Preemption granularity when tasks wake up.  Lower the value to
+# improve wake-up latency and throughput for latency critical tasks.
+sched_wakeup_granularity_ns=4000000
```
yarda, line 23: FYI we have recently dropped the sched_ tuning from nearly all upstream profiles, because after benchmarking it turned out the performance gain is negligible and it's better to go with the kernel defaults. So in case this is based on some stock TuneD profile, you may consider updating and maybe dropping this profile? There is no problem with the sched_ tuning, it is fully supported, just FYI.

yarda (approved):
LGTM. I only added comment about the sched_ tuning. If it is intended or worked for you in your benchmarks, I am OK to merge it.

profiles/openshift-control-plane/tuned.conf
jmencak: Right. So this is old tuning that goes back to RHEL 7.x where it was helping during HAProxy tests if I remember correctly.  So it is quite possible, that in RHEL 9.x the situation is quite different.  Let's keep this as is for the time being and I'll create a task to re-evaluate these settings (RHEL 8.x and RHEL 9.x) and adjust them later if the performance impact is negligible or better with the kernel defaults.  I assume you're talking about the RHEL 9.x kernels here?

yarda: Yes, RHEL-9, it's also considered for backporting into RHEL-8. OK, I will merge it as it is.

PR #380, opened 2021-09-22
Default delimiters errors, inline comments

### LINKED ISSUE #378: ConfigParser default delimiter causing errors
In my powertop2tuned generated tuned.conf I have several devices with a colon in the path, for example:

> /sys/bus/pci/devices/0000:01:00.0

As far as I know ConfigParser does not allow the delimiters to be escaped, which causes this error:

> Cannot load profile(s) 'powersave': ("Cannot parse '/etc/tuned/powersave/tuned.conf'.", While reading from '/etc/tuned/powersave/tuned.conf' [line 42]: option '/sys/bus/pci/devices/0000' in section 'sysfs' already exists)

A simple fix for my particular issue would be to set the delimiter to only "=" whenever a ConfigParser is created. For example:

`config_parser = ConfigParser(delimiters=('='))`

CZerta:
After transit to build in configparser in specific situations it can cause problems. This PR should resolve it.

Resolves #378

### DIFF
--- a/tuned/profiles/locator.py
+++ b/tuned/profiles/locator.py
@@ -1,12 +1,6 @@
 import os
 import tuned.consts as consts
-try:
-	from configparser import ConfigParser, Error
-	from io import StringIO
-except ImportError:
-	# python2.7 support, remove RHEL-7 support end
-	from ConfigParser import ConfigParser, Error
-	from StringIO import StringIO
+from tuned.utils.config_parser import ConfigParser, Error
 
 class Locator(object):
 	"""
@@ -54,10 +48,10 @@ class Locator(object):
 		if config_file is None:
 			return None
 		try:
-			config = ConfigParser()
+			config = ConfigParser(delimiters=('='), inline_comment_prefixes=('#'), allow_no_value=True)
 			config.optionxform = str
 			with open(config_file) as f:
-				config.readfp(StringIO("[" + consts.MAGIC_HEADER_NAME + "]\n" + f.read()))
+				config.read_string("[" + consts.MAGIC_HEADER_NAME + "]\n" + f.read())
 			return config
 		except (IOError, OSError, Error) as e:
 			return None
--- a/tuned/utils/config_parser.py
+++ b/tuned/utils/config_parser.py
@@ -1,0 +1,56 @@
+# ConfigParser wrapper providing compatibility layer for python 2.7/3
+
+try:
+	python3 = True
+	import configparser as cp
+except ImportError:
+	python3 = False
+	import ConfigParser as cp
+	from StringIO import StringIO
+	import re
+
+class Error(cp.Error):
+	pass
+
+if python3:
+
+	class ConfigParser(cp.ConfigParser):
+		pass
+
+else:
+
+	class ConfigParser(cp.ConfigParser):
+
+		def __init__(self, delimiters=None, inline_comment_prefixes=None, strict=True, *args, **kwargs):
+			delims = "".join(list(delimiters))
+			# REs taken from the python-2.7 ConfigParser
+			self.OPTCRE = re.compile(
+				r'(?P<option>[^' + delims + '\s][^' + delims + ']*)'
+				r'\s*(?P<vi>[' + delims + '])\s*'
+				r'(?P<value>.*)$'
+			)
+			self.OPTCRE_NV = re.compile(
+				r'(?P<option>[^' + delims + '\s][^' + delims + ']*)'
+				r'\s*(?:'
+				r'(?P<vi>[' + delims + '])\s*'
+				r'(?P<value>.*))?$'
+			)
+			cp.ConfigParser.__init__(self, *args, **kwargs)
+			self._inline_comment_prefixes = inline_comment_prefixes or []
+			self._re = re.compile("\s+(%s).*" % ")|(".join(list(self._inline_comment_prefixes)))
+
+		def read_string(self, string, source="<string>"):
+			sfile = StringIO(string)
+			self.read_file(sfile, source)
+
+		def readfp(self, fp, filename=None):
+			cp.ConfigParser.readfp(self, fp, filename)
+			# remove inline comments
+			all_sections = [self._defaults]
+			all_sections.extend(self._sections.values())
+			for options in all_sections:
+				for name, val in options.items():
+					options[name] = self._re.sub("", val)
+
+		def read_file(self, f, source="<???>"):
+			self.readfp(f, source)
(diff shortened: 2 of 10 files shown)


tuned/profiles/locator.py
```diff
@@ -54,10 +52,10 @@ def parse_config(self, profile_name):
 		if config_file is None:
 			return None
 		try:
-			config = ConfigParser()
+			config = ConfigParser(delimiters=('='), inline_comment_prefixes=('#'), allow_no_value=True)
```
jmencak, line 55: Why do we have in some `ConfigParser()` instantiations `allow_no_value=True` and not in others?  Could you share an example where `allow_no_value=True` is neccessary?

jmencak (commented):
I went through the code and the changes make sense to me (didn't look at Python 2.7 specific code in too much detail).  I also tested this fixes [rhbz#2060138](https://bugzilla.redhat.com/show_bug.cgi?id=2060138) and https://github.com/redhat-performance/tuned/issues/378 I have only 1 question.

tuned/profiles/locator.py
CZerta: Here are looser rules, so we can locate even wrongly written profiles, otherwise there would be no possibility to show it to user in list of profiles and he would need to know that if he creates profile with mistake, it won't show anywhere and he needs to manually locate it in directory.
In variables.py there is no other possibility to use None vars. I don't know if there is any real world situation when somebody would want that, but it was possible before and we didn't really check syntax of these files, so I left it compatible.

tuned/utils/config_parser.py
```diff
@@ -0,0 +1,121 @@
...
+# After support for p 2.7 is dropped, so can be this file
+
+try:
+	import ConfigParser as cp
+	from StringIO import StringIO
+except ImportError:
+	raise ImportError("This file should not be used in python 3")
+import re
+
+SECTION_REGEX = r'''^
```
yarda, line 11: I think we could simplify things here and for the 2.7 just emulate the inline comments after parsing the file with the re.sub() on the parsed values. I think it won't be problem if the parser doesn't work 100% as with the python 3.

yarda (commented):
Mostly LGTM, I will try to update the PR to reflect my ideas. It will require re-review :)

yarda:
Please review the simplified version. It may not be 1:1 python 2.7 drop-in replacement, but the main functionality should be there.

tuned/utils/config_parser.py
```diff
@@ -0,0 +1,43 @@
...
+		pass
+
+else:
+
+	class ConfigParser(cp.ConfigParser):
+
+		def __init__(self, delimiters=None, inline_comment_prefixes=None, strict=True, *args, **kwargs):
+			cp.ConfigParser.__init__(self, *args, **kwargs)
+			self._inline_comment_prefixes = inline_comment_prefixes or []
+			self._re = re.compile("\s+(%s).*" % ")|(".join(list(self._inline_comment_prefixes)))
```
jmencak, line 27: This means inline comments have to start with whitespace character.  To my mind this is not robust and this shouldn't be needed, i.e. they should start with the inline comment character itself.  Is this intentional?

jmencak (commented):
I've tested this PR and it mostly works.

This is the case where it fails:
```
[section]
key=value# this is an inline comment
```

tuned/utils/config_parser.py
yarda: Yes, it is intentional. From the python 2.7 ConfigParser source code:
>  ';' is a comment delimiter only if it follows a spacing character

I.e. in the python 2.7 the ConfigParsr takes the ';' as the comment (it is usually used this way in the INI file). Maybe we should also use the `inline_comment_prefixes = ['#', ';']` for better compatibility?

Also with the python 3 and `inline_comment_prefixes = ['#']` it seems there has to be a white space before the comment, so for me the behavior seems consistent.

yarda: > Yes, it is intentional. From the python 2.7 ConfigParser source code:
> 
> > ';' is a comment delimiter only if it follows a spacing character
> 
> I.e. in the python 2.7 the ConfigParser takes the ';' as the comment (it is usually used this way in the INI file). Maybe we should also use the `inline_comment_prefixes = ['#', ';']` for better compatibility?
> 
> Also with the python 3 and `inline_comment_prefixes = ['#']` it seems there has to be a white space before the comment, so for me the behavior seems consistent.

This holds only for the so called 'inline comments', the normal comments starts the line and do not need to be prefixed by the white space.

tuned/utils/config_parser.py
```diff
@@ -0,0 +1,43 @@
...
+if python3:
+
+	class ConfigParser(cp.ConfigParser):
+		pass
+
+else:
+
+	class ConfigParser(cp.ConfigParser):
+
+		def __init__(self, delimiters=None, inline_comment_prefixes=None, strict=True, *args, **kwargs):
```
CZerta, line 24: Am I blind or you don't use `delimiters` anywhere? This way this does not resolve #378

CZerta: @yarda You don't use `delimiters` and `strict` anywhere so it even get around unit tests.

yarda: Regarding delimiters it's a good catch, the 'strict' is currently ignored on the 2.7. If you think it's needed I could add duplicity check in the post parse.

yarda:
Added ';' to inline comments for backward compatibility.

yarda:
The idea behind the patch is to reuse the most possible things from the python 2.7 ConfigParser (where the code is frozen and will not change) and not rewrite it. Because e.g. the python 3 configparser REs could change with the python updates and it wouldn't be easy to keep it in sync.

CZerta:
> Added ';' to inline comments for backward compatibility.

I don't know if it's good idea, configobj that this replaces does not allow it and users can use ';' as delimiter in some of our "list" options, so I wouldn't gamble that this will break something (`_parse_service_options` in `plugin_service`, `_set_radeon_powersave` in `plugin_video` etc)

yarda:
> > Added ';' to inline comments for backward compatibility.
> 
> I don't know if it's good idea, configobj that this replaces does not allow it and users can use ';' as delimiter in some of our "list" options, so I wouldn't gamble that this will break something (`_parse_service_options` in `plugin_service`, `_set_radeon_powersave` in `plugin_video` etc)

OK, I will drop it.

yarda:
Dropped the ';' for comments.

CZerta (commented):
Seems OK, as original author I cannot approve

jmencak:
Thank you for the changes, Honza/Jarda.
LGTM
/lgtm

jmencak (approved):
/approve

PR #384, opened 2021-10-04
Tests: Be in sync with upstreaming guide.

RHEmployee:


tests/beakerlib/Program-tuned-tried-to-access-dev-mem-between/runtest.sh
```diff
@@ -8,20 +8,9 @@
 #
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 #
-#   Copyright (c) 2019 Red Hat, Inc.
+#   Copyright Red Hat
```
yarda, line 11: Additional trailing whitespace. The same problem is in other files in this PR as well.

Could you point me (e.g. in PM) to the doc requiring this change? The doc I was able to find states:
"May contain a Red Hat copyright notice (ordinarily use the form: 'Copyright <YEAR> Red Hat, Inc. and/or its affiliates' where <YEAR> is the year the code is being first released (the year need not be updated)."

I.e. there is probably no need to drop the year and the 'Inc.' should be probably also there.

yarda (changes requested):

tests/beakerlib/Program-tuned-tried-to-access-dev-mem-between/runtest.sh
RHEmployee: Hi Jarda. Thank you for heads up.
This documentation is private so I can share url only in private.

yarda (approved):

yarda:
The uncompleted test result (13/14) seems to be left over from the previous runs and seems unrelated to the PR.

PR #386, opened 2021-10-13
openshift profile: workaround high CPU utilization of [scheduler] plug-in

jmencak:
The `[scheduler]` plug-in can be very CPU intensive, especially on the OpenShift platform.  The bug for this issue is tracked by rhbz#1921738. Until this is fixed, work around this problem in the openshift parent profile by adding `runtime=0` `[scheduler]` plug-in option.


yarda (approved):
Temporal workaround seems OK to me.

PR #390, opened 2021-11-19
Remove rcu_nocbs from cpu-partitioning profile

amaumene:
This seem to be redundant, at least according to https://github.com/torvalds/linux/blob/4c388a8e740d3235a194f330c8ef327deef710f6/Documentation/admin-guide/kernel-parameters.txt#L3484

### DIFF
--- a/profiles/cpu-partitioning/tuned.conf
+++ b/profiles/cpu-partitioning/tuned.conf
@@ -66,4 +66,4 @@ priority=10
 initrd_remove_dir=True
 initrd_dst_img=tuned-initrd.img
 initrd_add_dir=${tmpdir}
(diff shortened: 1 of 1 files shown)


amaumene:
fyi rcu_nocbs is actually creating an issue with RHEL8.4 https://bugzilla.redhat.com/show_bug.cgi?id=2013408

Xyene:
Reference to kernel source showing that `nohz_full` implies `rcu_nocbs`: https://github.com/torvalds/linux/blob/6f38be8f2ccd9babf04b9b23539108542a59fcb8/kernel/rcu/tree_nocb.h#L1191

PR #391, opened 2021-12-06
spec: do not require subscription-manager on CentOS

yarda:
Resolves: rhbz#2028865


CZerta (approved):
Seems OK

yarda:
The f33 error was caused by CI failure and the failure of the eol f33 cannot be cleared now, thus it's OK.

PR #392, opened 2021-12-16
cpu_partitioning: fixed no_balance_cores on newer kernels

yarda:
…LOAD_BALANCE

Resolves: rhbz#1874596

### DIFF
--- a/profiles/cpu-partitioning/tuned.conf
+++ b/profiles/cpu-partitioning/tuned.conf
@@ -35,6 +35,8 @@ no_balance_cores_expanded=${f:cpulist_unpack:${no_balance_cores}}
 # Fail if isolated_cores contains CPUs which are not online
 assert2=${f:assertion:isolated_cores contains online CPU(s):${isolated_cores_expanded}:${isolated_cores_online_expanded}}
 
+cmd_isolcpus=${f:regex_search_ternary:${no_balance_cores}:\s*[0-9]: isolcpus=${no_balance_cores}:}
+
 [sysctl]
 kernel.hung_task_timeout_secs = 600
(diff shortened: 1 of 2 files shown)


jmencak (commented):
This PR looks good to me and I believe this fixes the issues for RHEL 9.0.  What is not quite clear to me is whether we can use this on RHEL 8.x systems as well.

profiles/cpu-partitioning/tuned.conf
```diff
@@ -35,6 +35,8 @@ no_balance_cores_expanded=${f:cpulist_unpack:${no_balance_cores}}
 # Fail if isolated_cores contains CPUs which are not online
 assert2=${f:assertion:isolated_cores contains online CPU(s):${isolated_cores_expanded}:${isolated_cores_online_expanded}}
 
+cmd_isolcpus=${f:regex_search_ternary:${no_balance_cores}:\s*[0-9]+: isolcpus=${no_balance_cores}:}
```
jmencak, line 38: We can drop the `+` in the regex.

yarda: OK, I will drop it.

yarda:
> This PR looks good to me and I believe this fixes the issues for RHEL 9.0. What is not quite clear to me is whether we can use this on RHEL 8.x systems as well.

Thanks for the review.

According to the discussion in the Red Hat bugzilla this should work OK on RHEL-8, but in the RHEL-8 downstream this PR will be reverted not to cause any regression or different behaviour. On newer RHELs this upstream version of the profile will be used.

CZerta (approved):
Seems OK

PR #395, opened 2022-01-03
Adding support for nvme subsystem block devices

CZerta:
Adding support for nvme subsystem block devices and adding tests so it won't throw errors on hdparms commands.

Resolves: rhbz#1854816

### DIFF
--- a/tuned/plugins/plugin_disk.py
+++ b/tuned/plugins/plugin_disk.py
@@ -27,22 +27,42 @@ class DiskPlugin(hotplug.Plugin):
 	def _init_devices(self):
 		super(DiskPlugin, self)._init_devices()
 		self._devices_supported = True
+		self._use_hdparm = True
 		self._free_devices = set()
+		self._hdparm_apm_devices = set()
 		for device in self._hardware_inventory.get_devices("block"):
 			if self._device_is_supported(device):
 				self._free_devices.add(device.sys_name)
+				if self._use_hdparm and self._is_hdparm_apm_supported(device.sys_name):
+					self._hdparm_apm_devices.add(device.sys_name)
 
 		self._assigned_devices = set()
 
 	def _get_device_objects(self, devices):
 		return [self._hardware_inventory.get_device("block", x) for x in devices]
 
+	def _is_hdparm_apm_supported(self, device):
+		(rc, out, err_msg) = self._cmd.execute(["hdparm", "-C", "/dev/%s" % device], \
+				no_errors = [errno.ENOENT], return_err=True)
+		if rc == -errno.ENOENT:
+			log.warn("hdparm command not found, ignoring for other devices")
+			self._use_hdparm = False
+			return False
+		elif rc:
+			log.info("Device '%s' not supported by hdparm" % device)
+			log.debug("(rc: %s, msg: '%s')" % (rc, err_msg))
+			return False
+		elif "unknown" in out:
+			log.info("Driver for device '%s' does not support apm command" % device)
+			return False
+		return True
+
 	@classmethod
 	def _device_is_supported(cls, device):
 		return  device.device_type == "disk" and \
 			device.attributes.get("removable", None) == b"0" and \
 			(device.parent is None or \
-					device.parent.subsystem in ["scsi", "virtio", "xen"])
+					device.parent.subsystem in ["scsi", "virtio", "xen", "nvme"])
 
 	def _hardware_events_init(self):
 		self._hardware_inventory.subscribe(self, "block", self._hardware_events_callback)
@@ -142,6 +162,8 @@ class DiskPlugin(hotplug.Plugin):
 		return not "standby" in out and not "sleeping" in out
 
 	def _instance_update_dynamic(self, instance, device):
+		if not device in self._hdparm_apm_devices:
+			return
 		load = instance._load_monitor.get_device_load(device)
 		if load is None:
 			return
@@ -202,7 +224,7 @@ class DiskPlugin(hotplug.Plugin):
 		diff = [new_old[0] - new_old[1] for new_old in zip(new_load, old_load)]
 		instance._stats[device]["diff"] = diff
 
-		# adapt maximum expected load if the difference is higer
+		# adapt maximum expected load if the difference is higher
 		old_max_load = instance._stats[device]["max"]
 		max_load = [max(pair) for pair in zip(old_max_load, diff)]
 		instance._stats[device]["max"] = max_load
@@ -218,6 +240,15 @@ class DiskPlugin(hotplug.Plugin):
 				instance._idle[device][operation] += 1
 			else:
 				instance._idle[device][operation] = 0
+
+	def _instance_apply_dynamic(self, instance, device):
+		# At the moment we support dynamic tuning just for devices compatible with hdparm apm commands
+		# If in future will be added new functionality not connected to this command,
+		# it is needed to change it here
+		if device not in self._hdparm_apm_devices:
+			log.info("There is no dynamic tuning available for device '%s' at time" % device)
+		else:
+			super(DiskPlugin, self)._instance_apply_dynamic(*args, **kwargs)
 
 	def _instance_unapply_dynamic(self, instance, device):
 		pass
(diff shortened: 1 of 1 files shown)


jmencak:
Thank you for the PR, @CZerta !

Unfortunately, with the following config file:
```
[main]
summary=Custom profile

[disk]
elevator=bfq
```

I'm getting:
```
# tuned          
2022-01-05 15:34:22,875 INFO     tuned.daemon.application: dynamic tuning is enabled (can be overridden in plugins)
2022-01-05 15:34:22,880 INFO     tuned.daemon.daemon: using sleep interval of 1 second(s)
2022-01-05 15:34:22,880 INFO     tuned.daemon.daemon: dynamic tuning is enabled (can be overridden by plugins)
2022-01-05 15:34:22,880 INFO     tuned.daemon.daemon: using update interval of 10 second(s) (10 times of the sleep interval)
2022-01-05 15:34:22,880 INFO     tuned.daemon.daemon: Running in automatic mode, checking what profile is recommended for your configuration.
2022-01-05 15:34:22,880 INFO     tuned.daemon.daemon: Using 'scheduler' profile
2022-01-05 15:34:22,881 INFO     tuned.profiles.loader: loading profile: scheduler
2022-01-05 15:34:22,883 INFO     tuned.daemon.controller: starting controller
2022-01-05 15:34:22,883 INFO     tuned.daemon.daemon: starting tuning
Exception in thread Thread-1:
  File "/usr/lib64/python3.9/threading.py", line 973, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.9/threading.py", line 910, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3.9/site-packages/tuned/daemon/daemon.py", line 193, in _thread_code
    self._unit_manager.create(self._profile.units)
  File "/usr/lib/python3.9/site-packages/tuned/units/manager.py", line 109, in create
    instance.plugin.init_devices()
  File "/usr/lib/python3.9/site-packages/tuned/plugins/base.py", line 49, in init_devices
    self._init_devices()
  File "/usr/lib/python3.9/site-packages/tuned/plugins/plugin_disk.py", line 33, in _init_devices
    if self._device_is_supported(device):
  File "/usr/lib/python3.9/site-packages/tuned/plugins/plugin_disk.py", line 49, in _device_is_supported
    (cls._is_hdparm_supported(device) or device.parent.subsystem == "nvme")
TypeError: _is_hdparm_supported() missing 1 required positional argument: 'device'
^C2022-01-05 15:34:26,626 INFO     tuned.daemon.controller: terminating controller
```

Nits: Looking at the commit message, I believe we should use `Add` vs. `Adding` and the commit title should be much shorter (50/72 rule).

tuned/plugins/plugin_disk.py
```diff
@@ -36,13 +37,16 @@ def _init_devices(self):
 
 	def _get_device_objects(self, devices):
 		return [self._hardware_inventory.get_device("block", x) for x in devices]
+
```
jmencak, line 40: Nit: extra whitespace.

tuned/plugins/plugin_disk.py
```diff
@@ -36,13 +37,16 @@ def _init_devices(self):
 
 	def _get_device_objects(self, devices):
 		return [self._hardware_inventory.get_device("block", x) for x in devices]
+	
+	def _is_hdparm_supported(self, device):
+		# Do we want is None here? It returns loop devices that are not supported
+		return device.parent is None or device.parent.subsystem in self._hdparm_subsystems
```
jmencak, line 43: I wonder if this is really the way to go to establish `hdparm` support or whether we could use `hdparm`'s exit code to establish support? Disclaimer: haven't looked into that closely.

jmencak (commented):
Only a few comments as the PR unfortunately didn't work for me.

tuned/plugins/plugin_disk.py
CZerta: `hdparm` does always return 0 except of invalid inputt (i.e. [https://bugzilla.redhat.com/show_bug.cgi?id=28890](url)). I left this test because of backward compatibility, maybe there would be better to run something like `hdparm -I`and parse output.

jmencak: Interesting,
```
# hdparm -Y /dev/nvme0n1

/dev/nvme0n1:
 issuing sleep command
 HDIO_DRIVE_CMD(sleep) failed: Inappropriate ioctl for device
# echo $?
25
```
seems to work.  I'll look through the code and test this a bit more.

CZerta: Yeah, but it would be nice not to shutdown drivers that support hdparm. I'm looking into `hdparm -C,` if there's possibility to return `unknown` even if the driver supports other commands (probably best option), or `hdparm -i`, but don't want to rely on old driver info.

tuned/plugins/plugin_disk.py
```diff
@@ -361,6 +388,6 @@ def _get_scheduler_quantum(self, device, ignore_missing=False):
 		value = self._cmd.read_file(sys_file, no_error=ignore_missing).strip()
 		if len(value) == 0:
 			if not ignore_missing:
-				log.info("disk_scheduler_quantum option is not supported by this HW")
+				log.warn("disk_scheduler_quantum option is not supported by this HW")
```
yarda, line 391: I think the info/warn should be consistent. E.g. in the `@command_get("spindown")` there is info level, but here in the `@command_get("scheduler_quantum")` is the warn level, why? Also are you using the info level for the 'get' and the warn level for the 'set'? Is there any reason for it? Or maybe the set/get levels should be also consistent?

yarda (changes requested):
The CI tests are failing with the traceback @jmencak pointed out, @CZerta could you fix it please?

tuned/plugins/plugin_disk.py
CZerta: We were talking about "higher" level of log before, so changed it to warn, info in `@command_get("spindown")` was incaution, biut I changed it all to info now.

tuned/plugins/plugin_disk.py
```diff
@@ -142,6 +158,9 @@ def _drive_spinning(self, device):
 		return not "standby" in out and not "sleeping" in out
 
 	def _instance_update_dynamic(self, instance, device):
+		if not device in self._hdparm_devices:
```
yarda, line 161: Can we log it just once in the `_instance_apply_dynamic` and maybe log it at the level info.

tuned/plugins/plugin_disk.py
```diff
@@ -248,6 +267,12 @@ def _get_elevator(self, device, ignore_missing=False):
 
 	@command_set("apm", per_device=True)
 	def _set_apm(self, value, device, sim):
+		if device not in self._hdparm_devices:
+			if not sim:
+				log.info("apm option is not supported by this HW")
```
yarda, line 272: I think you should log the device name and similarly in other logs.

tuned/plugins/plugin_disk.py
```diff
@@ -27,22 +27,38 @@ def __init__(self, *args, **kwargs):
...
 				self._free_devices.add(device.sys_name)
+				if self._use_hdparm and self._is_hdparm_supported(device.sys_name):
+					self._hdparm_devices.add(device.sys_name)
 
 		self._assigned_devices = set()
 
 	def _get_device_objects(self, devices):
 		return [self._hardware_inventory.get_device("block", x) for x in devices]
 
+	def _is_hdparm_supported(self, device):
```
yarda, line 44: This is checking whether hdparm is there and the drive supports APM/standby. Maybe rename the function to the _is_apm_supported or _is_hdparm_apm_supported or similar name?

tuned/plugins/plugin_disk.py
```diff
@@ -27,22 +27,38 @@ def __init__(self, *args, **kwargs):
...
+		(rc, out, err_msg) = self._cmd.execute(["hdparm", "-C", "/dev/%s" % device], \
+				no_errors = [errno.ENOENT], return_err=True)
+		if rc == -errno.ENOENT:
+			log.warn("hdparm command not found, ignoring for other devices")
+			self._use_hdparm = False
+			return False
+		elif rc:
+			log.debug("device %s not supported by hdparm (rc: %s, msg: '%s')" % (device, rc, err_msg))
+			return False
+		return "unknown" not in out
```
yarda, line 54: I think you should also log debug it, e.g. hdparm returned uknown for the drive, or similarly.

yarda (changes requested):

tuned/plugins/plugin_disk.py
yarda: I meant the '"unknown" in out' condition.

yarda (approved):

yarda:
Thanks.

PR #397, opened 2022-01-10
configparser: switch to the read_file(), python-3.11 fix

yarda:
ConfigParser.readfp() is deprecated since python-3.2 and was finally dropped
in the python-3.11 which caused breakage of the TuneD.

This commit switches from the readfp() to the read_file() while still
keeping backward compatibility with the python-3.2 and lower (e.g. python-2.7).

It utilizes wrapper which tries to call the read_file() first and if it
fails it fallbacks to the readfp().

Resolves: rhbz#2038886

### DIFF
--- a/tuned/gtk/gui_plugin_loader.py
+++ b/tuned/gtk/gui_plugin_loader.py
@@ -28,12 +28,13 @@ import importlib
 
 import tuned.consts as consts
 import tuned.logs
+from tuned.utils.config_parser import TuneDConfigParser
 try:
-    from configparser import ConfigParser, Error
+    from configparser import Error
     from io import StringIO
 except ImportError:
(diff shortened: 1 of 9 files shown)


yarda:
Another and maybe cleaner way how to fix it is to define own configparser class inherited from the ConfigParser with the wrapper method.

yarda:
> Another and maybe cleaner way how to fix it is to define own configparser class inherited from the ConfigParser with the wrapper method.

Right, I will probably rewrite it, it isn't good to introduce new hacks :)

yarda:
Rewritten to introduce new warpper class.

CZerta:
There is already PR #380, that fixes this and also this implementation does not take into consideration issue #378

yarda:
Handled  by #380.

PR #398, opened 2022-01-10
dbus: fix traceback on python-2.7

yarda:
The code which replaced python-decorator introduced python-2.7
incompatibility. The getfullargspec() from the inspect module is drop in
replacement for the getargspec() and both return the named tuple. Thus
the args member has to be extracted the same way in both cases
(python 2/3).


CZerta (approved):
Seems OK

PR #399, opened 2022-01-10
application: log TuneD and kernel version

yarda:
This may help with diagnosis of the problems, especially from
downstreams different than Fedora / RHEL.


CZerta (changes requested):
When we're checking 2.7 compatibility, `os.uname()` returns tuple in 2.7, so this fails

yarda:
> When we're checking 2.7 compatibility, `os.uname()` returns tuple in 2.7, so this fails

Good catch, hopefully fixed now.

Maybe we could add the python-2.7 target to the CI.

CZerta (approved):
Seems OK

PR #401, opened 2022-01-11
Fix improper parsing of include directive

jmencak:
Resolves rhbz#2017924

### DIFF
--- a/tuned/profiles/loader.py
+++ b/tuned/profiles/loader.py
@@ -90,7 +90,7 @@ class Loader(object):
 			config = self._load_config_data(filename)
 			profile = self._profile_factory.create(name, config)
 			if "include" in profile.options:
-				include_names = re.split(r"\b\s*[,;]\s*", self._variables.expand(profile.options.pop("include")))
+				include_names = re.split(r"\s*[,;]\s*", self._variables.expand(profile.options.pop("include")))
 				self._load_profile(include_names, profiles, processed_files)
 
 			profiles.append(profile)


yarda (changes requested):
This adds the space as the allowed delimiter which is change of the behavior, i.e.:
`profile1 profile2` is now parsed as two profiles: profile1, profile2 and not as one profile "profile1 profile2".

jmencak:
> This adds the space as the allowed delimiter which is change of the behavior, i.e.: `profile1 profile2` is now parsed as two profiles: profile1, profile2 and not as one profile "profile1 profile2".

Thank you for the review, @yarda!  Hopefully, we'll never see profile names with whitespace in them unless they're in built-ins.   In any case, fixed, but still not bullet-proof.  A much better fix would be a proper parser to take things such as built-ins into account.  We can hopefully leave this as future work...

yarda:
> Thank you for the review, @yarda! Hopefully, we'll never see profile names with whitespace in them unless they're in built-ins. In any case, fixed, but still not bullet-proof. A much better fix would be a proper parser to take things such as built-ins into account. We can hopefully leave this as future work...

Wouldn't just dropping the '\b' from the original code do the same?

jmencak:
> Wouldn't just dropping the '\b' from the original code do the same?

Actually, it would.  Thanks for pointing it out.  I wasn't aware leading/trailing spaces were already being stripped.

yarda (approved):

PR #402, opened 2022-01-16
readme: mention conflicting cpupower and power-profiles-daemon

yarda:
Also updated instructions to use dnf instead of the yum.

Related: #396


CZerta (approved):
Seems OK

PR #404, opened 2022-02-01
Expanding cpu plugin force_latency option syntax

CZerta:
Adding "cstate.id_no_zero:X" and "cstate.name_no_zero:X" syntax to force_latency option that skips given cstate if its latency is 0.

Resolves: rhbz#2034069
Resolves: rhbz#2002744


yarda (approved):
Thanks, LGTM. We also need to fix the profiles. I will do PR for it.

PR #406, opened 2022-02-01
profiles: use cstate.id_no_zero instead of the cstate.id

yarda:
Related: rhbz#2034069
Related: rhbz#2002744


CZerta (approved):
Seems OK

PR #407, opened 2022-02-01
daemon: don't do full rollback on systemd failure

yarda:
Resolves: rhbz#2011459


CZerta (approved):
Seems OK

PR #408, opened 2022-02-01
bootloader: on s390(x) remove TuneD variables from the BLS

yarda:
Resolves: rhbz#1978786


CZerta (approved):
Seems OK

PR #415, opened 2022-03-07
Do not hardcode disk name in tuned initrd path where not needed

CZerta:
On systems where /boot dir is not on separate partition grub cannot resolve disk name and cause kernel panic. This is not needed otherwise and can cause kernel panic when disk is renamed.

Resolves: rhbz#2050246

tuned_kernel_panic

### DIFF
--- a/tuned/plugins/plugin_bootloader.py
+++ b/tuned/plugins/plugin_bootloader.py
@@ -395,7 +395,7 @@ class BootloaderPlugin(base.Plugin):
 		initrd_grubpath = "/"
 		lc = len(curr_cmdline)
 		if lc:
-			path = re.sub(r"^\s*BOOT_IMAGE=\s*(\S*/).*$", "\\1", curr_cmdline)
+			path = re.sub(r"^\s*BOOT_IMAGE=\s*(?:\([^)]*\))?(\S*/).*$", "\\1", curr_cmdline)
 			if len(path) < lc:
 				initrd_grubpath = path
 		self._initrd_val = os.path.join(initrd_grubpath, img_name)


yarda (commented):
In the case it's on the same partition shouldn't it be:
```
/boot/....
```
instead of the e.g.:
```
(hd1,gpt2)/boot/...
```
I.e. could be the ismount() check removed and the regex updated to skip the device but include the additional directories?

I didn't check whether it works, just asking.

CZerta:
No. It's the reason why it was added in [commit](https://github.com/redhat-performance/tuned/commit/8442dabf544ca3f27635ef9f2331d36a6965e3a8), if the /boot is not on separate partition it cannot locate it in the first place.

yarda:
@CZerta what about the following patch?
```
--- a/tuned/plugins/plugin_bootloader.py
+++ b/tuned/plugins/plugin_bootloader.py
@@ -395,7 +395,7 @@ class BootloaderPlugin(base.Plugin):
                initrd_grubpath = "/"
                lc = len(curr_cmdline)
                if lc:
-                       path = re.sub(r"^\s*BOOT_IMAGE=\s*(\S*/).*$", "\\1", curr_cmdline)
+                       path = re.sub(r"^\s*BOOT_IMAGE=\s*(?:\([^)]*\))?(\S*/).*$", "\\1", curr_cmdline)
                        if len(path) < lc:
                                initrd_grubpath = path
                self._initrd_val = os.path.join(initrd_grubpath, img_name)
```
It worked for me.

yarda:
> @CZerta what about the following patch?
> 
> ```
> --- a/tuned/plugins/plugin_bootloader.py
> +++ b/tuned/plugins/plugin_bootloader.py
> @@ -395,7 +395,7 @@ class BootloaderPlugin(base.Plugin):
>                 initrd_grubpath = "/"
>                 lc = len(curr_cmdline)
>                 if lc:
> -                       path = re.sub(r"^\s*BOOT_IMAGE=\s*(\S*/).*$", "\\1", curr_cmdline)
> +                       path = re.sub(r"^\s*BOOT_IMAGE=\s*(?:\([^)]*\))?(\S*/).*$", "\\1", curr_cmdline)
>                         if len(path) < lc:
>                                 initrd_grubpath = path
>                 self._initrd_val = os.path.join(initrd_grubpath, img_name)
> ```
> 
> It worked for me.

I did few tests, it worked correctly for me on:
- PC BIOS, /boot on the separate partition
- PC BIOS, /boot on the root partition
- GPT/UEFI, /boot on the separate partition
- GPT/UEFI, /boot on the root partition

@CZerta if you find the new patch working for you, could you please update the PR?

NeilHanlon:
fwiw, I think my commit in https://github.com/redhat-performance/tuned/pull/393 is 'fixing' the same issue here as well, albeit in a much more bruteforce way.

CZerta:
> > @CZerta what about the following patch?
> > ```
> > --- a/tuned/plugins/plugin_bootloader.py
> > +++ b/tuned/plugins/plugin_bootloader.py
> > @@ -395,7 +395,7 @@ class BootloaderPlugin(base.Plugin):
> >                 initrd_grubpath = "/"
> >                 lc = len(curr_cmdline)
> >                 if lc:
> > -                       path = re.sub(r"^\s*BOOT_IMAGE=\s*(\S*/).*$", "\\1", curr_cmdline)
> > +                       path = re.sub(r"^\s*BOOT_IMAGE=\s*(?:\([^)]*\))?(\S*/).*$", "\\1", curr_cmdline)
> >                         if len(path) < lc:
> >                                 initrd_grubpath = path
> >                 self._initrd_val = os.path.join(initrd_grubpath, img_name)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It worked for me.
> 
> I did few tests, it worked correctly for me on:
> 
>     * PC BIOS, /boot on the separate partition
> 
>     * PC BIOS, /boot on the root partition
> 
>     * GPT/UEFI, /boot on the separate partition
> 
>     * GPT/UEFI, /boot on the root partition
> 
> 
> @CZerta if you find the new patch working for you, could you please update the PR?

I also tried removing disks, but there is probably no reason, why it should not work.

PR #418, opened 2022-03-14
Adding debug log for tuned-adm recommend command.

CZerta:
Resolves: rhbz#1645119

### DIFF
--- a/tuned/utils/profile_recommender.py
+++ b/tuned/utils/profile_recommender.py
@@ -3,6 +3,7 @@ import re
 import errno
 import procfs
 import subprocess
+import logging
 try:
 	from configparser import ConfigParser, Error
 except ImportError:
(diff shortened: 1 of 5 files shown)


tuned/utils/profile_recommender.py
```diff
@@ -123,6 +143,7 @@ def process_config(self, fname, has_root=True):
 					# remove the ",.*" suffix
 					r = re.compile(r",[^,]*$")
 					matching_profile = r.sub("", section)
+					log.debug("All options matched for profile '%s' from setion '%s', applying" % (matching_profile, section))
```
yarda, line 146: Typo 'setion'.

yarda (changes requested):
Please fix the typo, otherwise LGTM.

yarda:
After playing with it a bit I think `tuned-adm -d recommend` should write the debug on the console.

There are two scenarios:
1) Daemon not running - in such case it should call profile_recommender and set it to use the tuned-adm logger
2) Daemon running - in such case it should probably call log_capture_start/end to get the debug messages from the daemon and display it on the console (and also keep it in the main tuned.log).

If you have different idea, please share.

PR #419, opened 2022-03-14
gui: fixed possible traceback in the save_profile

yarda:
Resolves: rhbz#2058902

Signed-off-by: Jaroslav Škarvada <jskarvad@redhat.com>

PR #420, opened 2022-03-14
beakerlib: fix systemd rate limiting for variables-support-in-profiles

yarda:


yarda:
Regarding the centos-stream-9 this is the same problem as noted in the #419.

CZerta (approved):
Seems OK, in long run should be set before tests otherwise this can pop up in any fifth+ test if service is restarted 5+ times in ten seconds (default)

yarda:
This PR is just temporal fix, there is an effort to fix it upstream in the beakerlib/tuned library:
https://github.com/beakerlib/tuned/pull/3

Once merged upstream we can instrument the tests. @RHEmployee will create the PR.

PR #421, opened 2022-03-16
Update tests: disable systemd rate limiting.

RHEmployee:


tests/beakerlib/Tuned-takes-too-long-to-reload-start-when-ulimit/main.fmf
```diff
@@ -8,3 +8,5 @@ relevancy: |
     distro = rhel-4, rhel-5, rhel-6: False
 summary: Test for BZ#1663412 (TuneD takes too long to reload/start when \"ulimit)
 framework: beakerlib
+require:
+  - library(tuned/basic
```
CZerta, line 12: On first sight, typo missing closing parenthesis?

CZerta (changes requested):

tests/beakerlib/Tuned-takes-too-long-to-reload-start-when-ulimit/main.fmf
RHEmployee: Cut and pasta error. Fixed now.

yarda (changes requested):
@RHEmployee it seems you completely broke it:
```
[32m[10:23:38] [+] [test-schedule-tmt-connect] cloning repo https://github.com/RHEmployee/tuned-1 (branch master, ref d15b0d713f3b98016ff610e30e30fca53d5e8c18)[0m
[31m[10:23:39] [E] [test-schedule-tmt-connect] Failed to get list of plans:
---v---v---v---v---v---
Invalid yaml syntax: Failed to parse '/var/ARTIFACTS/workdir-repository-None-n_lf4L/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/main.fmf'.
while constructing a mapping
  in "/var/ARTIFACTS/workdir-repository-None-n_lf4L/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/main.fmf", line 1, column 1
found duplicate key "require" with value "[]" (original value: "[]")
  in "/var/ARTIFACTS/workdir-repository-None-n_lf4L/tests/beakerlib/bz1798183-RFE-support-post-loaded-profile/main.fmf", line 14, column 1
```
I.e. it seems you have multiple 'require' sections in the `beakerlib/bz1798183-RFE-support-post-loaded-profile/main.fmf"`

Also could you please be consistent with the tabs/spaces in the existing code and do not intermix them?

RHEmployee:
Hi. Require issue should be fixed.

RHEmployee:
Hi. Seems like changes from review has been reflected.
Would you please take a look?

Testing is still in progress.

yarda:
You are still inconsistent with tab & spaces, please fix it, e.g.:
```
         rlRun "pushd $TmpDir"
+\trlImport "tuned/basic"
```
E.g. there was 8 spaces in the test, but you used one tab. Either use spaces or convert the test to tabs.

RHEmployee:
> You are still inconsistent with tab & spaces, please fix it, e.g.:
> 
> ```
>          rlRun "pushd $TmpDir"
> +\trlImport "tuned/basic"
> ```
> 
> E.g. there were 8 spaces in the test, but you used one tab. Either use spaces or convert the test to tabs.

Thanks. It's just legacy of too many people touching tests :).
Should be fixed now.

tests/beakerlib/Tuned-takes-too-long-to-reload-start-when-ulimit/runtest.sh
```diff
@@ -41,6 +43,11 @@ rlJournalStart
         rlRun "popd"
         rlRun "rm -r $TmpDir" 0 "Removing tmp directory"
 
+        killall tuned
+        rlRun "sleep 3"
```
yarda, line 47: What's the purpose of the double SIGTERM? The first kilall sends SIGTERM then it waits 3 seconds and the SIGTERM is resend. Why? I would understand if you would send SIGKILL instead of the SIGTERM the second time. But probably much better approach is to let it on the systemd - it sends SIGTERM the first time and then after the systemd timeout it send SIGKILL.

yarda (changes requested):

tests/beakerlib/Tuned-takes-too-long-to-reload-start-when-ulimit/runtest.sh
RHEmployee: Systemd is out of way because reproducer. Also if you take a look at test, there is daemon mode disable and --debug output is used, which leads to TESTOUT file. Then, TESOUT file is checked for everything reproducer wants and then cleanup continues.

killall check is only soft check (nothing forced). So I can check visually that test which is next to this test, have system clean (so no tuned running). rlRun, on other hand, produces hard check so test can fail in cleanup phase (which can leads to warning state in beaker) which is something what I don't want (because, in beaker world, I'm forced to accept tests with warning state which I don't want). I prefer to have test with PASS and then I can check if next test fails because wrong cleanup or if test failed itself.

Is systemd able to stop service even if service doesn't run under systemd how you propose?

RHEmployee: Feel free to propose solution based on systemd.
It's seems easier to gather output via --debug than solve state of tuned logs.

yarda: I still don't get  it. If TuneD isn't gone and it should, the test should fail, shouldn't it? Or you are grepping through the beaker logs of passed tests? 

Moreover the test itself seems quite odd. It should test that TuneD start doesn't take too long, instead if checks TuneD log for message that tuned is not executing sysctl, i.e. it doesn't check that the start takes too long, it even doesn't check that TuneD isn't executing sysctl, it just check that such message isn't in the TuneD log.

RHEmployee: Ad beaker: tests are executed in queue so if next test will fail, I can check previous test (which can be this test) and take a look if tuned were killed or not. And yes, I just checking output of tests in case that something is wrong.

Ad test itself: Ok. We can move this (better handling of delay) to another MR/issue. This MR is about rate limiting and not about test itself which yes, it has been modified to pass MR checks.

RHEmployee: Additional info: 
bugzilla told me, that tuned start takes around 20 seconds without patch.
So this tests exploits this fact and if tuned start is slow, then there is no expected logs in next command (rlAssertGrep)
I admit, it's little quirky but this "race condition" just works.
tuned is expected to start around 0.5 second (almost instantly) how is described in bugzilla (link provided).

You can check comment 7 in:
https://bugzilla.redhat.com/show_bug.cgi?id=1663412

where tests results are presented.

yarda: I hope you will fix the tests in next PRs.

Regarding the two `killall` commands it's crazy hack, but if Robin Hack wants it this way, he will deserve it :)

RHEmployee: https://github.com/redhat-performance/tuned/issues/544

yarda: Btw, you could use `killall -9 tuned && rlLogWarning "TuneD still running, killed"` instead of the second `killall` then you wouldn't have to grep through the logs and it wouldn't break other tests.

PR #422, opened 2022-03-16
tests: enable unit tests on RHEL-7 (EPEL is needed)

yarda:


CZerta (approved):
Seems OK, looks that check and tests are run on epel in farm and not on clean RHEL7

PR #428, opened 2022-03-24
recommend: do not ignore syspurpose_role if there is no syspurpose

yarda:
Previously, if there wasn't syspurpose (usually provided by the
subscription-manger), all the 'syspurpose_role' matches were ignored.
This wasn't correct, because it didn't allow enforcing of the system
role. Now in such cases the syspurpose_role is matched against the
empty string, i.e. the empty string means 'no syspurpose role or
syspurpose error'.

E.g. previously the following rule matched in case of no syspurpose:
something_that_matched
syspurpose_role=.*atomic.*

Now it doesn't match, but the previous behavior can be still
emulated by e.g.:
something_that_matched
syspurpose_role=(.*atomic.*)|^$

Also the error about the missing syspurpose module is now displayed only
once (for each 'recommend' call) and not multiple times for each
recommend rule.

Resolves: rhbz#2030580

### DIFF
--- a/tuned/utils/profile_recommender.py
+++ b/tuned/utils/profile_recommender.py
@@ -56,6 +56,7 @@ class ProfileRecommender:
 
 	def process_config(self, fname, has_root=True):
 		matching_profile = None
+		syspurpose_error_logged = False
 		try:
 			if not os.path.isfile(fname):
 				return None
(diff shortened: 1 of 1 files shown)


martinpitt:
Thanks for the fix! Any chance this can come with a test, such as calling `tuned-adm auto_profile` in fmf? That always ought to succeed

martinpitt:
Any chance that this can be reviewed/landed? This has been broken in C9S for months now. [current log](https://artifacts.dev.testing-farm.io/83437dab-10c1-41af-bd77-73ae1c4d9265/)

yarda:
> Thanks for the fix! Any chance this can come with a test, such as calling `tuned-adm auto_profile` in fmf? That always ought to succeed

The test would have to handle situations where syspurpose is installed. Also package providing syspurpose had different name in the past. All such situations would have to be handled by the test, because we reuse tests on RHEL. But I am OK to review/merge if anybody PRs it.

PR #431, opened 2022-04-04
Fix in polkit authorization.

CZerta:
As *args is tuple we cannot change its values. This cause exception for example when calling dbus without sudo and fallback for tuned-adm, but this behavior at least confusing.


CZerta:
Maybe we can add here at least check to tuned-adm if it runs with root privileges and don't call dbus, otherwise it receives empty values without any warning for users.

yarda:
> Maybe we can add here at least check to tuned-adm if it runs with root privileges and don't call dbus, otherwise it receives empty values without any warning for users.

Yes, I agree that tuned-adm should be extended to report this state This can be done in future PRs or maybe please open RFE github ticket at least not to forget about it.

yarda (approved):
LGTM, thanks.

PR #432, opened 2022-04-04
bootloader: cmdline parser change and fixed escaping

yarda:
The following is now parsed differently:

[bootloader]
cmdline =+

Previously it was parsed as operator "=" adding symbol "+" to the cmdline,
now it is parsed as operator "=+" adding empty string "".

For the previous behaviour the leading "+" can now be escaped. It works the
following way:
cmdline =+ -> operator "=+" adding ""
cmdline =\\+ -> operator "=" adding "+"
cmdline =\\\\+ -> operator "=" adding "\\+"

Non leading escaping is ignored, e.g.:
cmdline =ab\\+ -> operator "=" adding "ab\\+"

Similarly for the operator "=-".

Also the arguments are now correctly escaped when further processed by
TuneD and the following problems should be now fixed:

cmdline =\n -> corruption of the /etc/tuned/bootcmdline with newlines
cmdline =\k -> TuneD traceback
cmdline =\1 -> operator "=" adding "TUNED_BOOT_CMDLINE="

There could be more undesired effects and that's why the escaping different
from the previously described leading escaping "\\\\", "\\+", "\\-" should be
now ignored.

Resolves: rhbz#1884472

### DIFF
--- a/tuned/plugins/plugin_bootloader.py
+++ b/tuned/plugins/plugin_bootloader.py
@@ -164,13 +164,16 @@ class BootloaderPlugin(base.Plugin):
 			if val is None or val == "":
 				continue
 			op = val[0]
+			op1 = val[1:2]
 			vals = val[1:].strip()
-			if op == "+" and vals != "":
-				cmdline += " " + vals
-			elif op == "-" and vals != "":
-				for p in vals.split():
-					regex = re.escape(p)
-					cmdline = re.sub(r"(\A|\s)" + regex + r"(?=\Z|\s)", r"", cmdline)
+			if op == "+" or (op == "\\" and op1 in ["\\", "+", "-"]):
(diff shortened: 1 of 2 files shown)


jmencak (approved):
Thank you for the PR, @yarda .  This looks good to me.  I've tested the PR too.  Could you fix the PR description though?  The double '\\\\' get squashed into one '\\'.  AFAICS fix is needed in 3 places.

yarda:
@jmencak thanks for the review and for noticing the bad escaping in the comment :) I fixed the comment, description in the PR should be OK.

jmencak:
> @jmencak thanks for the review and for noticing the bad escaping in the comment :) I fixed the comment, description in the PR should be OK.

Thank you for the fixes, looks good to me now.

PR #439, opened 2022-05-11
Closing fd from perf module in scheduler plugin

CZerta:
perf module does not close it's file descriptors and it seams there is no better/nicer interface in python.

Resolves: rhbz#2080227


yarda:
LGTM, thanks.

yarda (approved):

PR #442, opened 2022-06-09
Add profile_info command automatic complete

uniontech-lilinjie:
Add profile_info command automatic complete
Users performing profile_info use the tab key to view the system profile file


uniontech-lilinjie:
Both modifications are ok after testing

PR #450, opened 2022-06-21
spec: dropped weak deps where not needed

yarda:
Resolves: rhbz#2093841


yarda:
Kept weakdeps for kernel-tools, because it has excludearch on i686 thus theoretically somebody could experience failure when installing it from some unsupported spin (i686 RPMs are still built).

yarda:
The epel-7 failure is probably packit issue https://github.com/packit/packit-service/issues/1601

CZerta (approved):
LGTM

PR #453, opened 2022-08-08
builtin functions: added calc_isolated_cores function

yarda:
The calc_isolated_cores function expands to the list of cores to
isolate. It accepts optional argument which specifies how many cores
from each socket reserve for housekeeping. If not specified, one core
from each socket is reserved for housekeeping.

Example:
Machine with 2 sockets, each 4 cores, using the following user variable
configuration file, e.g. /etc/tuned/realtime-variables.conf:
isolated_cores=${f:calc_isolated_cores:2}

It will expand to:
isolated_cores=2, 3, 6, 7

I.e. cores 0, 1 and 4, 5 will be used for housekeeping.

Resolves: rhbz#2093847

### DIFF
--- a/tuned/profiles/functions/function_calc_isolated_cores.py
+++ b/tuned/profiles/functions/function_calc_isolated_cores.py
@@ -1,0 +1,44 @@
+import os
+import glob
+import tuned.logs
+from . import base
+import tuned.consts as consts
+
+log = tuned.logs.get()
+
+class calc_isolated_cores(base.Function):
+	"""
+	Calculates and returns isolated cores. The argument specifies how many
+	cores per socket reserve for housekeeping. If not specified, 1 core
+	per socket is reserved for housekeeping and the rest is isolated.
+	"""
+	def __init__(self):
+		# max 1 argument
+		super(calc_isolated_cores, self).__init__("calc_isolated_cores", 1)
+
+	def execute(self, args):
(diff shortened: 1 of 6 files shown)


yarda:
EPEL-7 failure is probably caused by some changes on the testing farm.

tuned/profiles/functions/function_calc_isolated_cores.py
```diff
@@ -0,0 +1,44 @@
...
+	def __init__(self):
+		# max 1 argument
+		super(calc_isolated_cores, self).__init__("calc_isolated_cores", 1)
+
+	def execute(self, args):
+		if not super(calc_isolated_cores, self).execute(args):
+			return None
+		cpus_reserve = 1
+		if len(args) > 0:
+			if not args[0].isdecimal() or int(args[0]) < 0:
```
CZerta, line 24: Shouldn't there be `int(args[0]) <= 0` to always have at least one housekeeping core? If there can be no isolated core the error log on next line probably should say `... non-negative integer`

CZerta (changes requested):

tuned/profiles/functions/function_calc_isolated_cores.py
yarda: 0 is valid setting - to have no housekeeping core. I will update the error message.

CZerta (approved):
LGTM

PR #454, opened 2022-08-10
profiles: drop sched_ tuning from openshift-control-plane

dagrayvid:
According to testing done by the kernel QE and performance teams,
the sched_* tunables should be dropped in several profiles.

The openshift-control-plane profile is based on other profiles
which have already dropped these tunables (network-latency,
throughput-performance). 

@yarda We would like to drop these tunables on RHEL9, but keep them on RHEL8. is it possible with a patch to ensure this change only takes effect on RHEL9, at least for the FDP builds that will feed into the Node Tuning Operator?

/cc @jmencak

### DIFF
--- a/profiles/openshift-control-plane/tuned.conf
+++ b/profiles/openshift-control-plane/tuned.conf
@@ -5,19 +5,3 @@
 [main]
 summary=Optimize systems running OpenShift control plane
 include=openshift
-
-[scheduler]
-# ktune sysctl settings, maximizing i/o throughput
-#
-# Minimal preemption granularity for CPU-bound tasks:
-# (default: 1 msec#  (1 + ilog(ncpus)), units: nanoseconds)
-sched_min_granularity_ns=10000000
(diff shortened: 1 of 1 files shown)


jmencak (approved):
LGTM. We [talked about this in the past](https://github.com/redhat-performance/tuned/pull/376#pullrequestreview-722019407).

yarda:
> @yarda We would like to drop these tunables on RHEL9, but keep them on RHEL8. is it possible with a patch to ensure this change only takes effect on RHEL9, at least for the FDP builds that will feed into the Node Tuning Operator?

@dagrayvid yes, it's possible.

FYI the EPEL-7 CI failure is caused by Packit upstream bug which will be addressed soon.

We now have dev freeze, but I am going to start merging the PRs in cca. week or so.

dagrayvid:
@yarda, is this PR able to be merged now that #455 has merged?

PR #455, opened 2022-08-15
Packit yaml improvements

TomasTomecek:
We have made a few changes to packit.yaml syntax - so that yours is up to date.

The most important change is the switch to building SRPMs in Copr which is our long-term plan. That should fix the el7 build problems right now.


yarda:
Thanks.

PR #461, opened 2022-09-06
Profile openshift: increase fs.aio-max-nr tunable

jmencak:
The kernel tunable `fs.aio-max-nr` specifies the maximum number of events specified on the `io_setup()` call for all currently active aio contexts.  On large systems this causes issues with the userspace `io_setup()` call failing with `EAGAIN`.  Libvirt already increases this setting to 1048576:
https://github.com/libvirt/libvirt/blob/96e99e49483de4fb293ce26691cbce19110fb3a9/src/remote/libvirtd.sysctl

The only potential drawback of a larger limit is that a malicious users could issue parallel `io_setup()` requests to cause the kernel to set aside memory.

### DIFF
--- a/profiles/openshift/tuned.conf
+++ b/profiles/openshift/tuned.conf
@@ -15,6 +15,7 @@ nf_conntrack_hashsize=1048576
 [sysctl]
 net.ipv4.ip_forward=1
 kernel.pid_max=>4194304
+fs.aio-max-nr=>1048576
 net.netfilter.nf_conntrack_max=1048576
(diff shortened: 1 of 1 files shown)


jmencak:
Linking [Jira issue](https://issues.redhat.com/browse/PSAP-900).

yarda:
CentOS-7 failure is probably caused by https://gitlab.com/testing-farm/general/-/issues/51.

yarda:
LGTM, thanks.

PR #464, opened 2022-09-29
realtime-virtual-*: Tune priority of new ktimers threads

valschneider:
The PREEMPT_RT upstream tree [1] is reintroducing a per-CPU kthread for handling timer expiry named ktimers/${cpu}, and RHEL-RT is following this approach [2].

The recommended priority tuning for that kthread is to simply match that of ksoftirqd, which is what is implemented here.

Note that ktimersoftd is the predecessor of ktimers, and isn't present on RT kernels newer than v5.0.19-rt11 / RHEL-RT 7.

[1]: https://git.kernel.org/pub/scm/linux/kernel/git/rt/linux-rt-devel.git
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=2122220

Resolves: rhbz#2122220
Signed-off-by: Valentin Schneider <vschneid@redhat.com>

### DIFF
--- a/profiles/realtime-virtual-guest/tuned.conf
+++ b/profiles/realtime-virtual-guest/tuned.conf
@@ -25,6 +25,7 @@ assert2=${f:assertion:isolated_cores contains online CPU(s):${isolated_cores_exp
 # group.group_name=rule_priority:scheduler_policy:scheduler_priority:core_affinity_in_hex:process_name_regex
 # for i in `pgrep ksoftirqd` ; do grep Cpus_allowed_list /proc/$i/status ; done
 group.ksoftirqd=0:f:2:*:^\[ksoftirqd
+group.ktimers=0:f:2:*:^\[ktimers
 
 # for i in `pgrep rcuc` ; do grep Cpus_allowed_list /proc/$i/status ; done
 group.rcuc=0:f:4:*:^\[rcuc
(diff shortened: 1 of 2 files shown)


profiles/realtime-virtual-guest/tuned.conf
```diff
@@ -25,6 +25,7 @@ assert2=${f:assertion:isolated_cores contains online CPU(s):${isolated_cores_exp
 # group.group_name=rule_priority:scheduler_policy:scheduler_priority:core_affinity_in_hex:process_name_regex
 # for i in `pgrep ksoftirqd` ; do grep Cpus_allowed_list /proc/$i/status ; done
 group.ksoftirqd=0:f:2:*:^\[ksoftirqd
+group.ktimers=0:f:2:*:^\[ktimers
```
valschneider, line 28: I'm not sure what the `# for i in ...` comment above is there for so I've just squashed the new kthread priority assignment against ksoftirqd. If the comment is a required convention, I'll change that.

yarda:
Failure of the centos-7 test is infrastructure issue, i.e. it's unrelated to this commit.

profiles/realtime-virtual-guest/tuned.conf
yarda: It seems these comments originated from patch provided by Jeremy Eder <jeder@redhat.com> and Clark Williams <williams@redhat.com>. I guess it's just helper to remember how to debug/verify CPU affinity for the rule. We could probably drop them.

yarda (approved):
LGTM, thanks.

PR #465, opened 2022-10-11
systemd: relax polkit requirement

yarda:
The polkit has D-bus activation thus it will be activated on demand and we don't need to explicitly activate it.

Resolves: rhbz#2065591


CZerta (approved):
LGTM

PR #467, opened 2022-10-20
Fixed sysvinit path

yarda:
It's kept for legacy purposes, but the path should be correct.

Resolves: rhbz#2118301


CZerta (approved):
LGTM

PR #468, opened 2022-10-26
add 'tuned-adm delete' command

uniontech-lilinjie:
add 'tuned-adm delete' command
Convenient for users to delete their own generated profile. (dir: /etc/tuned/)


yarda:
Please squash commits to one.

PR #470, opened 2022-10-26
Expose TuneD API to the Unix Domain Socket.

CZerta:
TuneD is listening on paths specified in config in option unix_socket_paths and send signals to paths in option unix_socket_signal_paths. Example call:

ssock_file = "/tmp/tuned_socket"
csock_file = "/tmp/tuned_socket_client"
if os.path.exists(csock_file):
	os.remove(csock_file)
s=socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM) s.bind(csock_file)
try:
	s.sendto(pickle.dumps(("switch_profile", "balanced")), ssock_file)
	print(pickle.loads(s.recvfrom(4096)[0]))
except Exception as e:
	print("E: %s" % e)
s.close()

This PR also introduce possibility to disable dbus API in main TuneD config.

Resolves: rhbz#2113900

### DIFF
--- a/tuned-main.conf
+++ b/tuned-main.conf
@@ -49,3 +49,29 @@ log_file_max_size = 1MB
 # It can be used to force tuning for specific architecture.
 # If commented, "/proc/cpuinfo" will be read to fill its content.
 # cpuinfo_string = Intel
+
+# Enable TuneD listening on dbus
+# enable_dbus = 1
+
+# Enable TuneD listening on unix domain socket
+# As this functionality is not used commonly, we disable it by default
+# and it is needed to allow it manually
+# enable_unix_socket = 0
+
+# Path to socket for TuneD to listen
+# Existing files on given path will be removed
+# unix_socket_path = /run/tuned/tuned.sock
+
+# Paths to sockets for TuneD to send signals to separated by , or ;
+# unix_socket_signal_paths =
+
+# Default unix socket ownership
+# Can be set as id or name, -1 or non-existing name leaves unchanged
+# unix_socket_ownership = -1 -1
+
+# Permissions for listening sockets
+# unix_socket_permissions = 0o600
+
+# Size of connections backlog for listen function on socket
+# Higher value allows to process requests from more clients
+# connections_backlog = 1024
--- a/tuned/consts.py
+++ b/tuned/consts.py
@@ -100,6 +100,13 @@ CFG_LOG_FILE_COUNT = "log_file_count"
 CFG_LOG_FILE_MAX_SIZE = "log_file_max_size"
 CFG_UNAME_STRING = "uname_string"
 CFG_CPUINFO_STRING = "cpuinfo_string"
+CFG_ENABLE_DBUS = "enable_dbus"
+CFG_ENABLE_UNIX_SOCKET = "enable_unix_socket"
+CFG_UNIX_SOCKET_PATH = "unix_socket_path"
+CFG_UNIX_SOCKET_SIGNAL_PATHS = "unix_socket_signal_paths"
+CFG_UNIX_SOCKET_OWNERSHIP = "unix_socket_ownership"
+CFG_UNIX_SOCKET_PERMISIONS = "unix_socket_permissions"
+CFG_UNIX_SOCKET_CONNECTIONS_BACKLOG = "connections_backlog"
 
 # no_daemon mode
 CFG_DEF_DAEMON = True
@@ -129,7 +136,24 @@ CFG_DEF_LOG_FILE_COUNT = 2
 CFG_FUNC_LOG_FILE_COUNT = "getint"
 # default log file max size
 CFG_DEF_LOG_FILE_MAX_SIZE = 1024 * 1024
-
+# default listening on dbus
+CFG_DEF_ENABLE_DBUS = True
+CFG_FUNC_ENABLE_DBUS = "getboolean"
+# default listening on unix socket
+# as it is not used commonly disabled by default
+CFG_DEF_ENABLE_UNIX_SOCKET = False
+CFG_FUNC_ENABLE_UNIX_SOCKET = "getboolean"
+# default unix socket path
+CFG_DEF_UNIX_SOCKET_PATH = "/run/tuned/tuned.sock"
+CFG_DEF_UNIX_SOCKET_SIGNAL_PATHS = ""
+# default unix socket ownership
+# (uid and gid, python2 does not support names out of box, -1 leaves default)
+CFG_DEF_UNIX_SOCKET_OWNERSHIP = "-1 -1"
+# default unix socket permissions
+CFG_DEF_UNIX_SOCKET_PERMISIONS = "0o600"
+# default unix socket conections backlog
+CFG_DEF_UNIX_SOCKET_CONNECTIONS_BACKLOG = "1024"
+CFG_FUNC_UNIX_SOCKET_CONNECTIONS_BACKLOG = "getint"
 
 PATH_CPU_DMA_LATENCY = "/dev/cpu_dma_latency"
 
@@ -137,7 +161,7 @@ PATH_CPU_DMA_LATENCY = "/dev/cpu_dma_latency"
 PROFILE_ATTR_SUMMARY = "summary"
 PROFILE_ATTR_DESCRIPTION = "description"
 
-DBUS_SIGNAL_PROFILE_CHANGED = "profile_changed"
+SIGNAL_PROFILE_CHANGED = "profile_changed"
 
 STR_HINT_REBOOT = "you need to reboot for changes to take effect"
 
--- a/tuned/daemon/daemon.py
+++ b/tuned/daemon/daemon.py
@@ -173,8 +173,8 @@ class Daemon(object):
 	# send notification when profile is changed (everything is setup) or if error occured
 	# result: True - OK, False - error occured
 	def _notify_profile_changed(self, profile_names, result, errstr):
-		if self._application is not None and self._application._dbus_exporter is not None:
-			self._application._dbus_exporter.send_signal(consts.DBUS_SIGNAL_PROFILE_CHANGED, profile_names, result, errstr)
+		if self._application is not None:
+			exports.send_signal(consts.SIGNAL_PROFILE_CHANGED, profile_names, result, errstr)
 		return errstr
 
 	def _full_rollback_required(self):
--- a/tuned/exports/unix_socket_exporter.py
+++ b/tuned/exports/unix_socket_exporter.py
@@ -1,0 +1,246 @@
+import os
+import re
+import pwd, grp
+
+from . import interfaces
+import tuned.logs
+import tuned.consts as consts
+from inspect import ismethod
+import socket
+import json
+import select
+
+log = tuned.logs.get()
+
+class UnixSocketExporter(interfaces.ExporterInterface):
+	"""
+	Export method calls through Unix Domain Socket Interface.
+
+	We take a method to be exported and create a simple wrapper function
+	to call it. This is required as we need the original function to be
+	bound to the original object instance. While the wrapper will be bound
+	to an object we dynamically construct.
+	"""
+
+	def __init__(self, socket_path=consts.CFG_DEF_UNIX_SOCKET_PATH,
+				 signal_paths=consts.CFG_DEF_UNIX_SOCKET_SIGNAL_PATHS,
+				 ownership=consts.CFG_DEF_UNIX_SOCKET_OWNERSHIP,
+				 permissions=consts.CFG_DEF_UNIX_SOCKET_PERMISIONS,
+				 connections_backlog=consts.CFG_DEF_UNIX_SOCKET_CONNECTIONS_BACKLOG):
+
+		self._socket_path = socket_path
+		self._socket_object = None
+		self._socket_signal_paths = re.split(r",;", signal_paths) if signal_paths else []
+		self._socket_signal_objects = []
+		self._ownership = [-1, -1]
+		if ownership:
+			ownership = ownership.split()
+			for i, o in enumerate(ownership[:2]):
+				try:
+					self._ownership[i] = int(o)
+				except ValueError:
+					try:
+						# user
+						if i == 0:
+							self._ownership[i] = pwd.getpwnam(o).pw_uid
+						# group
+						else:
+							self._ownership[i] = grp.getgrnam(o).gr_gid
+					except KeyError:
+						log.error("%s '%s' does not exists, leaving default" % ("User" if i == 0 else "Group", o))
+		self._permissions = permissions
+		self._connections_backlog = connections_backlog
+
+		self._unix_socket_methods = {}
+		self._signals = set()
+		self._conn = None
+		self._channel = None
+
+	def running(self):
+		return self._socket_object is not None
+
+	def export(self, method, in_signature, out_signature):
+		if not ismethod(method):
+			raise Exception("Only bound methods can be exported.")
+
+		method_name = method.__name__
+		if method_name in self._unix_socket_methods:
+			raise Exception("Method with this name (%s) is already exported." % method_name)
+
+		class wrapper(object):
+			def __init__(self, in_signature, out_signature):
+				self._in_signature = in_signature
+				self._out_signature = out_signature
+				
+			def __call__(self, *args, **kwargs):
+				return method(*args, **kwargs)
+
+		self._unix_socket_methods[method_name] = wrapper(in_signature, out_signature)
+
+	def signal(self, method, out_signature):
+		if not ismethod(method):
+			raise Exception("Only bound methods can be exported.")
+		
+		method_name = method.__name__
+		if method_name in self._unix_socket_methods:
+			raise Exception("Method with this name (%s) is already exported." % method_name)
+		
+		class wrapper(object):
+			def __init__(self, out_signature):
+				self._out_signature = out_signature
+			
+			def __call__(self, *args, **kwargs):
+				return method(*args, **kwargs)
+		
+		self._unix_socket_methods[method_name] = wrapper(out_signature)
+		self._signals.add(method_name)
+
+	def send_signal(self, signal, *args, **kwargs):
+		if not signal in self._signals:
+			raise Exception("Signal '%s' doesn't exist." % signal)
+		for p in self._socket_signal_paths:
+			log.debug("Sending signal on socket %s" % p)
+			try:
+				s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+				s.setblocking(False)
+				s.connect(p)
+				self._send_data(s, {"jsonrpc": "2.0", "method": signal, "params": args})
+				s.close()
+			except OSError as e:
+				log.warning("Error while sending signal '%s' to socket '%s': %s" % (signal, p, e))
+
+	def register_signal_path(self, path):
+		self._socket_signal_paths.append(path)
+
+	def _construct_socket_object(self):
+		if self._socket_path:
+			if os.path.exists(self._socket_path):
+				os.unlink(self._socket_path)
+			self._socket_object = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+			self._socket_object.bind(self._socket_path)
+			self._socket_object.listen(self._connections_backlog)
+			os.chown(self._socket_path, self._ownership[0], self._ownership[1])
+			if self._permissions:
+				os.chmod(self._socket_path, self._permissions)
+
+	def start(self):
+		if self.running():
+			return
+
+		self.stop()
+		self._construct_socket_object()
+
+	def stop(self):
+		if self._socket_object:
+			self._socket_object.close()
+
+	def _send_data(self, s, data):
+		log.debug("Sending socket data: %s)" % data)
+		try:
+			s.send(json.dumps(data).encode("utf-8"))
+		except Exception as e:
+			log.warning("Failed to send data '%s': %s" % (data, e))
+
+	def _create_response(self, data, id, error=False):
+		res = {
+			"jsonrpc": "2.0",
+			"id": id
+		}
+		if error:
+			res["error"] = data
+		else:
+			res["result"] = data
+		return res
+
+	def _create_error_responce(self, code, message, id=None, data=None):
+		return self._create_response({
+			"code": code,
+			"message": message,
+			"data": data,
+		}, error=True, id=id)
+
+	def _create_result_response(self, result, id):
+		return self._create_response(result, id)
+
+	def _check_id(self, data):
+		if data.get("id"):
+			return data
+		return None
+
+	def _process_request(self, req):
+		if type(req) != dict or req.get("jsonrpc") != "2.0" or not req.get("method"):
+			return self._create_error_responce(-32600, "Invalid Request")
+		id = req.get("id")
+		ret = None
+		if req["method"] not in self._unix_socket_methods:
+			return self._check_id(self._create_error_responce(-32601, "Method not found", id))
+		try:
+			if not req.get("params"):
+				ret = self._unix_socket_methods[req["method"]]()
+			elif type(req["params"]) in (list, tuple):
+				ret = self._unix_socket_methods[req["method"]](*req["params"])
+			elif type(req["params"]) == dict:
+				ret = self._unix_socket_methods[req["method"]](**req["params"])
+			else:
+				return self._check_id(self._create_error_responce(-32600, "Invalid Request", id))
+		except TypeError as e:
+			return self._check_id(self._create_error_responce(-32602, "Invalid params", id, str(e)))
+		except Exception as e:
+			return self._check_id(self._create_error_responce(1, "Error", id, str(e)))
+		return self._check_id(self._create_result_response(ret, id))
+
+	def period_check(self):
+		"""
+		Periodically checks socket object for new calls. This allows to function without special thread.
+		Interface is according JSON-RPC 2.0 Specification (see https://www.jsonrpc.org/specification)
+		
+		Example calls:
+		
+		printf '[{"jsonrpc": "2.0", "method": "active_profile", "id": 1}, {"jsonrpc": "2.0", "method": "profiles", "id": 2}]' | nc -U /run/tuned/tuned.sock
+		printf '{"jsonrpc": "2.0", "method": "switch_profile", "params": {"profile_name": "balanced"}, "id": 1}' | nc -U /run/tuned/tuned.sock
+		"""
+		if not self.running():
+			return
+		while True:
+			r, _, _ = select.select([self._socket_object], (), (), 0)
+			if r:
+				conn, _ = self._socket_object.accept()
+				try:
+					data = ""
+					while True:
+						rec_data = conn.recv(4096).decode()
+						if not rec_data:
+							break
+						data += rec_data
+				except Exception as e:
+					log.error("Failed to load data of message: %s" % e)
+					continue
+				if data:
+					try:
+						data = json.loads(data)
+					except Exception as e:
+						log.error("Failed to load json data '%s': %s" % (data, e))
+						self._send_data(conn, self._create_error_responce(-32700, "Parse error", str(e)))
+						continue
+					if type(data) not in (tuple, list, dict):
+						log.error("Wrong format of call")
+						self._send_data(conn, self._create_error_responce(-32700, "Parse error", str(e)))
+						continue
+					if type(data) in (tuple, list):
+						if len(data) == 0:
+							self._send_data(conn, self._create_error_responce(-32600, "Invalid Request", str(e)))
+							continue
+						res = []
+						for req in data:
+							r = self._process_request(req)
+							if r:
+								res.append(r)
+						if res:
+							self._send_data(conn, res)
+					else:
+						res = self._process_request(data)
+						if r:
+							self._send_data(conn, res)
+			else:
+				return
+		
--- a/tuned.py
+++ b/tuned.py
@@ -41,6 +41,7 @@ if __name__ == "__main__":
 	parser.add_argument("--log", "-l", nargs = "?", const = consts.LOG_FILE, help = "log to file, default file: " + consts.LOG_FILE)
 	parser.add_argument("--pid", "-P", nargs = "?", const = consts.PID_FILE, help = "write PID file, default file: " + consts.PID_FILE)
 	parser.add_argument("--no-dbus", action = "store_true", help = "do not attach to DBus")
+	parser.add_argument("--no-socket", action = "store_true", help = "do not attach to socket")
 	parser.add_argument("--profile", "-p", action = "store", type=str, metavar = "name", help = "tuning profile to be activated")
 	parser.add_argument('--version', "-v", action = "version", version = "%%(prog)s %s.%s.%s" % (ver.TUNED_VERSION_MAJOR, ver.TUNED_VERSION_MINOR, ver.TUNED_VERSION_PATCH))
 	args = parser.parse_args(sys.argv[1:])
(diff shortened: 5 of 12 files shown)


yarda (changes requested):

tuned/consts.py
```diff
@@ -129,15 +135,29 @@
...
+CFG_FUNC_DBUS = "getboolean"
+# default listening on unix socket
+CFG_DEF_UNIX_SOCKET = True
+CFG_FUNC_UNIX_SOCKET = "getboolean"
+# default unix socket paths
+CFG_DEF_UNIX_SOCKET_PATHS = "/tmp/tuned_socket"
+CFG_DEF_UNIX_SOCKET_SIGNAL_PATHS = ""
+# default unix socket permissions
+CFG_DEF_UNIX_SOCKET_PERMISIONS = 0o444
+CFG_FUNC_UNIX_SOCKET_PERMISIONS = "getint"
```
yarda, line 149: What is it used for?

tuned/consts.py
```diff
@@ -129,15 +135,29 @@
...
+# default byte len for unix sockets
+CFG_DEF_UNIX_SOCKET_BYTE_LEN = 4096
+CFG_FUNC_UNIX_SOCKET_BYTE_LEN = "getint"
```
yarda, line 152: What is it used for?

tuned/consts.py
```diff
@@ -129,15 +135,29 @@
 CFG_FUNC_LOG_FILE_COUNT = "getint"
 # default log file max size
 CFG_DEF_LOG_FILE_MAX_SIZE = 1024 * 1024
-
+# default listening on dbus
+CFG_DEF_DBUS = True
+CFG_FUNC_DBUS = "getboolean"
```
yarda, line 140: What is it used for?

tuned/consts.py
CZerta: This is default value and get function describing type of option. See https://github.com/redhat-performance/tuned/blob/420267fa3463ef4a06885847c7ca4cd24c781dc8/tuned/utils/global_config.py#L28

jmencak (commented):
Thank you for the PR!  I didn't go through the code carefully (yet), but I have a few initial observations.  Also, running the demo with this code:

```
#!/usr/bin/python3

import os
import socket
import pickle
import json
import struct

def sss(d):
	ssock_file = "/tmp/tuned_socket"
	csock_file = "/tmp/tuned_socket_client"
	if os.path.exists(csock_file):
		os.remove(csock_file)
	s=socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
	s.bind(csock_file)
	j = json.dumps(d).encode("utf-8")
	try:
		s.sendto(struct.pack("!L", len(j)), ssock_file)
		s.sendto(j, ssock_file)
		d = s.recvfrom(5)
		t, l = struct.unpack("!sL", d[0])
		print("TYPE: %s" % t)
		print("LEN: %s" % l)
		data = json.loads(s.recvfrom(l)[0])
		print("DATA: %s" % data)
	except Exception as e:
		print("E: %s" % e)
	s.close()
	
sss(("profiles",))
```

Takes more than 7 seconds for me.  Is this expected?

```
[root@b86 tuned]# time ./talk.py
TYPE: b'r'
LEN: 434
DATA: ['accelerator-performance', 'balanced', 'bz', 'chrony', 'cpu-partitioning', 'desktop', 'empty', 'hpc-compute', 'inheritance', 'intel-sst', 'latency-performance', 'loop', 'network-latency', 'network-throughput', 'openshift-realtime', 'optimize-serial-console', 'plus', 'powersave', 'provider-', 'provider-gce', 'provider_', 'realtime', 'scheduler', 'script', 'stalld', 'test', 'throughput-performance', 'virtual-guest', 'virtual-host']

real    0m7.777s
user    0m0.025s
sys     0m0.004s
```

tuned-main.conf
```diff
@@ -49,3 +49,19 @@ log_file_max_size = 1MB
...
+
+# Path to socket for TuneD to listen
+# Existing files on given path will be removed
+# unix_socket_path = /tmp/tuned_socket
+
+# Paths to sockets for TuneD to send signals to separated by , or ;
+# unix_socket_signal_paths =
+
+# Permissions for listening sockets
+# unix_socket_permissions = 0o444
```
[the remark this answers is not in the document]
CZerta: Good catch, classic reading from config does not support non 10-base literals, I extended it little bit, so should be fixed.

tuned-main.conf
```diff
@@ -49,3 +49,19 @@ log_file_max_size = 1MB
...
+
+# Enable TuneD listening on dbus
+# enable_dbus = 1
+
+# Enable TuneD listening on unix domain socket
+# enable_unix_socket = 1
+
+# Path to socket for TuneD to listen
+# Existing files on given path will be removed
+# unix_socket_path = /tmp/tuned_socket
```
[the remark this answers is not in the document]
CZerta: Sorry, my mistake, I moved the check to proper loop and it should process calls in less then second with default setting.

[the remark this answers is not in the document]
jmencak: Even speedup by a factor by 10 might not be sufficient.  The socket interface is intended to be in the path of a container being created and synchronous.  I.e. it will block/delay the container from starting.  How much of an issue this is, I'm not sure at the moment, but this needs discussing with Martin Sivak (don't seem to be able to tag him here).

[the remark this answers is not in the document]
yarda: @jmencak response time about cca. 1 s is usually OK for the deamon, it's currently not better with the D-Bus. Please let us know your expectations. If response time is important for you then the overhead of another thread is acceptable, but it may require more redesign steps, because response time on the control interface and latency for the own profile switch are two distinct things.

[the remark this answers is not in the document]
yarda: Also the latency is usually tied to the power consumption.

Nevertheless, I think on the control interface side there is still a room for improvement. E.g. the select could be used for timing of the main loop and exit sooner than 1s if there is message, process it and sleep for the rest of time (in select).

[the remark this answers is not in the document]
yarda: But it will not improve the profile switch latency.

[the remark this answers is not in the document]
yarda: > Imagine trying to start 200 containers on system reboot. One sec is probably OK as long as it does not add up with load. I would have to test how exactly cri-o handles this, but I sure hope at this phase the containers are not sequential. In such case this must support 200 or more simultaneous connections that are then processed in batches of all that arrive within the same second.

This will require some serialization code on the TuneD side, the current PoC PR #477 will not scale well.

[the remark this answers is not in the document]
jmencak: > > Wouldn't `/run/tuned/tuned.sock` be a better default place for the socket? That seems to be what other components such as `chrony`, `cups`, ... use.
> 
> Depends how we decide.
 
I would strongly suggest we decide to use `/run/tuned/tuned.sock`  unless I'm missing some important reason against that location.  On modern OSes, `/run` is also mounted with `tmpfs` as well as `/tmp`.  Also, to my mind, `/tmp` is really the last resort to put files important files in not only for security, but also for "keeping things nice and tidy" reason.  `/run/tuned` belongs to `TuneD`, so why not keep it there.

tuned-main.conf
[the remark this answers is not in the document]
jmencak: This is now fixed, thank you.  However, the default permissions do not make sense to me.  I guest this was only tested with root user permissions.  In that case, `444` doesn't really matter.  However, to quote:
```
sending a datagram to a datagram socket likewise requires write permission on that socket.
```
see: https://man7.org/linux/man-pages/man7/unix.7.html

I'd suggest `0o600`, similarly `CFG_DEF_UNIX_SOCKET_PERMISIONS = "0o444"`.

On a related note, configuring access permissions on its own does not make much sense unless you couple it with also the possibility to change the owner/group.  That functionality could be useful for others, even though for our use case having only root to use the socket is good enough.

tuned/consts.py
```diff
@@ -129,15 +134,25 @@
...
+CFG_DEF_ENABLE_DBUS = True
+CFG_FUNC_ENABLE_DBUS = "getboolean"
+# default listening on unix socket
+CFG_DEF_ENABLE_UNIX_SOCKET = True
+CFG_FUNC_ENABLE_UNIX_SOCKET = "getboolean"
+# default unix socket path
+CFG_DEF_UNIX_SOCKET_PATH = "/tmp/tuned_socket"
+CFG_DEF_UNIX_SOCKET_SIGNAL_PATHS = ""
+# default unix socket permissions
+CFG_DEF_UNIX_SOCKET_PERMISIONS = "0o444"
```
jmencak, line 147: I'd suggest changing this to `0o600`.

jmencak:
During testing, I noticed I can DoS `TuneD` over the socket like this:
```
printf '\x00\x00\x00\x03' | nc -u -U /run/tuned/tuned.sock
printf 'DoS' | nc -u -U /run/tuned/tuned.sock
```

```
2022-11-22 11:07:55,987 ERROR    tuned.exports.unix_socket_exporter: Failed to load json data 'b'DoS'': Expecting value: line 1 column 1 (char 0)
Exception in thread Thread-1 (_thread_code):
  File "/usr/lib/python3.10/site-packages/tuned/exports/unix_socket_exporter.py", line 185, in period_check
    data = json.loads(data.decode("utf-8"))
  File "/usr/lib64/python3.10/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.10/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib64/python3.10/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

During handling of the above exception, another exception occurred:

  File "/usr/lib64/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3.10/site-packages/tuned/daemon/daemon.py", line 221, in _thread_code
    exports.period_check()
  File "/usr/lib/python3.10/site-packages/tuned/exports/__init__.py", line 46, in period_check
    return ctl.period_check()
  File "/usr/lib/python3.10/site-packages/tuned/exports/controller.py", line 54, in period_check
    exporter.period_check()
  File "/usr/lib/python3.10/site-packages/tuned/exports/unix_socket_exporter.py", line 188, in period_check
    self._send_data(self._socket_object, address, "err", "Failed to load json data '%s': %s" % (data, e))
  File "/usr/lib/python3.10/site-packages/tuned/exports/unix_socket_exporter.py", line 131, in _send_data
    s.sendto(struct.pack("!sL", type.encode("utf-8"), len(j)), address)
FileNotFoundError: [Errno 2] No such file or directory
```

While the client should be trusted and send valid input, it would be nice if the server (TuneD) wouldn't fall over that easily.

CZerta:
> ...
> 
> While the client should be trusted and send valid input, it would be nice if the server (TuneD) wouldn't fall over that easily.

Did you pull the newest version? I added some tries around json loading yesterday, so it should handle invalid requests better.

jmencak:
> Did you pull the newest version? I added some tries around json loading yesterday, so it should handle invalid requests better.

Yes, I did, you cannot reproduce this?

CZerta:
You can, I'll write better example. I found out what's problem with the exception. ncat does not wait for response, deletes it's socket before we process the request and try to send response to non existing file, so it fails.

jmencak:
> You can, I'll write better example.

Great, in the end though, we'll probably need a golang client.  I managed to write one.

> I found out what's problem with the exception. ncat does not wait for response, deletes it's socket before we process the request and try to send response to non existing file, so it fails

Yeah, I can see that now.  However, that's no reason for the server to fail like that.  Imagine the client doesn't care about the response like the netcat example.  That's a valid use case for me in my point of view and the server should not fail.

CZerta:
@jmencak Definitely, I added warning log when sending msg fails with reason. Also changed default path and permissions and added ownership option.

tuned-main.conf
```diff
@@ -49,3 +49,23 @@ log_file_max_size = 1MB
...
+
+# Path to socket for TuneD to listen
+# Existing files on given path will be removed
+# unix_socket_path = /run/tuned/tuned.sock
+
+# Paths to sockets for TuneD to send signals to separated by , or ;
+# unix_socket_signal_paths =
+
+# default unix socket ownership
+# (uid and gid, python2 does not support names out of box, -1 leaves default)
```
jmencak, line 67: While I applaud the effort to go the extra mile to keep things compatible, we should really consider dropping support for python2 if/when that holds us back.

CZerta: This is above "my pay grade". @yarda ?

yarda: We will drop python 2 in the future, but currently we support RHEL-7, thus we shouldn't break it.  Once RHEL-7 will go EOL we could probably drop it.

tuned/daemon/daemon.py
```diff
@@ -218,6 +218,7 @@ def _thread_code(self):
 						self._unit_manager.update_monitors()
 						log.debug("performing tunings")
 						self._unit_manager.update_tuning()
+				exports.period_check()
```
yarda, line 221: I think there is still a room for improvement. You could replace the `self._cmd.wait()` above by `epoll` with timeout and sync the time with timer. I.e. if the epoll exits after 1s, then it will be the same. If the epoll exits before 1s, then process the sockets. When the sockets are processed check the timer and run epoll with the lower timeout set.

jmencak:
> Definitely, I added warning log when sending msg fails with reason. Also changed default path and permissions and added ownership option.

@CZerta Have you tried testing with `socat`?  There are basically two native tools we can use in our environment with this.  `nc` and `socat`.  `nc` seems to work for sending, but I do not believe it is possible to receive data through it at the same time.  `socat` seems a lot more complex tool that might be able to do this, but at the moment I'm only able to do the same thing as with `nc` -- just send data:

```
socat -ddd - UNIX-CONNECT:/run/tuned/tuned.sock
```

jmencak:
Good news, sort of.  I was able to make this work with `socat`:

```
send() {
  local data="$1"
  local len=${#data}
  local b0=$(printf '%02x' ${len})	# TODO: this is limited to the payload length of 255 B
  local b1=00
  local b2=00
  local b3=00
  local socket=/run/tuned/tuned.sock

  printf "\x$b3\x$b2\x$b1\x$b0" | socat - UNIX-SENDTO:$socket,bind=/tmp/client.sock
  printf "$data" | socat - UNIX-SENDTO:$socket,bind=/tmp/client.sock
}

send '["profiles"]'
```

However, it seems not to always succeed.  One example of a failure from `TuneD`:

```
2022-12-14 16:55:00,507 ERROR    tuned.exports.unix_socket_exporter: Failed to load data of message: [Errno 11] Resource temporarily unavailable
```

This needs to be 100% reliable and working with `netcat` or `socat`.  I wonder if should switch from [SOCK_DGRAM to SOCK_STREAM](https://stackoverflow.com/questions/13953912/difference-between-unix-domain-stream-and-datagram-sockets) as HAProxy does, for example, for their stats socket.

/cc @MarSik

tuned/exports/unix_socket_exporter.py
```diff
@@ -0,0 +1,228 @@
...
+		Sends data via socket s to address.
+		First byte of message specifies type of message (currently "r" for response, "e" for error message and "s" for signal).
+		Four more bytes are integer in network/big-endian with length of rest of message.
+		Rest of message is json data.
+		"""
+		log.debug("Sending socket data (type: %s, data: %s) to %s" % (type, data, address))
+		try:
+			j = json.dumps(data).encode("utf-8")
+			s.sendto(struct.pack("!sL", type.encode("utf-8"), len(j)), address)
+			s.sendto(j, address)
```
MarSik, line 152: Why two datagrams here? JSON does not need extra length, moreover recv is reading just one datagram per call and the buff size is the maximum received, extra data will be dropped (as this is datagram not stream).

So you really only need the json one.

tuned/exports/unix_socket_exporter.py
```diff
@@ -0,0 +1,228 @@
...
+				return
+			if data:
+				try:
+					log.debug(data.decode("utf-8"))
+					data = json.loads(data.decode("utf-8"))
+				except Exception as e:
+					log.error("Failed to load json data '%s': %s" % (data, e))
+					self._send_data(self._socket_object, address, "err", "Failed to load json data '%s': %s" % (data, e))
+					return
+				if type(data) not in (tuple, list) or len(data) < 1:
```
MarSik, line 215: The protocol must be a bit more complex than this to be fully reliable. The client has to provide an identifier (random or whatever) with every call to be able to identify the correct answer since it is datagram based.

I suggest you take a look at existing standards like json-rpc here: https://en.wikipedia.org/wiki/JSON-RPC#Version_2.0

It is simple enough and has all the necessary structure to support future development.

tuned/exports/unix_socket_exporter.py
```diff
@@ -0,0 +1,228 @@
...
+					log.error("Wrong format of call")
+					self._send_data(self._socket_object, address, "err", "Wrong format of call")
+				elif data[0] not in self._unix_socket_methods:
```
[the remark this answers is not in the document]
MarSik: Multiple methods are not really needed at long as there is a single method that does this :) But you are right that we will typically do a hotunplug / hotplug together (moving a CPU from one section to another).

tuned/exports/unix_socket_exporter.py
```diff
@@ -0,0 +1,228 @@
...
+		if r:
+			try:
+				l, address = self._socket_object.recvfrom(BYTES_LEN_OF_MSG)
+				l = struct.unpack("!L", l)[0]
+				log.debug("%s %s" % (address, l))
+			except Exception as e:
+				log.error("Failed to load length of message: %s" % e)
+				return
+			try:
+				data, address = self._socket_object.recvfrom(l)
```
MarSik, line 199: Just use a typical max datagram size: 4096. It will read less when the datagram is smaller.

tuned/exports/unix_socket_exporter.py
MarSik: See here for an example: https://gist.github.com/MarSik/e68acb69611d23939c835bc2dd6aa85d

jmencak:
Thank you for the latest commit, Honzo!  This is *great* progress not having to use the payload length and just:

```
printf '["active_profile"]' | nc -U /run/tuned/tuned.sock  # or
printf '["active_profile"]' | socat - UNIX-CONNECT:/run/tuned/tuned.sock
```

The default `socat` call often fails to output any result (with `WARNING  tuned.exports.unix_socket_exporter: Failed send data '<profile-name>' type 'res': [Errno 32] Broken pipe` in TuneD logs).  This is due to client disconnects, where `nc` seems to wait (longer?) for a response, whereas the default 0.5s timeout for `socat` is just not sufficient for this server-side implementation.

```
printf '["active_profile"]' | socat -t5 - UNIX-CONNECT:/run/tuned/tuned.sock
```

seems to work every time.

Testing against `HAProxy` implementation I had no such issues.  So at this point I'm wondering whether:

- it is worth/possible/practical improving the server-side (TuneD) implementation to improve the client (user) experience
- we should just document this

I think I'm probably fine with the latter.  @MarSik , thoughts?

tuned/exports/unix_socket_exporter.py
```diff
@@ -0,0 +1,195 @@
...
+		
+		self._unix_socket_methods[method_name] = wrapper(out_signature)
+		self._signals.add(method_name)
+
+	def send_signal(self, signal, *args, **kwargs):
+		if not signal in self._signals:
+			raise Exception("Signal '%s' doesn't exist." % signal)
+		for p in self._socket_signal_paths:
+			try:
+				s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
```
MarSik, line 105: Why the change to STREAM? STREAM would have been perfect for a bidirectional channel, but you still use two separate unix sockets here.

MarSik: Oh, you did, just not for signals it seems.

tuned/exports/unix_socket_exporter.py
```diff
@@ -0,0 +1,195 @@
...
+		self._construct_socket_object()
+
+	def stop(self):
+		if self._socket_object:
+			self._socket_object.close()
+
+	def _send_data(self, s, type, data):
+		"""
+		Sends data via socket s.
+		First byte of message specifies type of message (currently "r" for response, "e" for error message and "s" for signal).
```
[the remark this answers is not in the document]
MarSik: Btw, the code does not match the documentation. You send "err" and "res" here https://github.com/redhat-performance/tuned/pull/470/files#diff-1b8f90567f86039b1dfdcc66a8857985ba22cf3b129e2a8938c8805c6441bf71R191

jmencak:
Reporting a traceback during testing:
```
  File "/usr/sbin/tuned", line 98, in <module>
    app.run(args.daemon)
  File "/usr/lib/python3.6/site-packages/tuned/daemon/application.py", line 214, in run
    result = self._controller.run()
  File "/usr/lib/python3.6/site-packages/tuned/daemon/controller.py", line 65, in run
    exports.period_check()
  File "/usr/lib/python3.6/site-packages/tuned/exports/__init__.py", line 46, in period_check
    return ctl.period_check()
  File "/usr/lib/python3.6/site-packages/tuned/exports/controller.py", line 54, in period_check
    exporter.period_check()
  File "/usr/lib/python3.6/site-packages/tuned/exports/unix_socket_exporter.py", line 162, in period_check
    r, _, _ = select.select([self._socket_object], (), (), 0)
TypeError: argument must be an int, or have a fileno() method.
```
So far I've seen it only once

tuned/exports/unix_socket_exporter.py
```diff
@@ -0,0 +1,195 @@
...
+		Rest of message is json data.
+		"""
+		log.debug("Sending socket data (type: %s, data: %s)" % (type, data))
+		try:
+			s.send(type[0].encode("utf-8") + json.dumps(data).encode("utf-8"))
+		except Exception as e:
+			log.warning("Failed send data '%s' type '%s': %s" % (data, type, e))
```
jmencak, line 149: Nit: I belive the correct English would be "Failed *to* send data"

tuned/exports/unix_socket_exporter.py
```diff
@@ -0,0 +1,195 @@
...
+		printf '["switch_profile", "balanced"]' | sudo nc -U /run/tuned/tuned.sock
+		"""
+		r, _, _ = select.select([self._socket_object], (), (), 0)
+		if r:
+			conn, _ = self._socket_object.accept()
+			try:
+				data = ""
+				while True:
+					rec_data = conn.recv(4096).decode()
+					if not rec_data:
```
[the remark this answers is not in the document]
jmencak: Yeah, unfortunately, this doesn't scale.  Even this fails for some of the requests:

```
send() {
  local data="$1"
  printf "$data" | nc -U /run/tuned/tuned.sock
}

for i in $(seq 1 10)
do
  send '["active_profile"]' &
done
```

CZerta:
> Thank you for the latest commit, Honzo! This is _great_ progress not having to use the payload length and just:
> 
> ```
> printf '["active_profile"]' | nc -U /run/tuned/tuned.sock  # or
> printf '["active_profile"]' | socat - UNIX-CONNECT:/run/tuned/tuned.sock
> ```
> 
> The default `socat` call often fails to output any result (with `WARNING tuned.exports.unix_socket_exporter: Failed send data '<profile-name>' type 'res': [Errno 32] Broken pipe` in TuneD logs). This is due to client disconnects, where `nc` seems to wait (longer?) for a response, whereas the default 0.5s timeout for `socat` is just not sufficient for this server-side implementation.
> 
> ```
> printf '["active_profile"]' | socat -t5 - UNIX-CONNECT:/run/tuned/tuned.sock
> ```
> 
> seems to work every time.
> 
> Testing against `HAProxy` implementation I had no such issues. So at this point I'm wondering whether:
> 
>     * it is worth/possible/practical improving the server-side (TuneD) implementation to improve the client (user) experience
> 
>     * we should just document this
> 
> 
> I think I'm probably fine with the latter. @MarSik , thoughts?

Can add different logging, but I don't think socket allow to check if other side is still connected.

tuned/exports/unix_socket_exporter.py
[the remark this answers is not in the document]
CZerta: If it would be more comfortable for you, it's small change to pack it to something like 
`{"type": "error", "result": ...}`
It'd would be extendable, if we'd need to change it in future, so I'm pro.
For second one. there's `type[0]` on line 147.

tuned/exports/unix_socket_exporter.py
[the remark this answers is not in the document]
CZerta: Can add this const to config so you can set how many connections to listen to. Can add to hold connection too, but I don't think it would fix this scenario

CZerta:
> So the good news is that after the latest commit I can no longer see the trace above and that the requests are being served reliably. However, one per second. Is this expected?

You're right, we can do better.

jmencak:
> ... we can do better.

To give you some sense of what is expected.  On a 112 CPU r750, with the default 50 QPS I created 100 pods in parallel.  They were all PodScheduled within 1s delta and the average "startup" time between PodScheduled and Ready was 5.19s .  I may try tweaking the QPS, but I doubt it will have much significance in reducing the startup times.  I used CPUManager static policy and all containers were assigned to its own CPU.

tuned/exports/unix_socket_exporter.py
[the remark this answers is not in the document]
jmencak: Coming back to this subject as I believe API is probably the most important to get right.  Code can be fixed afterwards, API not so easily...

I suggest the following (yaml is easier to read, please convert to json on your own):
```
version: 0.0.1
methods:
- args:
  - arg1
  - arg2
  - arg3
  command: method1
- args:
  - arg1
  - arg2
  - arg3
  command: method2
```

or even simpler

```
version: 0.0.1
methods:
- command:
  - name
  - arg1
  - arg2
- command:
  - name
  - arg1
  - arg2
```

We can even drop `version` and add it later if/when needed, but as the `json` is structured now, it will be impossible to change/enhance in the future.  I'm not even suggesting to implement more than one method.  That can even be done later if it is too difficult to implement (would be nice though), but I belive the json structure has to change to be more flexible.  Thoughts? @MarSik , @yarda  I'm not good with naming things, feel free to suggest better names and structure.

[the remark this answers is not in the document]
MarSik: Please please do not reinvent the wheel. I mentioned it before: jsonrpc supports batches - https://www.jsonrpc.org/specification#batch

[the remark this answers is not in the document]
CZerta: @MarSik Do you want to implement similar structure to jsonrpc or is it ok as it's now? I f you have any suggestions, we can discuss it on other "faster" channels like IRC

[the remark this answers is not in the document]
jmencak: I suggest the `#tuned` IRC channel if/when Martin has time.  IMO the current implementation is not OK as it does not allow future enhancements easily enough.

tuned/exports/unix_socket_exporter.py
```diff
@@ -0,0 +1,246 @@
...
+		return self._check_id(self._create_result_response(ret, id))
+
+	def period_check(self):
+		"""
+		Periodically checks socket object for new calls. This allows to function without special thread.
+		Interface is according JSON-RPC 2.0 Specification (see https://www.jsonrpc.org/specification)
+		
+		Example calls:
+		
+		printf '[{"jsonrpc": "2.0", "method": "active_profile", "id": 1}, 1]' | sudo nc -U /run/tuned/tuned.sock
```
jmencak, line 199: I believe we can provide a better example combining the calls.  This one returns "Invalid request" and might confuse people.

Perhaps this one?

```
'[{"jsonrpc": "2.0", "method": "active_profile", "id": 1}, {"jsonrpc": "2.0", "method": "profiles", "id": 2}]'
```

jmencak (commented):
Thank you for the changes, Honzo!  I was able to test them with success, however, it needs a bit more work:

```
printf "DoS" | nc --send-only -U /run/tuned/tuned.sock
```

```
  File "/usr/lib/python3.10/site-packages/tuned/exports/unix_socket_exporter.py", line 220, in period_check
    data = json.loads(data)
  File "/usr/lib64/python3.10/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.10/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib64/python3.10/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.10/site-packages/tuned/exports/unix_socket_exporter.py", line 140, in _send_data
    s.send(json.dumps(data).encode("utf-8"))
BrokenPipeError: [Errno 32] Broken pipe

During handling of the above exception, another exception occurred:

  File "/usr/sbin/tuned", line 98, in <module>
    app.run(args.daemon)
  File "/usr/lib/python3.10/site-packages/tuned/daemon/application.py", line 215, in run
    result = self._controller.run()
  File "/usr/lib/python3.10/site-packages/tuned/daemon/controller.py", line 65, in run
    exports.period_check()
  File "/usr/lib/python3.10/site-packages/tuned/exports/__init__.py", line 46, in period_check
    return ctl.period_check()
  File "/usr/lib/python3.10/site-packages/tuned/exports/controller.py", line 54, in period_check
    exporter.period_check()
  File "/usr/lib/python3.10/site-packages/tuned/exports/unix_socket_exporter.py", line 223, in period_check
    self._send_data(conn, self._create_error_responce(-32700, "Parse error", str(e)))
  File "/usr/lib/python3.10/site-packages/tuned/exports/unix_socket_exporter.py", line 142, in _send_data
    log.warning("Failed to send data '%s' type '%s': %s" % (data, e))
TypeError: not enough arguments for format string

```

tuned/exports/unix_socket_exporter.py
CZerta: I wanted to provide example when call fails, but you're right it'd be unnecessarily confusing

tuned-main.conf
```diff
@@ -49,3 +49,27 @@ log_file_max_size = 1MB
...
+# Default unix socket ownership
+# Can be set as id or name, -1 or non-existing name leaves unchanged
+# unix_socket_ownership = -1 -1
+
+# Permissions for listening sockets
+# unix_socket_permissions = 0o600
+
+# Size of connections backlog for listen function on socket
+# Higher value allows to process requests from more clients
+# connections_backlog = 16
```
jmencak, line 75: After [reading on](https://arthurchiao.art/blog/the-mysterious-container-somaxconn/) and playing with this I suggest to bump this to 1024.  Even 128 is just way too low given how infrequently new connections are processed by `TuneD`.  I'd also advise to add a note about the fact the actual value is capped at `/proc/sys/net/core/somaxconn`.  See the manual page for `listen`.

tuned-main.conf
```diff
@@ -49,3 +49,27 @@ log_file_max_size = 1MB
 # It can be used to force tuning for specific architecture.
 # If commented, "/proc/cpuinfo" will be read to fill its content.
 # cpuinfo_string = Intel
+
+# Enable TuneD listening on dbus
+# enable_dbus = 1
+
+# Enable TuneD listening on unix domain socket
+# enable_unix_socket = 1
```
yarda, line 57: Please change to disabled by default.

yarda (approved):

PR #471, opened 2022-11-01
add profile name characters check

uniontech-lilinjie:
add profile name characters check
Exclusion file name is: '  ','@','#' ......

### DIFF
--- a/tuned-gui.py
+++ b/tuned-gui.py
@@ -50,6 +50,7 @@ import os
 import time
 import collections
 import subprocess
+import re
 
 import tuned.logs
 import tuned.consts as consts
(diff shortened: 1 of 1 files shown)


tuned-gui.py
```diff
@@ -508,6 +509,9 @@ def on_click_button_confirm_profile_update(self, data):
 
 	def data_to_profile_config(self):
 		name = self._gobj('entryProfileName').get_text()
+		if re.match("[A-Za-z0-9_]+", name) is None:
```
yarda, line 512: If I didn't miss anything I think the check is not correct, e.g. the following names also passes through:
profile%test
profile^123

Is it intended? If not,  maybe the regex should rather be "[A-Za-z0-9_]+$"?

Btw is there any reason for limiting the allowed characters? I think anything what's possible to get into the filesystem should be supported. If some special filesystem is used and it supports only subset of the characters it should return error that the profile cannot be created. E.g. I think that the 'profile-test' is also nice name somebody could use.

uniontech-lilinjie: Space characters can be created as file names, so add this limit

PR #472, opened 2022-11-02
fix 'is_active' does not work

uniontech-lilinjie:
fix 'is_active' does not work
tuned-gui: Delete --- the profile can still be deleted when edited, add self.editing_profile_name check to fix it.


yarda:
Could you please squash it into one commit?

PR #473, opened 2022-11-03
bootloader: create bootcmdline even when skip_grub_config=true

adriaan42:
Even when we instruct TuneD to not apply changes to the bootloader config (skip_grub_config=true) it's useful to expose the generated cmdline. External tools could then read TuneD's desired cmdline, and adapt configurations of (so far) unsupported bootloaders like systemd-boot.


yarda:
LGTM, thanks.

Maybe we should add method for cmdline generation not to repeat the same code .

PR #478, opened 2022-11-11
Allow selecting a different pkg-config executable

heirecka:
Helpful if you have an arch-prefixed pkg-config, e.g. when cross compiling.

Signed-off-by: Heiko Becker <heirecka@exherbo.org>


yarda:
LGTM, thanks.

PR #479, opened 2022-11-14
Adding support for cpu intel_pstate scaling driver

CZerta:
Resolves: rhbz#2095829

### DIFF
--- a/tuned/consts.py
+++ b/tuned/consts.py
@@ -100,6 +100,7 @@ CFG_LOG_FILE_COUNT = "log_file_count"
 CFG_LOG_FILE_MAX_SIZE = "log_file_max_size"
 CFG_UNAME_STRING = "uname_string"
 CFG_CPUINFO_STRING = "cpuinfo_string"
+CFG_CPU_EPP_FLAG = "hwp_epp"
 
 # no_daemon mode
 CFG_DEF_DAEMON = True
--- a/tuned/plugins/plugin_cpu.py
+++ b/tuned/plugins/plugin_cpu.py
@@ -65,7 +65,8 @@ class CPULatencyPlugin(base.Plugin):
 	Bias (EPB) values are supported. The alternative values are separated
 	using the '|' character. The following EPB values are supported
 	starting with kernel 4.13: "performance", "balance-performance",
-	"normal", "balance-power" and "power".
+	"normal", "balance-power" and "power". On newer processors is value
+	writen straight to file (see rhbz#2095829)
 	+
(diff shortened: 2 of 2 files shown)


joemario (commented):
Hi Jan:
A couple of comments:

DIR=/sys/devices/system/cpu/cpufreq/policyX
If the $DIR/energy_performance_available_preferences file exists, then so will the $DIR/energy_performance_preference file.   

The $DIR/energy_performance_available_preferences  file is checked to see if it contains the value that the user requested.  If it does, then write that value to the $DIR/energy_performance_preference file.

I suspect you need both a
  > def intel_energy_performance_available_preferences(...)
and a
  > def intel_energy_performance_preference(...)

You might want to remove the "pstate" from the "_intel_pstate_preference_path name(s).  That's because with the upcoming kernel patch discussed in the BZ should make the energy_performance_* files available whether intel_pstate is enabled or not.  

The comments in this section of new code (lines 539-555) mention "energy_perf_bias".  Unless I'm missing something, those comments should be mentioning energy_performance_preference.

CZerta:
@joemario 

> 
> I suspect you need both a
> 
> > def intel_energy_performance_available_preferences(...)
> > and a
> > def intel_energy_performance_preference(...)

The method _intel_preference_path is used for both of them depanding on available argument, so I don't need to hard code two basically identical path.

> You might want to remove the "pstate" from the "_intel_pstate_preference_path name(s). That's because with the upcoming kernel patch discussed in the BZ should make the energy_performance_* files available whether intel_pstate is enabled or not.

Removed.

> The comments in this section of new code (lines 539-555) mention "energy_perf_bias". Unless I'm missing something, those comments should be mentioning energy_performance_preference.

This is because it is set through "energy_perf_bias" option. We can add new one I don't think it makes sense as it is setting same thing and renaming would break already written profiles.

joemario:
Jan:
Before you complete this work:
1) We need to have @prarit  respond with confirmation that the algorithm he said  TuneD needs to use is indeed valid.  

2) The perf team and Jirka Hladly's team needs to finish its perf testing to show that not setting energy_perf_bias (e.g. to 0) and letting it remain with the kernel default of 6, does not hurt performance.  We've done a lot of testing so far which shows it will be fine to make this change.  But hold off for a final ack from us.

Thank you.
Joe

prarit:
The algorithm described in the BZ is correct.

CZerta:
@joemario Hi. I pushed changed version. Can you check if it's like you expected it?

PR #480, opened 2022-11-15
fix duplicate profile set

uniontech-lilinjie:
fix duplicate profile set
The same profile file can be written up to /etc/tuned/active_profile, fix it

### DIFF
--- a/tuned/admin/admin.py
+++ b/tuned/admin/admin.py
@@ -279,7 +279,7 @@ class Admin(object):
 	def _action_dbus_profile(self, profiles):
 		if len(profiles) == 0:
 			return self._action_dbus_list()
-		profile_name = " ".join(profiles)
+		profile_name = " ".join(set(profiles))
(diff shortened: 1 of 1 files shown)


yarda:
Is that really bug?

Let's say there is:
```
tuned-adm profile profile1 profile2 profile3 profile1
```

It works (or should work) the way:
- profile1 is merged with profile2, conflicting values are rewritten from profile2, we got profile12
- profile12 is merged with profile3, conflicting values are rewritten from profile3, we got profile123
- profile123 is merged with profile1, i.e. conflicting values from profile1 wins

The first profile1 is probably redundant now, but the order matters.

In case some plugin instance in the batch has different merge strategy I think that the first profile1 may not be even redundant in such case.

uniontech-lilinjie:
It only needs to be excluded:
 tuned-adm profile profile1 profile1
this pull request need optimized

PR #481, opened 2022-11-22
Add new cpu-partitioning-powersave profile

christophefontaine:
The new profile sets the intel_acpi driver to passive mode to be able to set the standard acpi governors (ondemand/userspace) and provides more flexibility on the C-states, as it is exposed as a new variable.

It relies on the existing cpu-partitioning profile ; this patch is required to enable further power optimisations in OVS-DPDK and OpenStack.


christophefontaine:
Forgot to include the profile to the proper rpm, I'll create a new PR soon.

PR #482, opened 2022-11-22
Add new cpu-partitioning-powersave profile

christophefontaine:
The new profile sets the intel_acpi driver to passive mode to be able to set the standard acpi governors (ondemand/userspace) and provides more flexibility on the C-states, as it is exposed as a new variable.


yarda:
LGTM.

Maybe in the future we could parametrize the cpu-partitioning profile.

PR #483, opened 2022-11-25
Fix for catching exceptions in profile loader.

CZerta:
This was caused by wrong inheritance of exceptions.

Resolves: rhbz#2071418


yarda (approved):
Thanks, LGTM, it fixes the traceback.

Maybe the original reporter will need the profile to be loaded with the warning. I asked him in https://bugzilla.redhat.com/show_bug.cgi?id=2071418#c3. But in such case let's handle it by separate PRs.

PR #484, opened 2022-11-28
configparser: use no strict parser to mimic old behavior

yarda:
Now it should allow duplicate options and sections in the INI files, the latest value is taken, the previous values are ignored, e.g.:

[selinux]
avc_cache_threshold=4096
avc_cache_threshold=8192

The avc_cache_threshold will be set to the 8192.

Related: rhbz#2071418

### DIFF
--- a/tuned/gtk/gui_plugin_loader.py
+++ b/tuned/gtk/gui_plugin_loader.py
@@ -76,7 +76,7 @@ class GuiPluginLoader():
         """
 
         try:
-            config_parser = ConfigParser(delimiters=('='), inline_comment_prefixes=('#'))
+            config_parser = ConfigParser(delimiters=('='), inline_comment_prefixes=('#'), strict=False)
             config_parser.optionxform = str
             with open(file_name) as f:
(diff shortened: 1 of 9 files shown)


yarda:
This doesn't handle warnings in case of a duplicate. This would require extension of the ConfigParser (IMHO it is also doable).

yarda:
It allows duplicates in all INI files.

jmencak:
Thank you for the PR, tested on RHEL-8.6 and Fedora 36.  No warning on duplicate lines is issued, but I guess we can live with that.

CZerta (commented):
Shouldn't it be added [here](https://github.com/redhat-performance/tuned/blob/ffc2725dca284b5d38efe5090bc590a31d7d6c2d/tuned/gtk/gui_profile_loader.py#L87) if we want to be consistent?

yarda:
> Shouldn't it be added [here](https://github.com/redhat-performance/tuned/blob/ffc2725dca284b5d38efe5090bc590a31d7d6c2d/tuned/gtk/gui_profile_loader.py#L87) if we want to be consistent?

You are right, I think it should be there.

CZerta (approved):
LGTM

PR #489, opened 2023-01-12
D-Bus: only send tracebacks through the D-Bus if in the debug mode

yarda:
Resolves: rhbz#2159680


CZerta (approved):
LGTM

PR #492, opened 2023-02-07
Inform users about reapplied sysctls

jmencak:
When the `reapply_sysctl` option in `tuned-main.conf` is set to 1, "system" sysctls from `/run/sysctl.d/*.conf`, `/etc/sysctl.d/*.conf` and `/etc/sysctl.conf` will be set *after* TuneD sysctl settings are applied. As `reapply_sysctl=1` is TuneD default, this can be confusing to TuneD users who only rely on TuneD profile settings and do not know about this functionality.  We also need a way to inform other components using TuneD about this.

This PR adds information about reapplied sysctls in TuneD logs.


jmencak:
/cc @yarda 
Please let me know if rhbz is needed for this.

yarda:
Thanks, LGTM, BZ is not needed.

PR #493, opened 2023-02-08
add 'tuned-adm delete' command

uniontech-lilinjie:
add 'tuned-adm delete' command
Convenient for users to delete their own generated profile. (dir: /etc/tuned/)

### DIFF
--- a/tuned/daemon/controller.py
+++ b/tuned/daemon/controller.py
@@ -5,6 +5,8 @@ from tuned.exceptions import TunedException
 import threading
 import tuned.consts as consts
 from tuned.utils.commands import commands
+import os
+import shutil
 
 __all__ = ["Controller"]
 
@@ -251,6 +253,27 @@ class Controller(tuned.exports.interfaces.ExportableInterface):
 		if profile_name is None or profile_name == "":
 			profile_name = self.active_profile()
 		return tuple(self._daemon.profile_loader.profile_locator.get_profile_attrs(profile_name, [consts.PROFILE_ATTR_SUMMARY, consts.PROFILE_ATTR_DESCRIPTION], [""]))
+
+	@exports.export("s", "b")
+	def profile_delete(self, profile, caller = None):
+		if caller == "":
+			return ""
+		if profile == "":
+			print("No set profile to delete.")
+			return False
+		try:
+			profile_list = os.listdir(consts.USER_PROFILE_DIR)
+			if profile not in profile_list:
+				print("File '%s' does not exists." % profile)
+				return False
+			else:
+				shutil.rmtree(os.path.join(consts.USER_PROFILE_DIR, profile))
+				print("Profile '%s' delete successfully" % profile)
+
+		except TunedException as e:
+			self._error(str(e))
+			return False
+		return True
 
 	@exports.export("", "s")
 	def recommend_profile(self, caller = None):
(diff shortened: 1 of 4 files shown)


tuned/daemon/controller.py
```diff
@@ -252,6 +254,27 @@ def profile_info(self, profile_name, caller = None):
 			profile_name = self.active_profile()
 		return tuple(self._daemon.profile_loader.profile_locator.get_profile_attrs(profile_name, [consts.PROFILE_ATTR_SUMMARY, consts.PROFILE_ATTR_DESCRIPTION], [""]))
 
+	@exports.export("s", "b")
+	def profile_delete(self, profile, caller = None):
```
yarda, line 258: I think it may be better to name it "user_profile_delete" because it operates only on user profiles.

tuned/daemon/controller.py
```diff
@@ -252,6 +254,27 @@ def profile_info(self, profile_name, caller = None):
...
+		if caller == "":
+			return ""
+		if profile == "":
+			print("No set profile to delete.")
```
yarda, line 262: It should log.* the messages instead of the print.

yarda: Maybe it can also return the error message on the D-bus.

tuned/daemon/controller.py
```diff
@@ -252,6 +254,27 @@ def profile_info(self, profile_name, caller = None):
...
+			return False
+		try:
+			profile_list = os.listdir(consts.USER_PROFILE_DIR)
+			if profile not in profile_list:
+				print("File '%s' does not exists." % profile)
+				return False
+			else:
+				shutil.rmtree(os.path.join(consts.USER_PROFILE_DIR, profile))
```
yarda, line 270: I would probably add some escape function sanitizing the final path.

I.e. escape special characters, norm path, check that the abs path is still under USER_PROFILE_DIR and check the the profiledir has tuned.conf, i.e. whether it's valid profile.

tuned/daemon/controller.py
```diff
@@ -252,6 +254,27 @@ def profile_info(self, profile_name, caller = None):
 			profile_name = self.active_profile()
 		return tuple(self._daemon.profile_loader.profile_locator.get_profile_attrs(profile_name, [consts.PROFILE_ATTR_SUMMARY, consts.PROFILE_ATTR_DESCRIPTION], [""]))
 
+	@exports.export("s", "b")
+	def profile_delete(self, profile, caller = None):
+		if caller == "":
+			return ""
+		if profile == "":
```
yarda, line 261: Maybe you could call the _action_profile_delete not to duplicate the code?

yarda (changes requested):
I think it should check whether the profile is not in the active profiles because it would cause problems after reboot.

You should also add the method to the polkit policy, file com.redhat.tuned.policy. It should be allowed only for the root for security reasons.

Finally, I am not sure whether it's worth the effort. It could open attacker possible new vectors how to delete system files and cause DoS (in cause of successful exploitation). Is there any use case for the need of deletion of profiles through the D-Bus instead of deleting it by normal CLI commands? Because at the moment you also cannot create new profiles or edit existing through the D-Bus, only with the GTK GUI, thus all the profiles management code should probably go there.

PR #495, opened 2023-02-08
fix 'is_active' does not work

uniontech-lilinjie:
fix 'is_active' does not work
tuned-gui: Delete --- the profile can still be deleted when edited, add self.editing_profile_name check to fix it.

### DIFF
--- a/tuned-gui.py
+++ b/tuned-gui.py
@@ -346,6 +346,7 @@ class Base(object):
 
 	def on_delete_event(self, window, data):
 		window.hide()
+		self.editing_profile_name = None
 		return True
 
 	def _get_active_profile_name(self):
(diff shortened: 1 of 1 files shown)


tuned-gui.py
```diff
@@ -362,7 +363,7 @@ def execute_remove_profile(self, button):
 			if profile is None:
 				self.error_dialog('No profile selected!', '')
 				return
-			if self._gobj('windowProfileEditor').is_active():
+			if self.editing_profile_name is not None:
```
yarda, line 366: Why not just use:
```
-                       if self._gobj('windowProfileEditor').is_active():
+                       if self._gobj('windowProfileEditor').is_visible():
```

yarda (changes requested):
I think that the 'is_visible()' method should do it.

uniontech-lilinjie:
> 
yes, this way is simple, I will take a new pull request

PR #499, opened 2023-02-09
fix 'is_active' does not work

uniontech-lilinjie:
fix 'is_active' does not work


yarda:
LGTM, thanks.

Btw, you don't need to close the PR and open new one. It's OK to `git commit -a --amend` the changes and then force push your by `git push -f`. It has advantage that the history and discussion is not lost.

yarda (approved):

PR #501, opened 2023-02-10
Report reapplied sysctls only on different values

jmencak:
PR492 introduced reporting on reapplied sysctls.  While it is a useful functionality to get notified about potential conflicts between filesystem-level and TuneD-level settings, it can generate a lot of noise when these settings are equal.  Report reapplied sysctls only when filesystem-level and TuneD-level settings differ.


yarda:
LGTM, thanks.

PR #502, opened 2023-02-13
fix log error

liuzhilin12:
I will add some log errors.


yarda:
Thanks, LGTM. The C9S test failure seems unrelated.

PR #505, opened 2023-02-15
Better log on unsupported hw for pm_qos_resume_latency_us option

CZerta:
Related: rhbz#2118786

### DIFF
--- a/tuned/plugins/plugin_cpu.py
+++ b/tuned/plugins/plugin_cpu.py
@@ -196,6 +196,7 @@ class CPULatencyPlugin(hotplug.Plugin):
 		self._is_amd = False
 		self._has_energy_perf_bias = False
 		self._has_intel_pstate = False
+		self._has_pm_qos_resume_latency_us = None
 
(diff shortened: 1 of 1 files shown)


tuned/plugins/plugin_cpu.py
```diff
@@ -708,6 +716,8 @@ def _set_pm_qos_resume_latency_us(self, pm_qos_resume_latency_us, device, sim):
 				log.warning("Invalid latency specified: '%s'" % str(pm_qos_resume_latency_us))
 				return None
 		if not sim:
+			if not self._check_pm_qos_resume_latency_us(device):
+				return None
```
yarda, line 720: I think it should return `None` here even if `sim` is `True` otherwise it may break verification.

CZerta: You're probably right, we're not 100% consistent with this behavior across plugins.

yarda (approved):

yarda:
LGTM, thanks.

PR #507, opened 2023-02-15
tuned-adm: better error message for unauthorized switch_profile

yarda:
As a sideeffect of 08b348b7a4b1a5d94bea160fecea4a2759b01a9d we got strange error messages for unauthorized switch_profile requests. This commit tries to improve it.


yarda:
It also adds the "Unable to switch profile: " prefix to other switch profile errors, but it unifies the error messages and should be OK.

CZerta (approved):
LGTM

PR #508, opened 2023-02-18
install dbus policy in /usr/share/dbus-1

evgeni:
/etc is for local overrides, not for package-shipped policies

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,7 +191,7 @@ install: install-dirs
 	install -Dpm 0644 tuned.service $(DESTDIR)$(UNITDIR)/tuned.service
 
 	# dbus configuration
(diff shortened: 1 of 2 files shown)


evgeni:
From the dbus 1.14 changelog (https://gitlab.freedesktop.org/dbus/dbus/-/blob/master/NEWS):

```
  • Third-party software should install default dbus policies for the system
  bus into ${datadir}/dbus-1/system.d (this has been supported since dbus
  1.10, released in August 2015). Installing default dbus policies in
  ${sysconfdir}/dbus-1/system.d is now considered to be deprecated. Policy
  files in ${sysconfdir}/dbus-1/system.d continue to be read, but this
  directory should only be used by system administrators wishing to
  override the default policies.

  The ${datadir} applicable to dbus is usually /usr/share and the
  ${sysconfdir} is usually /etc.
```

yarda:
LGTM, thanks.

PR #511, opened 2023-02-20
Allow to skip channels parameter in realtime profiles just by changing variables config.

CZerta:
Related: rhbz#2148990


yarda (changes requested):
Maybe we shouldn't complicate it. What about dropping the function check_net_queue_count and adding its functionality to the plugin_net?

It could work the way if there is e.g. the special value 'auto', then calculate it. And also change the realtime profiles appropriately.

PR #516, opened 2023-03-14
add error dialog when no active profile set

uniontech-lilinjie:
Signed-off-by: lilinjie lilinjie@uniontech.com
add error dialog when no active profile set
&&
error_dialog() missing 1 required positional argument: 'info'


yarda:
Thanks, both fixes LGTM.

yarda:
Unfortunately, the PR broke consistency of tabs and spaces and the code refuses to run with python3, fixed by #548.

PR #518, opened 2023-03-22
Don't touch sysctls set in sysctl.conf

lkundrak:
TuneD reverting sysctls to values configured in sysctl.conf was reported to have made one fine gentleman very sad.

The individual commits contains more details & a proposed remedy.


lkundrak:
Marked as draft since it seems I need to fix tests first.

Also sorry about the s/tuned/teamd/ in branch name, it's due to brain rot from my dealings with teamd.

PR #523, opened 2023-03-29
tests/beakerlib: Add new test which covers socket API.

RHEmployee:


yarda:
@RHEmployee please squash commits into one.

RHEmployee:
Ok. Done.

PR #527, opened 2023-04-20
Added sap-hana-kvm-guest profile

newkit:
As described in this [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=2173740) the sap-hana-kvm-guest profile should be added to the tuned-profiles.

### DIFF
--- a/profiles/sap-hana-kvm-guest/tuned.conf
+++ b/profiles/sap-hana-kvm-guest/tuned.conf
@@ -1,0 +1,18 @@
+#
+# tuned configuration
+#
+[main]
+summary=Optimize for running SAP HANA on KVM inside a virtual guest
+include=sap-hana
+
+[modules]
+cpuidle-haltpoll=+r force
+
+[sysfs]
+priority=10
+/sys/devices/system/clocksource/clocksource0/current_clocksource=tsc
+/sys/module/haltpoll/parameters/guest_halt_poll_ns=2400000
+/sys/module/haltpoll/parameters/guest_halt_poll_grow_start=2400000
+
+[bootloader]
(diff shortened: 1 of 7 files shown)


profiles/sap-hana-kvm-guest/tuned.conf
```diff
@@ -0,0 +1,18 @@
+#
+# tuned configuration
+#
+[main]
+summary=Optimize for running SAP HANA on KVM inside a virtual guest
+include=sap-hana
+
+[haltpoll]
```
yarda, line 8: Maybe better to use the modules plugin, e.g.:
```
[modules]
cpuidle-haltpoll=+r force
```
The `+r` special flag tells TuneD to remove this module (if it is loaded) and reload it with the `force` parameter. In case the module is not loaded it's loaded with the `force` parameter.

And increase priority of the plugin instances which uses it, i.e. `[sysfs]` to be processed later:
```
[sysfs]
priority=10
....
```
Example e.g.:
https://github.com/redhat-performance/tuned/blob/master/profiles/balanced/tuned.conf

yarda (changes requested):
Also please:
1) reference the bugzilla in the commit
2) decide whether this profile will go to the profiles-sap-hana subpackage or to the basic TuneD package or to the new sub package
3) edit tuned.spec accordingly to the decision in 2)
4) add manual page or edit existing (tuned-profiles-sap-hana.7) according to the decision in the 2)

Or provide us the information and we can update the commit ourselves.

newkit:
Addressed the requested changes:

- it's part of profiles-sap-hana 
- Updated documentation

yarda:
Could you please squash commits into one and force push?

PR #532, opened 2023-05-16
Fix logging in [scheduler] plug-in

jmencak:
`s/log.err()/log.error()/`


yarda:
Thanks, LGTM, good catch. Interesting that it hasn't been yet found by static analysis.

PR #533, opened 2023-05-22
Add rollback_on_exit option

jmencak:
Add rollback option to tuned-main.conf file.  The option specifies how
TuneD should perform rollbacks.  By default, the old behaviour is
preserved (rollback=auto) and settings are rolled back on TuneD daemon
exit and profile switches.  Using rollback=not_on_exit will result in
TuneD not performing any rollbacks on the daemon exit, however, on
profile switches, settings rollbacks will still be performed.

Resolves: rhbz#2203142

### DIFF
--- a/tuned-main.conf
+++ b/tuned-main.conf
@@ -75,3 +75,10 @@ log_file_max_size = 1MB
 # Size of connections backlog for listen function on socket
 # Higher value allows to process requests from more clients
 # connections_backlog = 1024
+
+# TuneD daemon rollback strategy. Supported values: auto|not_on_exit
+# - auto: rollbacks are always performed on a profile switch or 
+#   graceful TuneD process exit
+# - not_on_exit: rollbacks are always performed on a profile
+#   switch, but not on any kind of TuneD process exit
+# rollback = auto
--- a/tuned/daemon/daemon.py
+++ b/tuned/daemon/daemon.py
@@ -21,12 +21,14 @@ class Daemon(object):
 		self._update_interval = int(consts.CFG_DEF_UPDATE_INTERVAL)
 		self._dynamic_tuning = consts.CFG_DEF_DYNAMIC_TUNING
 		self._recommend_command = True
+		self._rollback = consts.CFG_DEF_ROLLBACK
 		if config is not None:
 			self._daemon = config.get_bool(consts.CFG_DAEMON, consts.CFG_DEF_DAEMON)
 			self._sleep_interval = int(config.get(consts.CFG_SLEEP_INTERVAL, consts.CFG_DEF_SLEEP_INTERVAL))
 			self._update_interval = int(config.get(consts.CFG_UPDATE_INTERVAL, consts.CFG_DEF_UPDATE_INTERVAL))
 			self._dynamic_tuning = config.get_bool(consts.CFG_DYNAMIC_TUNING, consts.CFG_DEF_DYNAMIC_TUNING)
 			self._recommend_command = config.get_bool(consts.CFG_RECOMMEND_COMMAND, consts.CFG_DEF_RECOMMEND_COMMAND)
+			self._rollback = config.get(consts.CFG_ROLLBACK, consts.CFG_DEF_ROLLBACK)
 		self._application = application
 		if self._sleep_interval <= 0:
 			self._sleep_interval = int(consts.CFG_DEF_SLEEP_INTERVAL)
@@ -229,21 +231,29 @@ class Daemon(object):
 
 		# if terminating due to profile switch
 		if self._terminate_profile_switch.is_set():
-			full_rollback = True
+			full_rollback = consts.ROLLBACK_FULL
 		else:
-			# with systemd it detects system shutdown and in such case it doesn't perform
-			# full cleanup, if not shutting down it means that TuneD was explicitly
-			# stopped by user and in such case do full cleanup, without systemd never
-			# do full cleanup
-			full_rollback = False
-			if self._full_rollback_required():
+			# Assume only soft rollback is needed. Soft rollback means reverting all
+			# non-persistent tunings applied by a plugin instance. In contrast to full
+			# rollback, information about what to revert is kept in RAM (volatile
+			# memory) -- TuneD data structures.
+			# With systemd TuneD detects system shutdown and in such a case it doesn't
+			# perform full cleanup. If the system is not shutting down, it means that TuneD
+			# was explicitly stopped by the user and in such case do the full cleanup. On
+			# systems without systemd, full cleanup is never performed.
+			full_rollback = consts.ROLLBACK_SOFT
+			if not self._full_rollback_required():
+				log.info("terminating TuneD due to system shutdown / reboot")
+			elif self._rollback == "not_on_exit":
+				# no rollback on TuneD exit whatsoever
+				full_rollback = consts.ROLLBACK_NOT_ON_EXIT
+				log.info("terminating TuneD and not rolling back any changes due to '%s' option in '%s'" % (consts.CFG_ROLLBACK, consts.GLOBAL_CONFIG_FILE))
+			else:
 				if self._daemon:
 					log.info("terminating TuneD, rolling back all changes")
-					full_rollback = True
+					full_rollback = consts.ROLLBACK_FULL
 				else:
 					log.info("terminating TuneD in one-shot mode")
-			else:
-				log.info("terminating TuneD due to system shutdown / reboot")
 		if self._daemon:
 			self._unit_manager.stop_tuning(full_rollback)
 		self._unit_manager.destroy_all()
--- a/tuned/plugins/base.py
+++ b/tuned/plugins/base.py
@@ -212,7 +212,7 @@ class Plugin(object):
 	def _instance_post_static(self, instance, enabling):
 		pass
 
-	def _call_device_script(self, instance, script, op, devices, full_rollback = False):
+	def _call_device_script(self, instance, script, op, devices, full_rollback = consts.ROLLBACK_SOFT):
 		if script is None:
 			return None
 		if len(devices) == 0:
@@ -228,7 +228,7 @@ class Plugin(object):
 			environ = os.environ
 			environ.update(self._variables.get_env())
 			arguments = [op]
-			if full_rollback:
+			if full_rollback == consts.ROLLBACK_FULL:
 				arguments.append("full_rollback")
 			arguments.append(dev)
 			log.info("calling script '%s' with arguments '%s'" % (script, str(arguments)))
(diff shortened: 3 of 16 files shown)


jmencak:
So, this is very much a WiP.  In order to actually prevent the rollback/unapplication of tuning, I either need to modify `instance_unapply_tuning()` in `tuned/plugins/base.py` or `_instance_unapply_static()` in every plug-in.  Obviously, the former is preferable, but I'm a bit confused about what `full_rollback` is intended to do.  @yarda , can we use that variable or do we need a new one for what I'm trying to do?  It is likely I probably need to rename the `rollback_on_exit` to something like `unapply_tuning_on_exit` and use it in `instance_unapply_tuning()` in `tuned/plugins/base.py`, right?  Thanks!

tuned/daemon/daemon.py
```diff
@@ -177,14 +179,14 @@ def _notify_profile_changed(self, profile_names, result, errstr):
 			exports.send_signal(consts.SIGNAL_PROFILE_CHANGED, profile_names, result, errstr)
 		return errstr
 
-	def _full_rollback_required(self):
+	def _system_shutdown_or_reboot(self):
```
yarda, line 182: It also returns `True` if we don't know (or more precisely if `systemd` doesn't know or is misconfigured/in the failed state, etc.), thus maybe rename the method to "_system_shutdown_or_reboot_or_unknown" or similarly to clearly indicate the condition.

yarda (changes requested):

yarda:
Thanks. I am now running some tests, but it mostly LGTM.

I initially thought about optimization not to rollback settings that will be changed consequently, but it would require deeper TuneD engine changes. We could of course implement it later.

The proposed PR is more like hack, but it's simple, it shouldn't break anything and if it works for you, let's go with it for now.

tuned/daemon/daemon.py
jmencak: I wasn't able to come up with a function name for this I liked, so I reverted the function name to what it was originally and also reverted the logic in the fuction.

jmencak:
> Thanks. I am now running some tests, but it mostly LGTM.
> 
> I initially thought about optimization not to rollback settings that will be changed consequently, but it would require deeper TuneD engine changes. We could of course implement it later.
> 
> The proposed PR is more like hack, but it's simple, it shouldn't break anything and if it works for you, let's go with it for now.

Would the future rewrite take advantage of the new `rollback_on_exit` function?  If not, I wouldn't like to add tech debt to TuneD.  Also, let's have another pair of eyes on this (Martin) if he has time.

tuned/daemon/daemon.py
```diff
@@ -228,6 +230,7 @@ def _thread_code(self):
 			i += 1
 
 		# if terminating due to profile switch
```
MarSik, line 232: I think this need a bit more verbose comment, because the if right bellow says the same thing and sets full_rollback. The naming is a bit confusing here.

yarda:
> > Thanks. I am now running some tests, but it mostly LGTM.
> > I initially thought about optimization not to rollback settings that will be changed consequently, but it would require deeper TuneD engine changes. We could of course implement it later.
> > The proposed PR is more like hack, but it's simple, it shouldn't break anything and if it works for you, let's go with it for now.
> 
> Would the future rewrite take advantage of the new `rollback_on_exit` function? If not, I wouldn't like to add tech debt to TuneD. Also, let's have another pair of eyes on this (Martin) if he has time.

If the RFE would ever happen we will keep this functionality. I think it shouldn't break anything and there could be people who do not have any interest in the rollback feature, thus having the configuration switch for it makes perfect sense to me.

For the RFE the idea is to "preload" the new profile and check whether there is an intersection of some plugins/commands. If it is, just skip the rollback for such command and instead of reading the initial value in the constructor, reuse the original value. This should be straightforward for the device/non-device class commands. But it could be tricky for the custom class commands which could do anything imaginable.

tuned-main.conf
```diff
@@ -75,3 +75,6 @@ log_file_max_size = 1MB
 # Size of connections backlog for listen function on socket
 # Higher value allows to process requests from more clients
 # connections_backlog = 1024
+
+# Perform rollback on TuneD daemon exit
+# rollback_on_exit = 1
```
yarda, line 80: Maybe for future extensibility use `rollback` option name and string values e.g. `on_exit` and `auto` with the default value `auto`, e.g.:
```
rollback = on_exit
```
And we could add later more options like e.g.: `never`, `always`, etc.

We should also add the `tuned-main.conf` description to the upstream documentation, but this could be done later. It seems there is no dedicated section in the documentation yet.

jmencak: Thank you for the suggestion.  I believe we need "auto" and "not_on_exit" right now.  Code adjusted.

tuned/daemon/daemon.py
jmencak: Comments added.

yarda (approved):

yarda:
> LGTM, waiting for @MarSik comments.

Just two minor things I noticed:
- Maybe the `soft_rollback` is a bit confusing name, because IMHO with the `soft_rollback == False` no rollback would happen at all. Maybe you could indicate the no rollback with the `full_rollback = None` (quite hackish) or instead of the `full_rollback` and `soft_rollback` use just `rollback` with e.g. numbers (e.g. 0 - no rollback, 1 - normal rollback, 2 - full rollback - we shouldn't use the enum type, because we didn't drop python 2).
- Maybe add more verbose comment to the `tuned-main.conf`.

jmencak:
Hopefully, all comments are addressed now.  I've kept the `full_rollback` variable.  Perhaps, we should rename it to `rollback_type` or similar in the future or perhaps even in this PR if you feel it should be done now.  The semantics are as proposed by Jarda:

```
full_rollback: 0 # no rollback
full_rollback: 1 # normal rollback (settings kept in volatile memory/RAM by TuneD are reverted)
full_rollback: 2 # full rollback (persistent changes done on the filesystem by TuneD are reverted)
```

This is a more intrusive change than what I had before when I added `soft_rollback`. Hopefully, I did not miss anything.  I'll run NTO e2e tests with this and report back.  Running beaker tests with this would also help.

tuned/daemon/daemon.py
```diff
@@ -227,23 +229,32 @@
 		while not self._cmd.wait(self._not_used, self._sleep_interval) and i < 3:
 			i += 1
 
+		# Assume soft rollback is needed. Soft rollback means reverting all non-persistent
+		# tunings applied by a plugin instance. In contrast to full rollback, information
+		# about what to revert is kept in RAM (volatile memory) -- TuneD data structures.
+		full_rollback = 1
```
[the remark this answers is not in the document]
jmencak: Thank you, bot, fixed.

tuned-main.conf
```diff
@@ -75,3 +75,10 @@ log_file_max_size = 1MB
 # Size of connections backlog for listen function on socket
 # Higher value allows to process requests from more clients
 # connections_backlog = 1024
+
+# TuneD daemon rollback strategy. Supported values: auto|not_on_exit
+# - auto: rollbacks are always performed on a profile switch or 
+#   gracefull TuneD process exit
```
MarSik, line 81:
```suggestion
#   graceful TuneD process exit
```

tuned/plugins/base.py
```diff
@@ -298,10 +298,13 @@ def instance_update_tuning(self, instance):
 		if instance.has_dynamic_tuning and self._global_cfg.get(consts.CFG_DYNAMIC_TUNING, consts.CFG_DEF_DYNAMIC_TUNING):
 			self._run_for_each_device(instance, self._instance_update_dynamic, instance.processed_devices.copy())
 
-	def instance_unapply_tuning(self, instance, full_rollback = False):
+	def instance_unapply_tuning(self, instance, full_rollback = 1):
 		"""
 		Remove all tunings applied by the plugin instance.
 		"""
+		if full_rollback == 0:
```
MarSik, line 305: Will this break profile change? Because we want to make sure that sysctls mentioned in the profile being unloaded are reverted. The new profile might not be touching them.

MarSik: But since it is a parameter, it probably gets decided elsewhere and here it just selects between soft (in-memory) and full rollback, right?

jmencak: > Will this break profile change? 

No

> Because we want to make sure that sysctls mentioned in the profile being unloaded are reverted.

They will be.  During profile change, `stop_tuning()` will be called with "full_rollback == 2".  However, on `TuneD` process termination, it will be called with "full_rollback == 0" if the user specified "rollback = not_on_exit".

> But since it is a parameter, it probably gets decided elsewhere

Yes.

tuned-main.conf
jmencak: Thank you, fixed.

yarda:
> This is a more intrusive change than what I had before when I added `soft_rollback`. Hopefully, I did not miss anything. I'll run NTO e2e tests with this and report back. Running beaker tests with this would also help.

Yep, it is, but I think it's much cleaner now.

Only nit-picking, but maybe you could use consts and not immediate int literals for profile rollback mode, well it will be still C-ish approach, but I think the code would be cleaner.

jmencak:
> Only nit-picking, but maybe you could use consts and not immediate int literals for profile rollback mode, well it will be still C-ish approach, but I think the code would be cleaner.

Done.  As for my testing on OCP, some of the e2e tests are failing, but that' because they rely on the original rollback functionality.  I'll exclude them and rerun.

jmencak:
All relevant e2e NTO tests with "rollback = not_on_exit" are passing now.

yarda:
Well I think the variable `full_rollback` should be now called just `rollback` and if I am not mistaken the `consts.ROLLBACK_NOT_ON_EXIT` should be probably called `consts.ROLLBACK_NONE` or similarly because the upper logic decides if to do rollback or not and the underlying method doesn't care about the reason - it just get the argument don't do the rollback.

But it's just a cosmetic and I don't want to delay the release end enforce new re-testing, thus I will merge it as it is, but If you agree with my arguments please do another PR.

jmencak:
Thank you for accepting this PR.

> Well I think the variable `full_rollback` should be now called just `rollback`

I intentionally kept `full_rollback` for easier search and replace.  For easier future search and replace, would you be OK with renaming it to `rollback_type`?

> and if I am not mistaken the `consts.ROLLBACK_NOT_ON_EXIT` should be probably called `consts.ROLLBACK_NONE` or similarly because the upper logic decides if to do rollback or not and the underlying method doesn't care about the reason - it just get the argument don't do the rollback.

I'm fine with this renaming.

> But it's just a cosmetic and I don't want to delay the release end enforce new re-testing, thus I will merge it as it is, but If you agree with my arguments please do another PR.

I'll definitely do this once we agree on the naming.  Thank you.

PR #541, opened 2023-06-21
Fix TextView to expand with window

### LINKED ISSUE #540: TextBox on raw editor window, won't expand with window
image

rabin-io:
This should resolve #540


yarda:
Thansk, LGTM.

PR #542, opened 2023-06-28
No errors when physical_package_id file does not exist

jmencak:
Show no errors when the physical_package_id file does not exist -- the CPUs may be offline.

Resolves: rhbz#2217015


jmencak:
@yarda would that be the right approach for fixing rhbz#2217015 ?

yarda:
Maybe you could write debug message that the physical_package_id doesn't exist and was skipped.

jmencak:
> Maybe you could write debug message that the physical_package_id doesn't exist and was skipped.

Thank  you for the review. Done.

PR #545, opened 2023-08-04
test: FIX: add timeout when tuned starts.

RHEmployee:
We give tuned 5 seconds to start.


tests/beakerlib/Tuned-takes-too-long-to-reload-start-when-ulimit/runtest.sh
```diff
@@ -35,7 +35,7 @@ rlJournalStart
         rlRun "mv /etc/tuned/tuned-main.conf.new /etc/tuned/tuned-main.conf"
         rlRun "ulimit -H -n 1048576"
         rlRun "ulimit -S -n 1048576"
-        rlRun "tuned --debug 2>&1 | tee TEST_OUT"
+        rlRun "timeout 5 tuned --debug 2>&1 | tee TEST_OUT"
```
pemensik, line 38: By using ``| tee TEST_OUT`` you lose return status from timeout, the result would be wrong.

``false | tee /dev/null`` gives successful output. The timeout return code should be checked.

Use instead ``rlRun -s`` and ``$rlRun_LOG`` in following command.

pemensik (changes requested):
Do not use ``| tee`` to store result where beakerlib offers better way to do that.

tests/beakerlib/Tuned-takes-too-long-to-reload-start-when-ulimit/runtest.sh
RHEmployee: It is true that output complicate stuff and is not much needed for testing.
What about: just remove whole output check?

yarda: Why not just use `rlWatchdog`?

RHEmployee: Hello. Thank you for your feedback.
So you basically want to use rlWatchdog here or do you want to describe difference between timeout and rlWatchdog beakerlib call which can be found in documentation?

yarda: What's the reason here for the `timeout` instead of the `rlWatchdog`?

RHEmployee: According to documentation:
```
rlWatchdog

Run "command". If it does not finish in specified time, then kill it using "signal".

...

Signal to use (optional, default KILL).
```

vs

```
Start COMMAND, and kill it if still running after DURATION.

...

       Upon  timeout, send the TERM signal to COMMAND, if no other SIGNAL specified.  The TERM signal kills
       any process that does not block or catch that signal.  It may be necessary to use the  KILL  signal,
       since this signal can't be caught.
```

So do  you want me to use rlWatchdog here?

timeout sends TERM signal and I don't want to still watch rlWatchdog and remember that I need to change default signal from KILL to TERM. timeout have this for free and it's just pattern used in tests.

I would like to let tuned cleanup itself after TERM signal.

So do  you want me to use rlWatchdog here?

yarda: Are you kidding? You aren't using BeakerLib API but external command only because it would require specifying signal name after the rlWatchdog command? Moreover, because you never send SIGKILL in the test you could end up with stuck TuneD process.

> So do you want me to use rlWatchdog here?
I think you should use BeakerLib API whenever possible.

RHEmployee: Hello,

Thank you once again for providing your valuable feedback. It's truly wonderful to be a part of such an excellent and inclusive tuned community.

Could you kindly direct me to the documentation for upstream contributors of Tuned, specifically outlining the process of writing upstream tests with Beakerlib? Doing so would greatly help us expedite the review process.

I managed to locate documentation, but it appears to be outdated:
https://github.com/redhat-performance/tuned/blob/master/CONTRIBUTING.md

For instance, I noticed that some commits lack the Signed-off-by line, which suggests that either the documentation is old or the guidelines are not being followed consistently.

I would like to suggest deferring this Merge Request until we have a clearer understanding of how to properly create tests for Tuned upstream. Your assistance in this matter is greatly appreciated.

Thank you.

yarda: If you think that tests (BeakerLib test) contribution guideline will be useful, feel free to propose one. As a QA you should know best how to write such guide. This project started to use BeakerLib in upstream just recently and we don't have guidelines for everything.

Intent of the review is to have the best possible code. If we use BeakerLib we should use it's API. It's stable, tested, supported and it should work cross-system. Argument that BeakerLib API is not used because signal name will have to be specified (which would in the end add just only one more character in comparison to the external command) is odd.

If you insist on it, I will merge it, because it's you who is responsible for the tests. But I am pretty sure you have some downstream guide / best practices guide and preferring external commands over BeakerLib API is surely not recommended there.

pemensik: Btw. timeout command allows also ``--kill-after=timeout`` parameter, which can terminate process not responding to TERM after additional time by kill signal. IMO slightly preferred variant.

yarda: Yep, this is valid argument for the timeout command.

RHEmployee: I would like to express that I am currently experiencing a state of confusion. Could you kindly clarify whether the preferred approach is to utilize 'rlWatchdog' or the 'timeout' function with the '--kill-after' switch?

Furthermore, I am inclined to minimize any potential time wastage, and I am of the opinion that deferring any BeakerLib tests until the documentation is fully prepared could be a prudent course of action. I would greatly appreciate your insights and thoughts on this matter.

Thank you for your assistance.

RHEmployee: I have recently implemented a code update that incorporates a timeout. However, for the full integration of the 'rlwatchdog' functionality, it is imperative to introduce additional logic that accurately emulates the timeout behavior.

pemensik: What I find more important than rlWatchdog or timeout command usage is usage of ``| tee TEST_OUT``, which completely ignores the return code. It just tests whether current directory is writeable. Noted that already in https://github.com/redhat-performance/tuned/pull/545#discussion_r1284226161.

RHEmployee: In which recent commit can the '| tee' pattern's usage be observed?
Apologies, I'm unable to comprehend your message.

```
@@ -35,8 +35,8 @@ rlJournalStart
        rlRun "mv /etc/tuned/tuned-main.conf.new /etc/tuned/tuned-main.conf"
        rlRun "ulimit -H -n 1048576"
        rlRun "ulimit -S -n 1048576"
        rlRun "tuned --debug 2>&1 | tee TEST_OUT"
        rlAssertNotGrep "tuned.plugins.plugin_sysctl: executing \['sysctl'," TEST_OUT
	# According to bug, tuned should start within ~0.5 second
        rlRun "timeout --kill-after=10 5 tuned --debug"
    rlPhaseEnd

    rlPhaseStartCleanup
 ```
 
 Bug with using '| tee' were already fixed so why do we need to review it again?
 Or is there some cache/push issue?

RHEmployee: @pemensik I apologize for my unfamiliarity with this upstream process. Could you clarify whether you are a designated maintainer or contributor? I've been unable to locate any references to you in the git log messages or AUTHORS file.
Or how review process works here?

yarda: >@pemensik I apologize for my unfamiliarity with this upstream process. Could you clarify whether you are a designated maintainer or contributor? I've been unable to locate any references to you in the git log messages or AUTHORS file.
> Or how review process works here?

@RHEmployee maybe you should ask your manager to signup you to some course about opensource/fsf to be more acquainted with the opensource development model and meritocracy principles. TL/DR: every voice matter especially if it has useful feedback/comments and @pemensik comments are on topic and much more useful than yours, probably chatGPT generated, passive aggressively comments.

> I would like to express that I am currently experiencing a state of confusion. Could you kindly clarify whether the preferred approach is to utilize 'rlWatchdog' or the 'timeout' function with the '--kill-after' switch?

The preferred approach is to do it right and establishing that there will be no leftover process when the test finishes is probably the best way. Can `rlWatchdog` do the `--kill-after`? I don't think so. This seems to be the most efficient way how to do it. Should `rlWatchdog` be learned to know it? Very probably, but it's out of the scope of this PR.  Usually in cases when the BeakerLib API is on par with the external commands, the BeakerLib API is preferred, thus the solutions how I would prefer them (number 1 is the most preferred):
1. `timeout` with the `--kill-after`
2. `rlWatchdog`
3. the current solution

I can even accept the 3. if you insist on it, because I have much more important work to do and I don't want to waste more non-productive time on this.

RHEmployee:
Just note: additional merge commit need to be removed and branch rebased.
Looks like github doing just merge and not rebase.

PR #547, opened 2023-08-17
Serialize SIGHUP processing

yarda:
Resolves: rhbz#2215298


jmencak (approved):
Thank you for the changes, @yarda . They make sense to me and I also tested them on OpenShift with containerized TuneD where SIGHUP is being used.

LGTM

yarda:
@jmencak thanks for the quick review.

PR #549, opened 2023-08-17
gui: multiple fixes and warnings removal

yarda:


yarda:
It would be nice to have review for it, but taking into account that the GUI is very rarely used and this PR heavily improves things from the state "it doesn't work" to the state "it somehow works" and that we are currently out of reviewers, this one will go without review.

PR #555, opened 2023-09-14
RHEL-3707: Add support for sysctl names with slash

MarSik:
The logic for converting sysctl name to path did not work for sysctls like:

[sysctl]
net.ipv4.conf.bond0/355.forwarding=1

### DIFF
--- a/tuned/plugins/plugin_sysctl.py
+++ b/tuned/plugins/plugin_sysctl.py
@@ -157,7 +157,11 @@ class SysctlPlugin(base.Plugin):
 		self._write_sysctl(option, value, ignore_missing = True)
 
 	def _get_sysctl_path(self, option):
-		return "/proc/sys/%s" % option.replace(".", "/")
+		# The sysctl name in sysctl tool and in /proc/sys differs.
+		# All dots (.) in sysctl name are represented by /proc/sys
+		# directories and all slashes in the name (/) are converted
+		# to dots (.) in the /proc/sys filenames.
+		return "/proc/sys/%s" % self._cmd.tr(option, "./", "/.")
 
 	def _read_sysctl(self, option):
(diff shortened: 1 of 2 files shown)


yarda:
Hmm, it seems like a side-effect of the sysctl tool internal conversions, but we should follow its behaviour.

I don't like the triple replacement with internal escaping, what about the following maybe more python way which should be more "bullet-proof"?:
```
try:
  trans = str.maketrans("./", "/.")
except AttributeError:
  import string
  trans = string.maketrans("./", "/.")
return "/proc/sys/%s" % option.translate(trans)
```
The exception handler is there to also support python2 :) otherwise it's one-liner.

Or maybe we could add the 'tr' command to the the utils/commands.py doing all the magic and then just:
```
return "/proc/sys/%s" % self._cmd.tr(option, "./", "/.")
```

MarSik:
Translate works for me. I agree it is better, I simply did not know the function so thanks for teaching me something new :)

zacikpa:
@MarSik I pushed the change proposed by @yarda into your branch, hope you don't mind :)

zacikpa:
Sorry for the many force pushes :)

I updated the code and the commit message according to your comments @yarda. To be able to use `_cmd ` in the function, I had to make it (and others) a class method, so I created a separate commit for that. Let me know if you want that squashed as well.

yarda:
> Sorry for the many force pushes :)
> 
> I updated the code and the commit message according to your comments @yarda. To be able to use `_cmd ` in the function, I had to make it (and others) a class method, so I created a separate commit for that. Let me know if you want that squashed as well.

NP, thanks for the cleanup. I think here it makes sense to have two commits.

yarda:
LGTM, thanks.

PR #556, opened 2023-09-24
Return 0 when ksm services do not exist

jmencak:
When ksm or ksmtuned services do not exist, return 0 so that we do not generate ERRORs in TuneD logs.

Resolves: OCPBUGS-19656


jmencak:
Also see: https://redhat-internal.slack.com/archives/C04KZV0RD6Z/p1696258682294799

ramperher:
Hi @jmencak , I've updated OCPBUGS-19656 with a recurrent case we're having in our labs, it's 100% reproducible in any OCP 4.14 release, I suppose this change should fix the issue, but I don't know if it's expected to be merged soon or not.

PR #558, opened 2023-09-28
style: fix inconsistent whitespace in plugin_cpu.py

adriaan42:
Mixed indentation, tabs and spaces. Python does not mind, but Code/Pylance complains about it:

image


yarda:
Thanks.

PR #562, opened 2023-10-29
Allow equal characters in kernel module parameter value

gsauthof:
this fixes tuned verification when configuring module parameters such as:

    onload=cplane_server_params=--affinity=0x100000000

Tuned correctly sets them like this:

    $ cat /etc/modprobe.d/tuned.conf
    options onload cplane_server_params=--affinity=0x100000000
    $ cat /sys/module/onload/parameters/cplane_server_params
    --affinity=0x100000000

However, before this change tuned reports verification failure whenever a parameter value contains an equal sign.


yarda:
LGTM, thank for contribution.

PR #565, opened 2023-11-03
powersave: add EPP powersave

### LINKED ISSUE #563: `powersave` profile should set epp to `power` or `balance_power`
FYI: My system is an Asus ROG Zephyrus G14 GA402RJ with an AMD Ryzen 6900HS CPU

Hello, I'm new to using tuned and would like to suggest a change for better power savings.

When using the `powersave` profile in tuned, the energy_performance_preference defined in `/sys/devices/system/cpu/cpu*/cpufreq/energy_performance_preference` should be set to `balance_power` or for more strict configurations, `power`.

image

yarda:
Fixes #563

### DIFF
--- a/profiles/powersave/tuned.conf
+++ b/profiles/powersave/tuned.conf
@@ -8,6 +8,7 @@ summary=Optimize for low power consumption
 [cpu]
 governor=ondemand|powersave
 energy_perf_bias=powersave|power
+energy_performance_preference=powersave
 
 [eeepc_she]
 


ryanabx:
I think this is separate from #563 , though this is still a good change.

I think there's two different settings, one that you can set to powersave, and another that takes in arguments such as `performance`, `balance_performance`, `balance_power`, and `power`. 

Here's a reddit post describing `amd-pstate` and the epp driver: https://www.reddit.com/r/linux/comments/15p4bfs/amd_pstate_and_amd_pstate_epp_scaling_driver/

And this is a screenshot of the section `Configure amd_pstate_epp Energy Performance Preference`
image

zacikpa:
@ryanabx is correct. The changed profile will, in fact, fail to load, since `powersave` is not one of the valid options listed in `/sys/devices/system/cpu/cpufreq/policyX/energy_performance_available_preferences`.

yarda:
> @ryanabx is correct. The changed profile will, in fact, fail to load, since `powersave` is not one of the valid options listed in `/sys/devices/system/cpu/cpufreq/policyX/energy_performance_available_preferences`.

Yes, I realized it after the @ryanabx comment, but I haven't time to update the PR. I will try to do something about it.

PR #566, opened 2023-11-08
Add a tuned-adm command for moving devices between plugin instances

zacikpa:
Resolves: RHEL-15141

Until now, this was only possible by manually sending a message to TuneD via the DBus API.

### DIFF
--- a/man/tuned-adm.8
+++ b/man/tuned-adm.8
@@ -24,7 +24,7 @@
 tuned\-adm - command line tool for switching between different tuning profiles
 .SH SYNOPSIS
 .B tuned\-adm
-.RB [ list " | " active " | " "profile \fI[profile]\fP..." " | " "profile_info \fI[profile]\fP..." " | " off " | " auto_profile " | " profile_mode " | " "verify \fI[\-i | \-\-ignore\-missing]\fP" " | " recommend ]
+.RB [ list " | " active " | " "profile \fI[profile]\fP..." " | " "profile_info \fI[profile]\fP..." " | " off " | " auto_profile " | " profile_mode " | " "verify \fI[\-i | \-\-ignore\-missing]\fP" " | " recommend " | " "instance_acquire_devices \fIdevices\fP \fIinstance\fP" ]
 
 .SH DESCRIPTION
 This command line utility allows you to switch between user definable tuning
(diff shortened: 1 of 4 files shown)


man/tuned-adm.8
```diff
@@ -110,6 +110,16 @@ Enable automatic profile selection mode, switch to the recommended profile.
...
 Show current profile selection mode.
 
+.TP
+.B "instance_acquire_devices \fI[\-c | \-\-cpu\-list]\fP \fIdevices\fP \fIinstance\fP"
+Move all provided \fIdevices\fP under the given plugin \fIinstance\fP. The plugin
+instance is identified using its name from the TuneD profile.
+
+The devices are specified using a comma-separated list. When the \fI\-c\fP
+or the \fI\-\-cpu\-list\fP flag is enabled, the tool assumes that CPU devices
+are specified using a list of their numbers, such as "0,2-4", meaning that
```
yarda, line 120: It's not just a list, it has also ranges, e.g. 2-4. The syntax [1] is called cpulist, sometimes also cpu list, CPU list or cpu-list, CPU-list, the convention is not clear. In TuneD for consistence we usually use the term "cpulist".

Recently the group modifiers were added to the syntax (:size/group) and N which we do not support yet in TuneD, but I probably wouldn't mention this maybe temporal limitation in the man.

[1] https://www.kernel.org/doc/html/next/admin-guide/kernel-parameters.html#cpu-lists

yarda:
Thanks, maybe just update the man and built-in help, otherwise OK.

yarda:
I was thinking a bit more about it. Maybe we could drop the -c | --cpu-list arguments and rely only on the "cpulist:" prefix the API supports and just document it. Maybe it's more consistent approach? What do you think?

zacikpa:
> I was thinking a bit more about it. Maybe we could drop the -c | --cpu-list arguments and rely only on the "cpulist:" prefix the API supports and just document it. Maybe it's more consistent approach? What do you think?

I think that makes sense. As you say, it's more consistent, and we don't duplicate the handling of the option. I will remove it and mention the `cpulist:` prefix in the manpages.

yarda (approved):
LGTM, thanks.

PR #567, opened 2023-11-09
Multi Profile option in TuneD

veda10:
Introducing multi_profile option in TuneD which does the below: 
1) Apply a profile on specific set of cores
2) Apply multiple profiles on different sets of cores simultaneously.

More on multi_profile :
https://docs.google.com/document/d/1Tb6LygN8aM5pdX7akqNe3Wsn65O3oBkPZBEKEX6tZ_s/edit

### DIFF
--- a/tuned/plugins/base.py
+++ b/tuned/plugins/base.py
@@ -95,13 +95,15 @@ class Plugin(object):
 
 	def create_instance(self, name, devices_expression, devices_udev_regex, script_pre, script_post, options):
 		"""Create new instance of the plugin and seize the devices."""
-		if name in self._instances:
-			raise Exception("Plugin instance with name '%s' already exists." % name)
 
 		effective_options = self._get_effective_options(options)
 		instance = self._instance_factory.create(self, name, devices_expression, devices_udev_regex, \
 			script_pre, script_post, effective_options)
-		self._instances[name] = instance
+
+		if name in self._instances:
+			self._instances[name].append(instance)
+		else:
+			self._instances[name] = [instance]
 
 		return instance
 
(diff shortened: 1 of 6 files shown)


yarda:
Sorry for the delay, I was busy with the TuneD FDP pre-release and related tasks.

I think that with the current syntax you can achieve nearly everything from the proposal, e.g.:
1) To apply a profile on specific set of cores
It's currently possible, e.g. stock profile named 'PROFILE':
```
[cpu]
governor=performance
```
Custom profile:
```
[main]
include=PROFILE

[cpu]
devices=${f:cpulist2devs:1,3-5}
```
The custom profile can be even named PROFILE and put into /etc/tuned, then the `tuned-adm profile PROFILE` will apply it only on the CPUS 1,3-5.

2) Apply Multiple profiles
There is a bit limitation, you need to have the "template" profiles with different instance names, e.g:

PROFILE1:
```
[cpu_perf]
type=cpu
gorvernor=performance
```
PROFILE2:
```
[cpu_power]
type=cpu
gorvernor=powersave
```
And the final profile loading multiple profiles and aplying them on specific CPUs:
```
[main]
include=PROFILE1, PROFILE2

[cpu_power]
type=cpu
devices=cpu1

[cpu_perf]
type=cpu
devices=cpu3,cpu4,cpu5
```

For the 2) to work with the stock profiles all we would have to do is to rewrite the profiles to use unique instance names, e.g. instead of the following in the balanced profile:
```
[cpu]
...
```

which is semantically equivalent and just a shortcut to:
```
[cpu]
type=cpu
devices=*
...
```

we could have the following:
```
[cpu_balanced]
type=cpu
devices=*
```

I am also thinking about per-instance include for some time,  but I have never implemented it, which could be also useful, e.g.:
```
[cpu_power]
type=cpu
include=powersave:cpu
...
```
Which would include to the instance [cpu_power] just the instance [cpu] from the profile "powersave" and then you could override whatever you want.

But other than that I wouldn't probably complicate the current behavior. I am keeping this open for a discussion..

veda10:
Hi Yarda, Thanks for your feedback. I was not aware of the multiprofile option already being present in TuneD. I see it is not properly documented anywhere. Maybe that's the reason. Anyway, we are trying to understand it and check if it serves our purpose. Will reach out to you when we need help. Thank you.

PR #569, opened 2023-11-13
Add API functions to retrieve active instances and their devices

zacikpa:
Resolves: RHEL-15137

Appropriate commands are added to `tuned-adm` as well:
- `get_instances`: Retrieves a list of active instances of a given plugin or all active instances.
- `instance_get_devices`: Retrieves a list of devices assigned to a given instance.

To differentiate between cases when a plugin does not exist (error) and when it does not have any active
instances (not an error), `get_instances` also returns a boolean and an error string.

Similarly in `instance_get_devices` - if an instance exists but does not have any assigned devices, it is _not_ an error.
If it does not exist, it _is_ an error.

Let me know if this behaviour is OK or whether I should simplify it by just returning empty lists.


tuned/daemon/controller.py
```diff
@@ -373,3 +373,48 @@ def instance_acquire_devices(self, devices, instance_name, caller = None):
...
+
+		Return:
+		bool -- True on success
+		string -- error message or "OK"
+		list of string pairs -- [(instance_name, plugin_name)]
+		"""
+		if caller == "":
+			return (False, "Unauthorized", [])
+		active_instances = filter(lambda instance: instance.active, self._daemon._unit_manager.instances)
+		all_pairs = map(lambda instance: (instance.name, instance.plugin.name), active_instances)
```
yarda, line 392: Maybe better to filter first according to the plugin name and then construct the final map? Maybe do all filtering in the first filter lambda together with the active instance. Nevertheless even in this implementation the memory footprint is probably negligible and memory is consumed only during the API call.

tuned/daemon/controller.py
```diff
@@ -373,3 +373,48 @@ def instance_acquire_devices(self, devices, instance_name, caller = None):
 			log.info(rets)
 			return (False, rets)
 		return (True, "OK")
+
+	@exports.export("s", "(bsa(ss))")
```
yarda, line 377: Are the outer parenthesis needed for the D-Bus type signature?

yarda (commented):
Thanks, it looks good. I have only few minor comments which aren't blockers.

tuned/daemon/controller.py
```diff
@@ -373,3 +373,48 @@ def instance_acquire_devices(self, devices, instance_name, caller = None):
...
+		Return:
+		bool -- True on success
+		string -- error message or "OK"
+		list of strings -- device names
+		"""
+		if caller == "":
+			return (False, "Unauthorized", [])
+		for instance in self._daemon._unit_manager.instances:
+			if instance.name == instance_name:
+				return (True, "OK", list(instance.processed_devices))
```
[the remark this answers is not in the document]
zacikpa: Great idea, updated.

tuned/daemon/controller.py
zacikpa: Makes sense. I refactored it a bit more to make the code prettier.

tuned/daemon/controller.py
zacikpa: Apparently it doesn't matter, but we do use them in other functions as well (e.g., `auto_profile`), so I added them to `instance_get_devices` too and kept them here.

zacikpa:
@yarda Please review the updates and I will squash then.

tuned/daemon/controller.py
yarda: OK, thanks for the info.

yarda:
EPEL-8 CI failure is infra error, i.e. it is not blocking the merge.

yarda:
LGTM, please squash.

zacikpa:
> LGTM, please squash.

Done.

PR #570, opened 2023-11-14
Fixed REs escape sequences not to trigger SyntaxWarning

yarda:
Python 3.12 is more strict about correct escaping and outputs SyntaxWarning for invalid escape sequences that were previously silently corrected.

### DIFF
--- a/tuned/plugins/plugin_scheduler.py
+++ b/tuned/plugins/plugin_scheduler.py
@@ -135,7 +135,7 @@ class SchedulerUtilsSchedutils(SchedulerUtils):
 		return schedutils.get_priority_max(sched)
 
 class SchedulerPlugin(base.Plugin):
-	"""
+	r"""
 	`scheduler`::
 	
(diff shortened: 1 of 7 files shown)


tuned/plugins/plugin_scheduler.py
```diff
@@ -135,7 +135,7 @@ def get_priority_max(self, sched):
 		return schedutils.get_priority_max(sched)
 
 class SchedulerPlugin(base.Plugin):
-	"""
+	r"""
```
zacikpa, line 138: With this change you also accidentally fixed another issue :slightly_smiling_face:.

In multiple places, the docstring contains `\b` which escaped to backspace characters and caused the docs to display incorrectly when calling `get_plugin_documentation`. Of course, this doesn't happen with raw strings.

zacikpa (approved):
Looks good!

yarda:
Thanks.

PR #572, opened 2023-11-20
Be explicit about amd-pstate support

zacikpa:
Since intel_pstate and amd-pstate use the same per-CPU interface for setting EPP hints, the current cpu plugin supports amd-pstate out of the box.

This commit modifies current P-state related logs and variable names to be explicit about the support for amd-pstate.

Note that while intel_pstate supports global attributes such as `max_perf_pct`, amd-pstate does not.

### DIFF
--- a/tuned/plugins/plugin_cpu.py
+++ b/tuned/plugins/plugin_cpu.py
@@ -81,7 +81,8 @@ class CPULatencyPlugin(hotplug.Plugin):
 	
 	`energy_performance_preference`:::
 	[option]`energy_performance_preference` supports managing energy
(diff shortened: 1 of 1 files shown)


yarda:
Please add Jira ticket reference to the commit message.

zacikpa:
Right, forgot about that. Fixed.

yarda:
Thanks, LGTM.

Personally, I don't like that the Intel driver is called intel_pstate, but the AMD driver is called amd-pstate (i.e. "_" vs "-"). It seems this inconsistency is in the kernel sources and even the registered cpufreq driver names are inconsistent, so let's follow the kernel.

PR #573, opened 2023-11-24
Add a plugin for the ACPI driver

zacikpa:
Resolves: RHEL-16966

This PR enables explicit support of ACPI `platform_profile` option, which was previously only supported via the generic `sysfs` plugin.

Appropriate values of the `platform_profile` option were also added to basic profiles - `balanced`, `powersave` and `*-performance`.

When ACPI platform profile is not supported by the system and a profile contains the `platform_profile` option, tuning verification fails unless `--ignore-missing` is turned on. @yarda, please confirm whether this is the desired behavior - especially since the option is being added to the stock profiles.

I'm also thinking of refactoring all `@command_set` functions, which accept alternative values separated by `|`. There are currently at least 4 of them, and there is a lot of unnecessary code duplication. That, would, however, be done in a separate PR.


yarda (changes requested):

yarda:
I think the behavior of ignore-missing is OK. It seems many people already run it in this mode and we have never guaranteed that we won't add more platform specific features to the stock profiles. Maybe we could add global configuration option for it.

tuned/plugins/plugin_acpi.py
```diff
@@ -0,0 +1,78 @@
...
+		return os.path.join(ACPI_DIR, "platform_profile_choices")
+
+	@classmethod
+	def _platform_profile_path(cls):
+		return os.path.join(ACPI_DIR, "platform_profile")
+
+	@command_set("platform_profile")
+	def _set_platform_profile(self, profiles, sim):
+		if sim:
+			return str(profiles)
```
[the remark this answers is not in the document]
zacikpa: The default value verification does, in fact, have the [option to check alternatives](https://github.com/redhat-performance/tuned/blob/2652d7c29d32e1ec97d6dcb2ed2e424594058523/tuned/plugins/base.py#L547) and it's used, e.g.,  when verifying EPP in the CPU plugin. That being said, I do like the option you propose more.

As I said in the initial comment, it would be nice to refactor the setters with alternatives so they behave the same (or do you think that the behavior differs on purpose?).

tuned/plugins/plugin_acpi.py
```diff
@@ -0,0 +1,78 @@
...
+		if not os.path.isfile(self._platform_profile_path()):
+			log.debug("ACPI platform_profile is not supported on this system")
+			return None
+		profiles = [profile.strip() for profile in profiles.split('|')]
+		avail_profiles = set(self._cmd.read_file(self._platform_profile_choices_path()).split())
+		for profile in profiles:
+			if profile in avail_profiles:
+				self._cmd.write_to_file(self._platform_profile_path(), profile)
+				log.info("Setting platform_profile to '%s'" % profile)
+				return None
```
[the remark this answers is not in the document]
zacikpa: Updated.

profiles/powersave/tuned.conf
```diff
@@ -9,6 +9,9 @@ summary=Optimize for low power consumption
 governor=ondemand|powersave
 energy_perf_bias=powersave|power
 
+[acpi]
+platform_profile=low-power
```
ryanabx, line 13: On some devices (including my own, ASUS GA402RJ) the `low-power` option doesn't exist, and is instead `quiet`. I'm not sure how tuned configs handle the case where an option doesn't exist, but maybe putting `low-power|quiet` would solve it? Someone more familiar with tuned than I could probably answer that.

doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
```diff
@@ -215,3 +215,26 @@ cmdline=isolcpus=2
...
+a generic power/performance preference API for other drivers.
+Multiple profiles can be specified, separated by `|`.
+The first available profile is selected.
++
+--
+.Selecting a platform profile
+====
+----
+[acpi]
+platform_profile=balanced|low-power
```
ryanabx, line 235: See my other comment, but `low-power` sometimes doesn't exist on certain devices, and is instead `quiet`

ryanabx (changes requested):

profiles/powersave/tuned.conf
zacikpa: Thanks for noticing this. Yes, your proposed solution should work, I updated the profile.

doc/manual/modules/performance/ref_available-tuned-plug-ins.adoc
zacikpa: I updated the example in the doc to the one that is now present in the power-saving profile (`low-power|quiet`).

ryanabx (approved):
LGTM

yarda (changes requested):
```
2024-01-18 01:05:48,566 ERROR    tuned.units.manager: BUG: Unhandled exception in start_tuning: ACPIPlugin._set_platform_profile() got an unexpected keyword argument 'remove'
2024-01-18 01:05:48,566 ERROR    tuned.units.manager: Traceback (most recent call last):
  File "/home/yarda/tuned-private/tuned/units/manager.py", line 119, in _try_call
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
  File "/home/yarda/tuned-private/tuned/plugins/instance/instance.py", line 80, in apply_tuning
    self._plugin.instance_apply_tuning(self)
  File "/home/yarda/tuned-private/tuned/plugins/base.py", line 261, in instance_apply_tuning
    self._instance_apply_static(instance)
  File "/home/yarda/tuned-private/tuned/plugins/base.py", line 320, in _instance_apply_static
    self._execute_all_non_device_commands(instance)
  File "/home/yarda/tuned-private/tuned/plugins/base.py", line 436, in _execute_all_non_device_commands
    self._execute_non_device_command(instance, command, new_value)
  File "/home/yarda/tuned-private/tuned/plugins/base.py", line 517, in _execute_non_device_command
    command["set"](new_value, sim = False, remove = False)
TypeError: ACPIPlugin._set_platform_profile() got an unexpected keyword argument 'remove'
```
Please update the PR to reflect the recent internal API change caused by #578. Sorry for the out of order merge, the #578 had priority due to FDP release. Otherwise LGTM.

zacikpa:
Updated, rebased, squished. The pipeline error is unrelated to the change.

yarda:
Thanks LGTM.

The rawhide CI failure is probably caused by some temporal breakage in rawhide:
```
Error: 
 Problem: package tuned-utils-systemtap-2.21.0-1.20240122103218878382.pr573.34.g9b72d69.fc40.noarch from @commandline requires /usr/bin/stap, but none of the providers can be installed
  - conflicting requests
  - nothing provides libboost_system.so.1.81.0()(64bit) needed by systemtap-client-5.0~pre16958465gca71442b-1.fc40.x86_64 from testing-farm-tag-repository
  - nothing provides libboost_system.so.1.81.0()(64bit) needed by systemtap-devel-5.0~pre16958465gca71442b-1.fc40.x86_64 from testing-farm-tag-repository
  - nothing provides libboost_system.so.1.81.0()(64bit) needed by systemtap-client-5.0~pre16958465gca71442b-1.fc40.x86_64 from rawhide
  - nothing provides libboost_system.so.1.81.0 needed by systemtap-devel-5.0~pre16958465gca71442b-1.fc40.i686 from rawhide
  - nothing provides libboost_system.so.1.81.0()(64bit) needed by systemtap-devel-5.0~pre16958465gca71442b-1.fc40.x86_64 from rawhide
```
It seems `/usr/bin/stap` is still provided by systemtap-client-5.1-17056139g6c0b92b3, thus ignoring this failure.

PR #574, opened 2023-11-28
profiles: Improve 'isolated_cores=' help text

yarda:
Just explain what isolated_cores= is for and recommend to only have one isolated_cores= line to avoid wrong system configuration.


zacikpa (approved):
Looks good, approving.

PR #576, opened 2023-12-04
cpu_plugin: Add property "uncore_max_freq_khz"

sgruszka:
Add a property "uncore_max_delta_khz" to cpu_plugin, so that uncore maximum frequency can be set. The property is exported per die, hence add extra logic to map it to per cpu setting and check for wrong config.

Refer to the following link to get details uncore frequency scaling: https://docs.kernel.org/admin-guide/pm/intel_uncore_frequency_scaling.html

This extends PR #519 to make uncore frequency defined per cpu as proposed in https://github.com/redhat-performance/tuned/pull/519#issuecomment-1602367982

For now, for simplicity , uncore max_freq_khz is configured as direct value.  Also uncore's that are not defined as package_NN_die_MM are not covered in this PR.

### DIFF
--- a/tuned/plugins/plugin_cpu.py
+++ b/tuned/plugins/plugin_cpu.py
@@ -14,6 +14,11 @@ log = tuned.logs.get()
 
 cpuidle_states_path = "/sys/devices/system/cpu/cpu0/cpuidle"
 
+class Die:
+	def __init__(self):
+		self.cpus = set()
+		self.new_values = dict()
+
 class CPULatencyPlugin(hotplug.Plugin):
 	"""
 	`cpu`::
@@ -185,6 +190,10 @@ class CPULatencyPlugin(hotplug.Plugin):
 	pm_qos_resume_latency_us=100
 	----
 	Allows any C-state with a resume latency less than value.
+	[cpu]
+	uncore_max_freq_khz=0
+	----
+	Set the maximum uncore frequency that hardware will use.
 	"""
 
 	def __init__(self, *args, **kwargs):
@@ -199,14 +208,63 @@ class CPULatencyPlugin(hotplug.Plugin):
 		self._has_intel_pstate = False
 		self._has_amd_pstate = False
 		self._has_pm_qos_resume_latency_us = None
+		self._has_topology = False
 
+		self._uncore_max_delta_khz_save = None
 		self._min_perf_pct_save = None
 		self._max_perf_pct_save = None
 		self._no_turbo_save = None
 		self._governors_map = {}
 		self._cmd = commands()
 
+		self._topology = dict()
+		self._topology_config_wrong = False
+
 		self._flags = None
+
+	def _cpu_get_package_die_id(self, device):
+		dir = '/sys/devices/system/cpu/%s/topology/' % device
+		if not os.path.exists(dir):
+			return (-1, -1)
+		die_id = self._str2int(self._cmd.read_file(dir + "/die_id"))
+		package_id = self._str2int(self._cmd.read_file(dir + "/physical_package_id"))
+		return (package_id, die_id)
+
+	def _init_topology(self):
+		for device in self._free_devices:
+			package_id, die_id = self._cpu_get_package_die_id(device)
+			if package_id == -1:
+				return
+
+			if package_id not in self._topology:
+				self._topology[package_id] = dict()
+			if die_id not in self._topology[package_id]:
+				self._topology[package_id][die_id] = Die()
+
+			self._topology[package_id][die_id].cpus |= {device}
+		self._has_topology = True
+
+	def _check_topology(self):
+		if not self._has_topology:
+			return
+
+		# All assigned devices has to be part of the same die's
+		assigned_devices = self._assigned_devices
+		while len(assigned_devices) > 0:
+			device = assigned_devices.pop()
+			assigned_devices.add(device)
+
+			package_id, die_id = self._cpu_get_package_die_id(device)
+			if package_id == -1:
+				self._topology_config_wrong = True
+				return
+
+			cpus = self._topology[package_id][die_id].cpus
+			if len(cpus - assigned_devices) > 0:
+				self._topology_config_wrong = True
+				return
+
+			assigned_devices -= cpus
 
 	def _init_devices(self):
 		self._devices_supported = True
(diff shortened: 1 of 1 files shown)


tuned/plugins/plugin_cpu.py
```diff
@@ -582,6 +644,57 @@ def _get_sampling_down_factor(self, device, ignore_missing=False):
 			return None
 		return self._cmd.read_file(path).strip()
 
+	@command_set("uncore_max_freq_khz", per_device=True)
+	def _set_uncore_max_freq_khz(self, uncore_max_freq_khz, device, sim):
```
bartwensley, line 648: Allowing per-cpu configuration of something that cannot be configured for individual cpus does not make sense to me. The uncore frequency is configurable for each package/die combination - not for individual cpus.

sgruszka: I generally agree with that. From simplicity perspective I think better would be separate plugin that allow configure individual uncore's based on name in sysfs i.e. package_NN_die_MM or uncoreNN. I'm ok to implement such plugin, if that's better. 
However @yarda requested per cpus option. My understanding is that, this is more friendly for users using CPU-partitioning profiles , as they could then easier match cpus on which partitioning setting are used.

Anyway I'm ok with both implementations, but would like to know which is preferable,  @yarda could you please provide feedback here?

sgruszka: If the new plugin i.e. plugin_intel_uncore.py would be created , 
it could have config/device specification in configs files like this:

[intel_uncore]
max_freq_khz_delta=200000

[intel_io_uncores]
type=intel_uncore
devices=uncore*
max_freq_khz = 4000000

[intel_package_2]
type=intel_uncore
devices=package_02_die_00
max_freq_khz_procentage=95

The device names would correspond to entries in  /sys/devices/system/cpu/intel_uncore_frequency/
The frequency specification could be any: percentage, direct value or delta. 

Would that be ok? Feedback appreciated.

bartwensley: I'm OK with doing this in the existing cpu plugin - but with a single value for all CPUs (i.e. don't set "per_device=True"). Would that work?

sgruszka: I think yes,  that is basically how it was done is original PR #519  - we have one global settings and apply it to all uncore's .

yarda: The cpu plugin has 'per CPU' granularity. So the `uncore_max_freq_khz` needs to be settable per CPU, otherwise we could break things / TuneD logic. If it's unacceptable we would probably need new plugin with the package/die granularity.

Imagine you can now have e.g. the following TuneD profile:

```
[cpu_perf]
type=cpu
devices=${f:cpulist2devs:1-4}
governor=performance
...

[cpu_idle]
type=cpu
devices=*
governor=powersave
...
```

I.e. CPUs 1-4 will be configured for performance and the rest of CPUs will be configured for powersave. Then at the runtime through the TuneD API you can move e.g. CPU 6 from the [cpu_idle] instance to the [cpu_perf] instance and it will be reconfigured accordingly.

In case of per CPU uncore it could be the following:
```
[cpu_perf]
type=cpu
devices=${f:cpulist2devs:1-4}
governor=performance
uncore_max_freq_khz=4000000

[cpu_idle]
type=cpu
devices=*
governor=powersave
uncore_max_freq_khz=3000000
```

In [cpu_perf] it will go through CPUs 1, 2, 3, 4, it will find out which package/die it is and sets there the uncore freq 4000000. In case of conflicts the last write wins. Then the same for the [cpu_idle] instance - it will go through the rest of CPUs, matches each CPU with the package/die and sets there the uncore freq 3000000.

Then if the CPU 6 from the [cpu_idle] instance is moved to the [cpu_perf] instance through the API, the package/die is matched and its uncore is set to 4000000. 

It's a clean extension of the current logic, but I agree that it may be a bit confusing. But you could also extend it by e.g. the uncore matcher function like the cpulist2devs function we already have. This should make things a little cleaner, e.g.:
```
[intel_package_2]
type=cpu
devices=${f:uncore2devs:package_02_die_00}
uncore_max_freq_khz=4000000

[cpu_default_uncore]
type=cpu
devices=*
uncore_max_freq_khz=3000000
```

I think it could work. If not, we would probably need new plugin with the package/die granularity.

yarda: The matcher function could have various syntax, e.g.:
```
...
devices=${f:uncore2devs:02}  # package 02, any die
...
devices=${f:uncore2devs:02:00}  # package 02, die 00
...
devices=${f:uncore2devs:0*:0*}  # any package starting with 0, any die starting with 0
...
```
etc. This selector function would go under the tuned/profiles/functions directory, where built-in functions live.

spandruvada: In processors like GNR and beyond there are uncore control for IO dies, which are not associated with any CPU.
So we should avoid relating CPUs to uncore group. If we do we need to do special for IO dies.
As a first step can we do for whole package or system?

yarda: Thanks for the info. Then it probably doesn't make sense to add this functionality to the CPU plugin which works over CPUs and we probably need the uncore TuneD plugin that will work over devices which has uncore control.

sgruszka: Ok, thanks for feedback. Going to work on the plugin ... should be relatively easy to implement .

bartwensley: @yarda - you said "The cpu plugin has 'per CPU' granularity. So the uncore_max_freq_khz needs to be settable per CPU, otherwise we could break things / TuneD logic. If it's unacceptable we will probably need new plugin with the package/die granularity."

But what does the "per_device" option do when creating the command?
`@command_set("uncore_max_freq_khz", per_device=True)
`

Doesn't setting "per_device=False" allow a command to be defined that cannot be settable per CPU? Or is it for some other purpose?

yarda: `per_device=False` (non-device commands) are mostly used for non-device plugins (e.g. `bootloader`). For device based plugins it can be used for setting some property of the plugin instance (for example delay, path prefix, units, ...) to modify device commands behavior of the instance. It shouldn't interfere with other instances of the plugin. It also doesn't support hotplug, because hotplug is device based thing, not instance based.

In case of uncore implementation as individual plugin you could for example in the TuneD profile define uncore plugin instance for high performance and uncore plugin instance for powersave (similarly as we now do with the CPUs) and later move dynamically packages/dies between the instances as needed through the runtime API. This wouldn't work if implemented in the cpu plugin with the non-device command. You could also use the built-in regex matcher to match the packages/dies via regular expressions in the TuneD profile, i.e. you could work with them in TuneD as with other devices.

sgruszka:
Created new PR #590 with intel uncore plugin , closing this one.

PR #578, opened 2023-12-06
hotplug: do not report ENOENT errors on device remove

yarda:
It requires slight internal plugin API change, thus 3rd party plugins needs updating.

Namely the 'command_set' methods for devices were extended by the 'remove' boolean parameter which is set to 'True' on the hotplug remove event. Then the method should silent the ENOENT errors where appropriate, because the device interface may be removed (by kernel/udev) before the method finishes.

Resolves: RHEL-11342


jmencak (approved):
I want through the code and the changes make sense to me.  Thank you for fixing this not only for the [disk] plug-in.  I've also tested the new code with the [disk] plug-in and removing a block disk device when TuneD was running.  No ERRORs were reported.

tuned/plugins/plugin_audio.py
```diff
@@ -96,12 +98,13 @@ def _get_timeout(self, device, ignore_missing=False):
 		return None
 
 	@command_set("reset_controller", per_device = True)
-	def _set_reset_controller(self, value, device, sim):
+	def _set_reset_controller(self, value, device, sim, no_error, remove):
```
zacikpa, line 101: I assume the `no_error` parameter here is a mistake?

zacikpa (commented):
Looks ok, just one note:

tuned/plugins/plugin_audio.py
yarda: Good catch, thanks. It was residue of my previous attempt to fix this problem :) It should be fixed now.

PR #580, opened 2023-12-08
Tuning of individual interrupts

adriaan42:
To optimize latencies in realtime applications, we have a need to direct certain interrupts to certain cpus running our realtime processes. In addition, the cpu allocation happens dynamically (e.g., applications deployed as containers).

There's two parts to this:

- Handle interrupts individually: current implementation of the `scheduler` plugin can only set affinities of ALL interrupts.
- Set the affinities dynamically: current implementation requires the affinity to be set statically in the profile.

This PoC

- Adds a new plugin `irq`, which treats interrupts as "devices", so we can have multiple plugin instances, each with different affinities, and each tuning different irqs.
- Adds functions to dynamically create and destroy plugin instances. This way, when a new application is started, we can create an instance of the irq plugin with the desired affinity, and then use the existing `instance_acquire_devices` to apply the selected affinity to individual interrupts.

Example, on a 4-core machine, with an `irq` section in the profile:
```ini
[irq]
affinity=0
```

And then dynamically changing the affinity of irq16:

```sh
# busctl call com.redhat.tuned /Tuned com.redhat.tuned.control instance_create ssa{ss} irq myinstance 1 affinity 2
bs true "OK"
# busctl call com.redhat.tuned /Tuned com.redhat.tuned.control instance_acquire_devices ss irq16 myinstance
(bs) true "OK"
```

And in the log:
```
2023-12-08 15:16:05,638 INFO     tuned.daemon.controller: Created dynamic instance 'myinstance' of plugin 'irq'
2023-12-08 15:16:29,263 INFO     tuned.daemon.controller: Moving devices '{'irq16'}' from instance 'irq' to instance 'myinstance'.
2023-12-08 15:16:29,263 INFO     tuned.plugins.plugin_irq: _instance_unapply_dynamic(irq, irq16)
2023-12-08 15:16:29,264 INFO     tuned.plugins.plugin_irq: Setting SMP affinity of IRQ 16 to '0000000f'
2023-12-08 15:16:29,264 INFO     tuned.plugins.hotplug: instance myinstance: adding new device irq16
2023-12-08 15:16:29,265 INFO     tuned.plugins.plugin_irq: _instance_apply_dynamic(myinstance, irq16)
2023-12-08 15:16:29,265 INFO     tuned.plugins.plugin_irq: Setting SMP affinity of IRQ 16 to '00000004'
```

Open questions

- What happens to the irq affinity feature of the `scheduler` plugin? For backwards-compatibility I guess we'll need to keep the feature and a switch to be able to tell `scheduler` to not touch interrupt affinities in case we want to use the new plugin?
- Can we move devices between plugin instances without running "unapply"? In my example above the interrupt affinity briefly gets restored to 0xf before the new desired value is applied...
- What happens when TuneD is restarted? Currently, any dynamic instances would be lost. But currently, also any device assignments that have been made using `instance_aquire_devices` are lost, right?

### DIFF
--- a/tuned/plugins/plugin_irq.py
+++ b/tuned/plugins/plugin_irq.py
@@ -1,0 +1,272 @@
+from . import hotplug
+from .decorators import *
+import tuned.consts as consts
+import tuned.logs
+
+import errno
+import os
+
+log = tuned.logs.get()
+
+# the plugin manages each IRQ as a "device" and keeps a IrqInfo object for it
+class IrqInfo(object):
+	def __init__(self, irq):
+		self.irq = irq
+		self.device = "irq%s" % irq
+		self.unchangeable = False
+		self.original_affinity = None
+
+class IrqPlugin(hotplug.Plugin):
+	r"""
+	`irq`::
+
+	Allows tuning of IRQ affinities, and thus re-implements functionality
+	already present in the `scheduler` plugin. However, this plugin offers
+	more flexibility, as it allows tuning of individual interrupts with
+	different affinities. When using the `irq` plugin, make sure to disable
+	IRQ processing in the `scheduler` plugin by setting its option
+	[option]`irq_process=false`.
+	The plugin handles individual IRQs as `devices`, and multiple plugin
+	instances can be defined, each addressing different devices/irqs.
+	The `device` names used by the plugin are `irq<n>`, where `<n>` is the
+	IRQ number. The special `device` `DEFAULT` controls values written to
+	`/proc/irq/default_smp_affinity`, which applies to all non-active IRQs.
+	===
+	The option [option]`affinity` controls the IRQ affinity to be set. It is
+	a string in "cpulist" format (such as 1,3-4). If the configured affinity
+	is empty, then the affinity of the respective IRQs is not touched.
+	===
+	The option [option]`mode` is a string which can either be `set` (default)
+	or `intersect`. In `set` mode the [option]`affinity` is always written
+	as configured, whereas in `intersect` mode, the new affinity will be
+	calculated as the intersection of the current and the configured affinity.
+	If that intersection is empty, the configured affinity will be used.
+	--
+	.Example moving all IRQs to CPU0, except irq16, which is directed to CPU2
+	====
+	----
+	[irq_special]
+	type=irq
+	devices=irq16
+	affinity=2
+	[irq]
+	affinity=0
+	----
+	====
+	"""
+
+	def __init__(self, monitor_repository, storage_factory, hardware_inventory, device_matcher, device_matcher_udev, plugin_instance_factory, global_cfg, variables):
+		super(IrqPlugin, self).__init__(monitor_repository, storage_factory, hardware_inventory, device_matcher, device_matcher_udev, plugin_instance_factory, global_cfg, variables)
+		self._irqs = {}
+
+	#
+	# plugin-level methods: devices and plugin options
+	#
+	def _init_devices(self):
+		"""Read /proc/irq to collect devices
+		"""
+		self._devices_supported = True
+		self._free_devices = set()
+		self._assigned_devices = set()
+		for i in os.listdir("/proc/irq"):
+			p = os.path.join("/proc/irq", i)
+			if os.path.isdir(p) and i.isdigit():
+				info = IrqInfo(i)
+				self._irqs[i] = info
+				self._free_devices.add(info.device)
+		# add the virtual device for default_smp_affinity
+		default_info = IrqInfo("DEFAULT")
+		default_info.device = "DEFAULT"
+		self._irqs["DEFAULT"] = default_info
+		self._free_devices.add(default_info.device)
+
+	@classmethod
+	def _get_config_options(cls):
+		return {
+			"affinity": "",
+			"mode": "set",
+		}
+
+	#
+	# instance-level methods: implement the Instance interface
+	#
+	def _instance_init(self, instance):
+		instance._has_static_tuning = True
+		instance._has_dynamic_tuning = False
+
+		affinity = instance.options.get("affinity")
+		affinity_list = self._cmd.cpulist_unpack(affinity)
+		if len(affinity.strip()) == 0:
+			# empty affinity in profile -> assume it's intentional
+			log.info("Instance '%s' configured with empty affinity. Deactivating." % instance.name)
+			instance._active = False
+		elif len(affinity_list) == 0:
+			# non-empty affinity string evaluates to empty list -> assume parse error
+			log.error("Instance '%s' with invalid affinity '%s'. Deactivating." % (instance.name, affinity))
+			instance._active = False
+
+		mode = instance.options.get("mode")
+		if mode not in ["set", "intersect"]:
+			log.error("Invalid operating mode '%s' for instance '%s'. Using the default 'set' instead."
+					% (mode, instance.name))
+			instance.options["mode"] = "set"
+
+	def _instance_cleanup(self, instance):
+		pass
+
+	def _instance_apply_static(self, instance):
+		log.debug("Applying IRQ affinities (%s)" % instance.name)
+		super(IrqPlugin, self)._instance_apply_static(instance)
+
+	def _instance_unapply_static(self, instance, rollback):
+		log.debug("Unapplying IRQ affinities (%s)" % instance.name)
+		super(IrqPlugin, self)._instance_unapply_static(instance, rollback)
+
+	def _instance_verify_static(self, instance, ignore_missing, devices):
+		log.debug("Verifying IRQ affinities (%s)" % instance.name)
+		return super(IrqPlugin, self)._instance_verify_static(instance, ignore_missing, devices)
+
+	#
+	# "low-level" methods to get/set irq affinities
+	#
+	def _get_irq_affinity(self, irq):
+		"""Get current IRQ affinity from the kernel
+
+		Args:
+			irq (str): IRQ number (as string) or "DEFAULT"
+
+		Returns:
+			affinity (set): set of all CPUs that belong to the IRQ affinity mask,
+				if reading of the affinity fails, an empty set is returned
+		"""
+		try:
+			filename = "/proc/irq/default_smp_affinity" if irq == "DEFAULT" else "/proc/irq/%s/smp_affinity" % irq
+			with open(filename, "r") as f:
+				affinity_hex = f.readline().strip()
+			return set(self._cmd.hex2cpulist(affinity_hex))
+		except (OSError, IOError) as e:
+			log.debug("Failed to read SMP affinity of IRQ %s: %s" % (irq, e))
+			return set()
+
+	def _set_irq_affinity(self, irq, affinity, restoring):
+		"""Set IRQ affinity in the kernel
+
+		Args:
+			irq (str): IRQ number (as string) or "DEFAULT"
+			affinity (set): affinity mask as set of CPUs
+			restoring (bool): are we rolling back a previous change?
+
+		Returns:
+			status (int):  0 on success, -2 if changing the affinity is not
+				supported, -1 if some other error occurs
+		"""
+		try:
+			affinity_hex = self._cmd.cpulist2hex(list(affinity))
+			log.debug("Setting SMP affinity of IRQ %s to '%s'" % (irq, affinity_hex))
+			filename = "/proc/irq/default_smp_affinity" if irq == "DEFAULT" else "/proc/irq/%s/smp_affinity" % irq
+			with open(filename, "w") as f:
+				f.write(affinity_hex)
+			return 0
+		except (OSError, IOError) as e:
+			# EIO is returned by
+			# kernel/irq/proc.c:write_irq_affinity() if changing
+			# the affinity is not supported
+			# (at least on kernels 3.10 and 4.18)
+			if hasattr(e, "errno") and e.errno == errno.EIO and not restoring:
+				log.debug("Setting SMP affinity of IRQ %s is not supported" % irq)
+				return -2
+			else:
+				log.error("Failed to set SMP affinity of IRQ %s to '%s': %s" % (irq, affinity_hex, e))
+				return -1
+
+	#
+	# "high-level" methods: apply tuning while saving original affinities
+	#
+	def _apply_irq_affinity(self, irqinfo, affinity, mode):
+		"""Apply IRQ affinity tuning
+
+		Args:
+			irqinfo (IrqInfo): IRQ that should be tuned
+			affinity (set): desired affinity
+		"""
+		original = self._get_irq_affinity(irqinfo.irq)
+		if mode == "intersect":
+			# intersection of affinity and original, if that is empty fall back to configured affinity
+			affinity = affinity & original or affinity
+		if irqinfo.unchangeable or affinity == original:
+			return
+		res = self._set_irq_affinity(irqinfo.irq, affinity, False)
+		if res == 0:
+			if irqinfo.original_affinity is None:
+				irqinfo.original_affinity = original
+		elif res == -2:
+			irqinfo.unchangeable = True
+
+	def _restore_irq_affinity(self, irqinfo):
+		"""Restore IRQ affinity
+
+		Args:
+			irqinfo (IrqInfo): IRQ that should be restored
+		"""
+		if irqinfo.unchangeable or irqinfo.original_affinity is None:
+			return
+		self._set_irq_affinity(irqinfo.irq, irqinfo.original_affinity, True)
+		irqinfo.original_affinity = None
+
+	def _verify_irq_affinity(self, irqinfo, affinity, mode):
+		"""Verify IRQ affinity tuning
+
+		Args:
+			irqinfo (IrqInfo): IRQ that should be verified
+			affinity (set): desired affinity
+
+		Returns:
+			status (bool): True if verification successful, False otherwise
+		"""
+		if irqinfo.unchangeable:
+			return True
+		affinity_description = "IRQ %s affinity" % irqinfo.irq
+		desired_affinity = affinity
+		desired_affinity_string = self._cmd.cpulist2string(self._cmd.cpulist_pack(list(desired_affinity)))
+		current_affinity = self._get_irq_affinity(irqinfo.irq)
+		current_affinity_string = self._cmd.cpulist2string(self._cmd.cpulist_pack(list(current_affinity)))
+		if mode == "intersect":
+			# In intersect mode, we don't use a strict comparison; it's sufficient
+			# if the current affinity is a subset of the desired one
+			desired_affinity_string = "subset of " + desired_affinity_string
+		if ((mode == "intersect" and current_affinity <= desired_affinity) or
+					(mode == "set" and current_affinity == desired_affinity)):
+			log.info(consts.STR_VERIFY_PROFILE_VALUE_OK
+					% (affinity_description, desired_affinity_string))
+			return True
+		else:
+			log.error(consts.STR_VERIFY_PROFILE_VALUE_FAIL
+					% (affinity_description, current_affinity_string, desired_affinity_string))
+			return False
+
+	#
+	# command definitions: entry to device-specific tuning
+	#
+	@command_custom("mode", per_device=False, priority=-10)
+	def _mode(self, enabling, value, verify, ignore_missing):
+		if (enabling or verify) and value is not None:
+			# Store the operating mode of the current instance in the plugin
+			# object, from where it is read by the "affinity" command.
+			# This works because instances are processed sequentially by the engine.
+			self._mode_val = value
+
+	@command_custom("affinity", per_device=True)
+	def _affinity(self, enabling, value, device, verify, ignore_missing):
+		irq = "DEFAULT" if device == "DEFAULT" else device[len("irq"):]
+		if irq not in self._irqs:
+			log.error("Unknown device: %s" % device)
+			return None
+		irqinfo = self._irqs[irq]
+		if verify:
+			affinity = set(self._cmd.cpulist_unpack(value))
+			return self._verify_irq_affinity(irqinfo, affinity, self._mode_val)
+		if enabling:
+			affinity = set(self._cmd.cpulist_unpack(value))
+			return self._apply_irq_affinity(irqinfo, affinity, self._mode_val)
+		else:
+			return self._restore_irq_affinity(irqinfo)


adriaan42:
@yarda Any thoughts on this? Do you think a feature like this could be accepted?

yarda:
Hi @adriaan42  this is interesting RFE.

> Open questions
> 
>     * What happens to the irq affinity feature of the `scheduler` plugin? For backwards-compatibility I guess we'll need to keep the feature and a switch to be able to tell `scheduler` to not touch interrupt affinities in case we want to use the new plugin?
>
We are already trying to address this  https://issues.redhat.com/browse/RHEL-21923

>     * Can we move devices between plugin instances without running "unapply"? In my example above the interrupt affinity briefly gets restored to 0xf before the new desired value is applied...
>
This is worth investigation and I was already thinking about it, because the "unapply" can be costly and can cause various interference. Generally, we need it, because one instance can change things which the other instance doesn't touch and  wouldn't be then rolled back. Maybe we could handle it by more intelligent e.g. "switch" method that would combine "apply" and "unapply". I.e. it could compare the options and roll back only those which are not handled by the second instance. For those which are handled by both instances the second instance could just take the saved previous values from the first instance. 

> 
>     * What happens when TuneD is restarted? Currently, any dynamic instances would be lost. But currently, also any device assignments that have been made using `instance_aquire_devices` are lost, right?
>
Yes, I think it's OK. Usually it maybe useful to start-over with the clear stock profile. If it is required to preserve the state, I think we could add API call to e.g. dump the profile into /run/tuned and make the /run/tuned preference over /etc/tuned.

yarda:
Please fix centos-7 CI failures. We are going to drop python 2.7 support once RHEL-7 is retired.

```
...
Compiling /builddir/build/BUILDROOT/tuned-2.21.0-1.20240118103851136139.pr580.33.ga7d8364.el7.x86_64/usr/lib/python2.7/site-packages/tuned/plugins/plugin_irq.py ...
  File "/usr/lib/python2.7/site-packages/tuned/plugins/plugin_irq.py", line 33
    log.info(f"_instance_init({instance.name})")
                                              ^
SyntaxError: invalid syntax
...
```

yarda:
For better backward compatibility I would drop 7799dd2b4bafcbc23f677145ef5f9266ec5657df and rely on fix from the  https://issues.redhat.com/browse/RHEL-21923 and would just change the TuneD profiles where needed.

tuned/daemon/controller.py
```diff
@@ -417,3 +417,74 @@ def instance_get_devices(self, instance_name, caller = None):
...
+		plugin = plugins[plugin_name]
+		instance = plugin.create_instance(instance_name, None, None, None, None, options)
+		plugin.initialize_instance(instance)
+		instance._created_dynamically = True
+		self._daemon._unit_manager.instances.append(instance)
+		log.info("Created dynamic instance '%s' of plugin '%s'" % (instance_name, plugin_name))
+		return (True, "OK")
+
+	@exports.export("ss", "bs")
+	def instance_destroy(self, plugin_name, instance_name, caller = None):
```
yarda, line 448: Maybe `instance_name` is enough, because AFAIK the instance name has to be unique.

yarda: But if it's there as a 'safety measure' then probably OK.

tuned/daemon/controller.py
```diff
@@ -417,3 +417,74 @@ def instance_get_devices(self, instance_name, caller = None):
...
+		if not instance._created_dynamically:
+			rets = "Cannot delete instance '%s' because it was not created dynamically" % instance_name
+			log.error(rets)
+			return (False, rets)
+		if not instance.plugin.name == plugin_name:
+			rets = "Instance '%s' does not belong to plugin '%s'" % (instance_name, plugin_name)
+			log.error(rets)
+			return (False, rets)
+		if len(instance._processed_devices) > 0:
+			rets = "Cannot delete instance '%s' because it still has attached devices" % instance_name
```
yarda, line 477: For simple PoC it's probably OK, but maybe you could free the devices and try to hotplug them (if possible).

tuned/daemon/controller.py
```diff
@@ -417,3 +417,74 @@ def instance_get_devices(self, instance_name, caller = None):
...
+		instance = None
+		for i in self._daemon._unit_manager.instances:
+			if i.name == instance_name:
+				instance = i
+				break
+		if instance is None:
+			rets = "Instance '%s' not found" % instance_name
+			log.error(rets)
+			return (False, rets)
+		if not instance._created_dynamically:
```
yarda, line 468: I think you have to define `_created_dynamically` in the instance constructor, otherwise it may traceback in case you try to destroy statically created instance:
```
('g-io-error-quark: GDBus.Error:org.freedesktop.DBus.Python.AttributeError: '
 "AttributeError: 'Instance' object has no attribute '_created_dynamically'\n"
 ' (36)')
```

yarda (changes requested):
Please fix the traceback.

yarda:
Maybe the instance_create/destroy could be more generic, i.e.:
* instance_create could have a list of devices or regular expressions to match devices to immediately try to hotplug. Watchout for the inconsistency that empty list of devices is now handled as a glob "*".  Maybe we could use some special keyword for it, like e.g. "EMPTY" or "none" which could be also used for static instance definition in the TuneD profile.
* instance_destroy could destroy all instances not only the dynamically created, i.e. statically created instances with no devices and in case there are some devices attached, it could try to release them and hotplug (like the hotplug code does).

Or maybe for simple PoC just allow destroying instances with no devices and do not distinguish between statically and dynamically created instances, i.e. if there are no devices it can be destroyed.

tuned/daemon/controller.py
yarda: This could happen if destroying instance of other type than `irq`, which has the initialization.

tuned/plugins/plugin_irq.py
```diff
@@ -0,0 +1,113 @@
...
+		log.info(f"_instance_cleanup({instance.name})")
+
+	def _instance_apply_static(self, instance):
+		log.info(f"_instance_apply_static({instance.name})")
+
+	def _instance_unapply_static(self, instance, rollback):
+		log.info(f"_instance_unapply_static({instance.name})")
+
+	def _instance_apply_dynamic(self, instance, device):
+		log.info(f"_instance_apply_dynamic({instance.name}, {device})")
```
yarda, line 58: This requires dynamic tuning enabled, which is by default disabled on RHEL. I think it should go under the static method. Or maybe better to implement it in the `command_custom` .

You could also implement it in the `command_set` /  `command_get` which will be called for each device/irq. It's probably the most clean solution, but it could have some overhead. Taking into account that on typical system there wouldn't be usually much IRQs, this could work.

yarda (changes requested):
I think it shouldn't be implemented as a dynamic tuning, because it's not run periodically (the `_instance_update_dynamic` method).

tuned/daemon/controller.py
adriaan42: I have removed the `plugin_name` argument. It's not needed, and now the function is consistent with others like `instance_get_devices`.

tuned/daemon/controller.py
adriaan42: I removed the `_created_dynamically` flag, making dynamically created instances equal to the "static" ones defined in the profile.

tuned/daemon/controller.py
```diff
@@ -417,3 +417,74 @@ def instance_get_devices(self, instance_name, caller = None):
...
+
+		Parameters:
+		plugin_name -- name of the plugin
+		instance_name -- name of the new instance
+
+		Return:
+		bool -- True on success
+		string -- error message or "OK"
+		"""
+		instance = None
```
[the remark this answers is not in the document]
adriaan42: Changed the implementation.

adriaan42:
### PR updates
* Changed scheduler plugin according to https://issues.redhat.com/browse/RHEL-21923
* Removed (hopefully) all code that doesn't work on Python2.7
* other changes commented above

### Some open points

* **Deleting instances**: I have removed the `_created_dynamically` flag and respective checks, so also non-dynamically-created instances can be deleted. Should there be rules here? E.g., what happens if **all** instances of a plugin are deleted?
* **Behavior of dynamic instance creation/deletion, and using hotplug**: I kept everything explicit and manual for now. At first glance, using hotplug could make this a little harder to use correctly, to avoid accidentally having and irq assigned to the wrong instance. E.g., if we say when an instance with assigned devices is deleted, we just automatically hotplug the devices to other instances -> What happens if there is more than one suitable instance?
* How do we want to handle (physical) hotplugging of devices, which leads to adding/removal of IRQs in the system? Do we need to periodically scan `/proc/irq` and adapt the devices known to the plugin?
* **Moving devices between plugin instances without running "unapply"**: That is certainly interesting, and I have some ideas, but I think that's for a future PR.
* **Preserving state across daemon restarts**: In general, I think a daemon restart (without changing the profile) should not have an impact on current tuning, so instances and attached devices should be persistent. But this is a bigger topic, and won't be addressed in this PR.

### And one more thing

The goal of this is to be able to tune a device (usually a NIC) to be used exclusively and with low latency by a realtime application. The IRQ affinity is just one part of that. There is also:

* IRQ threads (kernel threads `irq/*`). The kernel automatically migrates them when the IRQ affinity is changed, so we don't need to touch their affinities manually, and in case we need to set their priority I believe we can do that with the scheduler plugin.
* NAPI threads (kernel threads `napi/*`, created for a NIC, depending on the setting in `/sys/class/net/%iface%/threaded`). Those are created `SCHED_OTHER`, and we can use the scheduler plugin to give them a suitable policy and priority. Their default affinity is "all cores", and currently we could use the scheduler plugin to set a static affinity for them. **But we currently can't dynamically set their affinity.**
* Maybe other/future kernel threads that we'd also like to migrate dynamically.

This maybe goes beyond the scope if an "irq plugin", but logically it could fit here. But I haven't thought it through, especially the conflicts with the scheduler plugin. Maybe you have an opinion/idea?

tuned/plugins/plugin_irq.py
zacikpa: I believe you should call
```python
super(IrqPlugin, self)._instance_apply_static(instance)
```
at the beginning of `IrqPlugin._instance_apply_static` to execute all decorated commands (in this case, the custom affinity command). Then, you don't have to manually loop through the list of assigned devices.

zacikpa:
So far I tested the new `irq_process` option in the `scheduler` plugin and it seems to work nice. When doing the final rebase/squash, you should probably add `Resolves: RHEL-21923` into the commit message.

> I have removed the _created_dynamically flag and respective checks, so also non-dynamically-created instances can be deleted. Should there be rules here? E.g., what happens if all instances of a plugin are deleted?

I can't think of a reason why deleting the last plugin instance should be disallowed. Technically, we could also permit creating instances of inactive plugins via the `create_instance` command (though I'm not sure whether that has a real-life use case).

> E.g., if we say when an instance with assigned devices is deleted, we just automatically hotplug the devices to other instances -> What happens if there is more than one suitable instance?

IIUC the [hotplug implementation](https://github.com/redhat-performance/tuned/blob/b15de12ec9673f9cc6c780dc7c03516cac9b8b65/tuned/plugins/hotplug.py#L44) currently assigns new devices to suitable plugin instances based on their `priority` option, as sorted [here](https://github.com/redhat-performance/tuned/blob/b15de12ec9673f9cc6c780dc7c03516cac9b8b65/tuned/units/manager.py#L80), so maybe that code could be re-used (and possibly overrided in the future if we want custom behavior in specific plugins). Although to be able to set the priority of newly created instances, the code would need to be refactored a bit.

yarda:
Fedora rawhide CI failure is caused by regression in the kernel: https://bugzilla.redhat.com/show_bug.cgi?id=2262526

com.redhat.tuned.policy
```diff
@@ -247,4 +247,24 @@
     </defaults>
   </action>
 
+  <action id="com.redhat.tuned.instance_create">
```
yarda, line 250: I think this requires auth_admin otherwise regular user could reconfigure the machine. I.e. the rule of thumb is: any user can list config, but only authorized user (admin) can make changes because it could cause DOS and have impact on the system performance.

yarda: I think both instance_create/instance_destroy should require auth_admin.

yarda (changes requested):

com.redhat.tuned.policy
yarda: Even now when the instance is empty and attach/detach devices requires auth_admin, thus instance_create/destroy shouldn't cause any harm, it's change of the configuration and as such it should require auth_admin.

tuned/plugins/plugin_irq.py
```diff
@@ -0,0 +1,167 @@
...
+
+		affinity = instance.options.get("affinity")
+		affinity_list = self._cmd.cpulist_unpack(affinity)
+		instance._affinity = affinity_list
+		if len(affinity_list) == 0:
+			log.info("Instance '%s' has no affinity. Deactivating." % instance.name)
+			instance._active = False
+
+		if instance._is_main_instance:
+			default_affinity = instance.options.get("default_affinity", "calc")
```
yarda, line 48: I think you could simplify the code, e.g.:
```
instance._default_affinity =  instance.options.get("default_affinity", "calc")
if instance._default_affinity not in ["calc", "ignore"]:
    instance._default_affinity = self._cmd.cpulist_unpack(default_affinity)
```

tuned/plugins/plugin_irq.py
```diff
@@ -0,0 +1,167 @@
...
+
+	def __init__(self, monitor_repository, storage_factory, hardware_inventory, device_matcher, device_matcher_udev, plugin_instance_factory, global_cfg, variables):
+		super(IrqPlugin, self).__init__(monitor_repository, storage_factory, hardware_inventory, device_matcher, device_matcher_udev, plugin_instance_factory, global_cfg, variables)
+		self._irq_original = {}
+		self._irq_unchangeable = []
+
+	def _init_devices(self):
+		self._devices_supported = True
+		self._free_devices = set()
+		self._assigned_devices = set()
```
yarda, line 22: Could you match using udev? Like other device plugins do. I think it could work (I didn't test it), because e.g.:
```
# udevadm info /sys/kernel/irq/59
P: /kernel/irq/59
M: 59
```

tuned/plugins/plugin_irq.py
```diff
@@ -0,0 +1,167 @@
...
+	def _get_config_options(self):
+		return {
+			"affinity": "",
+			"default_affinity": "calc",
+		}
+
+	def _instance_init(self, instance):
+		instance._has_static_tuning = True
+		instance._has_dynamic_tuning = False
+		instance._is_main_instance = instance.name == instance.plugin.name
```
yarda, line 38: I wouldn't do this. Instance name can be arbitrary string.  So maybe you want the first instance?

Do we really need `default_affinity`?

E.g.:
```
[irq]
affinity=1,2,3
```

Or:
```
[irq_a]
type=irq
devices=1,2
affinity=1,2

[irq_b]
devices=*
affinity=3
```

If I am not missing anything, I think we could live without it and simplify the code.

tuned/plugins/plugin_irq.py
```diff
@@ -0,0 +1,167 @@
...
+		if current_affinity_string == desired_affinity_string:
+			log.info(consts.STR_VERIFY_PROFILE_VALUE_OK
+					% (affinity_description, current_affinity_string))
+			return True
+		else:
+			log.error(consts.STR_VERIFY_PROFILE_DEVICE_VALUE_FAIL
+					% (affinity_description, current_affinity_string, desired_affinity_string))
+			return False
+
+	@command_custom("affinity", per_device=True)
```
yarda, line 159: Wouldn't the device set/get fit better?

yarda:
IMHO:
dee5d6f63aab46427ade7415babf3c5ef1488c52 - it's OK for merge now
aad02c76e152bed9aea7a34846a013278db973f8 - tighten the policy (per inline comment) and it's probably OK
17b528c0d2065ff902d6ceb00407569c2a8f2c05 - this still needs some work, see comments

yarda:
>### Some open points
> 
>* **Deleting instances**: I have removed the `_created_dynamically` flag and respective checks, so also non-dynamically-created instances can be deleted. Should there be rules here? E.g., what happens if **all** instances of a plugin are deleted?
> 
As @zacikpa wrote I don't think we need any limitation here. Currently, at runtime I think you won't be able to achieve it, because through the API you can only move devices between instances, thus there will be always at least one instance with some devices which cannot be removed. If you startup with empty instances then you will be able to delete all instances, but I think it shouldn't cause any problem.

>* How do we want to handle (physical) hotplugging of devices, which leads to adding/removal of IRQs in the system? Do we need to periodically scan `/proc/irq` and adapt the devices known to the plugin?
>
This is usually done by udev.  It seems udev can parse the irqs:
```
udevadm info /sys/kernel/irq/59
P: /kernel/irq/59
M: 59
```
But I don't know whether it support hotplug for irqs. If yes we are probably done just by switching to udev enumeration. If not, we probably cannot support hotplug of irqs without periodic rescans and in such case I wouldn't cope with it for now.

>* **Moving devices between plugin instances without running "unapply"**: That is certainly interesting, and I have some ideas, but I think that's for a future PR.
> 
OK, I agree.

>* **Preserving state across daemon restarts**: In general, I think a daemon restart (without changing the profile) should not have an impact on current tuning, so instances and attached devices should be persistent. But this is a bigger topic, and won't be addressed in this PR.
> 
I agree.
> 
>### And one more thing
>
>The goal of this is to be able to tune a device (usually a NIC) to be used exclusively and with low latency by a realtime application. The IRQ affinity is just one part of that. There is also:
>
>* IRQ threads (kernel threads `irq/*`). The kernel automatically migrates them when the IRQ affinity is changed, so we don't need to touch their affinities manually, and in case we need to set their priority I believe we can do that with the scheduler plugin.
>
>* NAPI threads (kernel threads `napi/*`, created for a NIC, depending on the setting in `/sys/class/net/%iface%/threaded`). Those are created `SCHED_OTHER`, and we can use the scheduler plugin to give them a suitable policy and priority. Their default affinity is "all cores", and currently we could use the scheduler plugin to set a static affinity for them. **But we currently can't dynamically set their affinity.**
> 
> * Maybe other/future kernel threads that we'd also like to migrate dynamically.
>
>
> This maybe goes beyond the scope if an "irq plugin", but logically it could fit here. But I haven't thought it through, especially the conflicts with the scheduler plugin. Maybe you have an opinion/idea?

I think we can duplicate the functionality in plugins if necessary and cover the limitations in documentation. Otherwise I don't have more ideas at the moment.

yarda:
> IMHO: 
>[dee5d6f](https://github.com/redhat-performance/tuned/commit/dee5d6f63aab46427ade7415babf3c5ef1488c52) - it's OK for merge now
>[aad02c7](https://github.com/redhat-performance/tuned/commit/aad02c76e152bed9aea7a34846a013278db973f8) - tighten the policy (per inline comment) and it's probably OK 
>[17b528c](https://github.com/redhat-performance/tuned/commit/17b528c0d2065ff902d6ceb00407569c2a8f2c05) - this still needs some work, see comments

Maybe could you split the PR? The first commits is ready, the second almost ready and we would like to have them in the upcoming release which will happen very soon. The last commit (irq plugin) may take some more time.

adriaan42:
> > IMHO:
> > [dee5d6f](https://github.com/redhat-performance/tuned/commit/dee5d6f63aab46427ade7415babf3c5ef1488c52) - it's OK for merge now
> > [aad02c7](https://github.com/redhat-performance/tuned/commit/aad02c76e152bed9aea7a34846a013278db973f8) - tighten the policy (per inline comment) and it's probably OK
> > [17b528c](https://github.com/redhat-performance/tuned/commit/17b528c0d2065ff902d6ceb00407569c2a8f2c05) - this still needs some work, see comments
> 
> Maybe could you split the PR? The first commit is ready, the second almost ready and we would like to have them in the upcoming release which will happen very soon. The last commit (irq plugin) may take some more time.

Yes, the commits are pretty independent and I'm happy to split the PR. I agree that the first one is ready. For the second one, I'm not sure: this currently does not consider hotplugging, and with hotplugging we might need additional parameters for the `instance_create` call, like a `devices_udev_regex` and an instance `priority`. It looks like the priority can be passed as part of the `options`, but I'm not sure about the udev regex. That seems to be a separate argument to the instance creation.

I'll open independent PRs and we can discuss there...

yarda:
> > > IMHO:
> > > [dee5d6f](https://github.com/redhat-performance/tuned/commit/dee5d6f63aab46427ade7415babf3c5ef1488c52) - it's OK for merge now
> > > [aad02c7](https://github.com/redhat-performance/tuned/commit/aad02c76e152bed9aea7a34846a013278db973f8) - tighten the policy (per inline comment) and it's probably OK
> > > [17b528c](https://github.com/redhat-performance/tuned/commit/17b528c0d2065ff902d6ceb00407569c2a8f2c05) - this still needs some work, see comments
> > 
> > 
> > Maybe could you split the PR? The first commit is ready, the second almost ready and we would like to have them in the upcoming release which will happen very soon. The last commit (irq plugin) may take some more time.
> 
> Yes, the commits are pretty independent and I'm happy to split the PR. I agree that the first one is ready. For the second one, I'm not sure: this currently does not consider hotplugging, and with hotplugging we might need additional parameters for the `instance_create` call, like a `devices_udev_regex` and an instance `priority`. It looks like the priority can be passed as part of the `options`, but I'm not sure about the udev regex. That seems to be a separate argument to the instance creation.
> 
> I'll open independent PRs and we can discuss there...

Thanks. Regarding the second PR I thought it's good enough for the basic functionality and could be improved later. But if you want to do it right +1 from me.

adriaan42:
> Thanks. Regarding the second PR I thought it's good enough for the basic functionality and could be improved later. But if you want to do it right +1 from me.

I was worried that we might need to change the dbus interface later to support hotplugging. But I had another look, and I believe we could still pass all required argument in `options` via dbus, and can then extract them just like the `Unit` constructor (https://github.com/redhat-performance/tuned/blob/master/tuned/profiles/unit.py#L12).

So if it's ok for you, we can already use it and refine later. I opened #596, with the updated policies, and a `NOTE` in the code about the current shortcomings.

yarda:
> > Thanks. Regarding the second PR I thought it's good enough for the basic functionality and could be improved later. But if you want to do it right +1 from me.
> 
> I was worried that we might need to change the dbus interface later to support hotplugging. But I had another look, and I believe we could still pass all required argument in `options` via dbus, and can then extract them just like the `Unit` constructor (https://github.com/redhat-performance/tuned/blob/master/tuned/profiles/unit.py#L12).
> 
> So if it's ok for you, we can already use it and refine later. I opened #596, with the updated policies, and a `NOTE` in the code about the current shortcomings.

Thanks for the split. Currently, only the first commit is important for the upcoming release. The second commit is nice to have and I think it can wait, so I will keep it open.

yarda:
> > * How do we want to handle (physical) hotplugging of devices, which leads to adding/removal of IRQs in the system? Do we need to periodically scan `/proc/irq` and adapt the devices known to the plugin?
> 
> This is usually done by udev. It seems udev can parse the irqs:
> 
> ```
> udevadm info /sys/kernel/irq/59
> P: /kernel/irq/59
> M: 59
> ```
> 
> But I don't know whether it support hotplug for irqs. If yes we are probably done just by switching to udev enumeration. If not, we probably cannot support hotplug of irqs without periodic rescans and in such case I wouldn't cope with it for now.
> 
Hmm, I am afraid this will not work with udev, I tried simple reproducer:
```
import pyudev

udev_context = pyudev.Context()
for d in udev_context.list_devices():
  print(d)
```
Unfortunately, it seems it doesn't know the IRQs.

com.redhat.tuned.policy
adriaan42: The calls now require auth_admin (and the feature now has its own separate PR #596)

tuned/plugins/plugin_irq.py
adriaan42: This `default_affinity` is not a default used by TuneD/plugin_irq, but it's actually set in the kernel (`/proc/irq/default_smp_affinity`) and affects newly appearing irqs, even before TuneD sees them. So yes, I'd say we need the setting, and it only makes sense to set it once.

But yes, instead of using the name it's probably better to take the "first instance" approach used by the `cpu` plugin.

tuned/daemon/controller.py
adriaan42: #596 now implements this.

yarda:
Sorry for the delay, but hopefully it's not urgent, although it's very nice RFE.

>     * Determine the "first instance": The way as implemented in the `cpu` plugin does not work (basically the same issue I mentioned in [cpu: add disable_thread_siblings option #366 (comment)](https://github.com/redhat-performance/tuned/pull/366#issuecomment-1738633828)):
>       
>       * I'd like to be able to have a configuration like this:
>         ```ini
>         [irq_default]
>         type=irq
>         affinity=0
>         default_affinity=0
>         devices=*
>         [irq_special]
>         type=irq
>         priority=-1
>         affinity=2
>         devices=irq42
>         ```
>       * So there is a default instance that should handle global/static tuning and all devices not matching any other instance, and then dedicated instance for exceptions.

You can define the priority also with the order change, e.g.:
```ini
[irq_special]
type=irq
affinity=2
devices=irq42

[irq_default]
type=irq
affinity=0
default_affinity=0
devices=*
```

>       * The current implementation in the CPU plugin (which is the only other plugin with a "first instance") always takes the instance with the highest priority (lowest value) as "first instance", but by doing that, this instance is started first, and automatically claims all free devices.
>       * In my current proposal here, I determine the "first instance" by name (`instance._is_main_instance = instance.name == instance.plugin.name`), but I'm open to other suggestions, e.g. an explicit option (`first_instance=True`)...
> 

Do we need the first instance here? Although your approach with the first instance being the instance with the plugin name works, it's a new TuneD paradigm. What about using special device name, e.g. "DEFAULT", then:
```ini
[irq_default]
type=irq
affinity=0
devices=DEFAULT
```
I.e. handle the default SMP affinity as a special IRQ, then it could be matched by the glob '*', so you could write:
```ini
[irq_default]
type=irq
affinity=0
```
To cover all remaining IRQs including the default setting (implicit `devices=*`). In case you don't want the default SMP, which I think you usually want:
```ini
[irq_nondefault]
type=irq
affinity=0
devices=!DEFAULT
```
Then I think you could also use the `calc` parameter in the `affinity` which could simplify things. What do you think?

adriaan42:
> Do we need the first instance here? Although your approach with the first instance being the instance with the plugin name works, it's a new TuneD paradigm. What about using special device name, e.g. "DEFAULT", then:

Good idea! Yes, that works and actually simplifies the code.

> Then I think you could also use the `calc` parameter in the `affinity` which could simplify things. What do you think?

I'm not entirely sure how to handle the special values `ignore`/`calc` present in the scheduler plugin:

* We no longer need `ignore`. An empty affinity list means "don't touch".
* `calc` means "derive affinity from `isolated_cores`". We don't have that variable in the new plugin, and the user now needs to calculate the desired set explicitly. I think that's ok, and there already are functions available to do this (e.g. `cpulist_invert`).
* `calc` also means "intersect the desired affinity with the current one". I assume that's so TuneD does not undo affinity settings made by someone else? Is that still needed? If so, then we could introduce a parameter `mode` which can be `set`, to always apply the affinity value (default), or `intersect` to calculate the new value based on the current one.
  ```ini
  [scheduler]
  isolated_cores=${isolcpus}
  default_irq_smp_affinity=calc
  ```
  would then become
  ```ini
  [irq]
  affinity=${f:cpulist_invert:${isolcpus}}
  devices=!default
  [irqdefault]
  affinity=${f:cpulist_invert:${isolcpus}}
  devices=default
  mode=intersect
  ```

yarda:
>     * We no longer need `ignore`. An empty affinity list means "don't touch".
> 
OK.

>     * `calc` means "derive affinity from `isolated_cores`". We don't have that variable in the new plugin, and the user now needs to calculate the desired set explicitly. I think that's ok, and there already are functions available to do this (e.g. `cpulist_invert`).
>
OK, it's always good to use what's currently available.
 
>     * `calc` also means "intersect the desired affinity with the current one". I assume that's so TuneD does not undo affinity settings made by someone else? Is that still needed? If so, then we could introduce a parameter `mode` which can be `set`, to always apply the affinity value (default), or `intersect` to calculate the new value based on the current one.
>
I think `mode` is OK, I think it's the most flexible approach.

>       ```ini
>       [scheduler]
>       isolated_cores=${isolcpus}
>       default_irq_smp_affinity=calc
>       ```
>       would then become
>       ```ini
>       [irq]
>       affinity=${f:cpulist_invert:${isolcpus}}
>       devices=!default
>       [irqdefault]
>       affinity=${f:cpulist_invert:${isolcpus}}
>       devices=default
>       mode=intersect
>       ```
I think this can be even simplified to:
```ini
[irqdefault]
affinity=${f:cpulist_invert:${isolcpus}}
devices=default
mode=intersect
[irq]
affinity=${f:cpulist_invert:${isolcpus}}
```

Maybe I would use uppercase name i.e. `DEFAULT`, as it is some special (virtual) device. IIRC some special values were previously defined in uppercase in TuneD and IMHO there is also lower probability of the name conflict with some existing device name. But if you like the lowercase variant more, I am not against.

adriaan42:
Pushed updated implementation:

* `default_smp_affinity` is now handled by a virtual device `DEFAULT`, eliminating the need for a "first/main instance".
* instances have a `mode` option: `set` always applies the configured affinity, `intersect` calculates the new affinity as intersection of the current and the configured one, providing behavior of the `calc` setting of the scheduler plugin.

tuned/plugins/plugin_irq.py
```diff
@@ -0,0 +1,269 @@
...
+		Args:
+			irqinfo (IrqInfo): IRQ that should be tuned
+			affinity (set): desired affinity
+		"""
+		instance = self._get_device_instance(irqinfo.device)
+		if instance is None:
+			return
+		original = self._get_irq_affinity(irqinfo.irq)
+		if instance._mode == "intersect":
+			affinity = affinity & original or affinity
```
yarda, line 199: Maybe add comment here for better readability, e.g.:
```
# intersection of affinity and original, affinity if the intersection is empty
```

tuned/plugins/plugin_irq.py
```diff
@@ -0,0 +1,269 @@
...
+			if os.path.isdir(p) and i.isdigit():
+				info = IrqInfo(i)
+				self._irqs[i] = info
+				self._free_devices.add(info.device)
+		default_info = IrqInfo("DEFAULT")
+		default_info.device = "DEFAULT"
+		self._irqs["DEFAULT"] = default_info
+		self._free_devices.add("DEFAULT")
+
+	def _get_device_instance(self, device):
```
yarda, line 82: What about the following:
```python
	@command_custom("mode", per_device = False, priority = -10)
	def _mode(self, enabling, value, verify, ignore_missing):
		if verify:
			return None
		if enabling and value is not None:
			self._mode_val = value
```
Then you could use the `self._mode_val` instead of the `instance._mode` and you could omit the whole `get_device_instance` method. The resulting code will be not only cleaner but will probably also run a bit faster, because it wouldn't need to scan devices to find the current instance and query it for mode.

yarda (changes requested):

tuned/plugins/plugin_irq.py
yarda: I agree the engine could do it better, e.g. by passing the link to instance in the `command_custom` args, but it's for future engine RFEs.

tuned/plugins/plugin_irq.py
```diff
@@ -0,0 +1,269 @@
...
+	# instance-level methods: implement the Instance interface
+	#
+	def _instance_init(self, instance):
+		instance._has_static_tuning = True
+		instance._has_dynamic_tuning = False
+
+		affinity = instance.options.get("affinity")
+		instance._affinity = set(self._cmd.cpulist_unpack(affinity))
+		if len(instance._affinity) == 0:
+			log.info("Instance '%s' has no affinity. Deactivating." % instance.name)
```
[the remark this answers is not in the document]
adriaan42: I've changed this, and we're now distinguishing between "empty affinity" (with an info log) and "non-empty but parsed to empty set", which produces an error, and a better error message.

I also changed the "invalid mode" warning to an error.

tuned/plugins/plugin_irq.py
```diff
@@ -0,0 +1,269 @@
...
+		"""Get current IRQ affinity from the kernel
+
+		Args:
+			irq (str): IRQ number (as string) or "DEFAULT"
+
+		Returns:
+			affinity (set): set of all CPUs that belong to the IRQ affinity mask
+		"""
+		filename = "/proc/irq/default_smp_affinity" if irq == "DEFAULT" else "/proc/irq/%s/smp_affinity" % irq
+		with open(filename, "r") as f:
```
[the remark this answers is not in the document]
adriaan42: Yes. I added a try/catch, and decided to output a log.debug in case of failure. As we're reading only in connection with a write (to get the original value), and we can assume that the write will fail for the same reason, the user will still see the "Failed to set SMP affinity" error.

tuned/plugins/plugin_irq.py
adriaan42: done.

tuned/plugins/plugin_irq.py
adriaan42: I also don't like the `_get_device_instance`, and your suggestion solves that nicely. However, I'm also including the `verify` case, as this also needs to know the mode.

yarda:
Thanks, LGTM.

yarda (approved):

PR #581, opened 2023-12-11
profiles: add energy_performance_preference hints to profiles

ryanabx:
This commit adds 'energy_performance_preference' tunings for some of the premade profiles in TuneD. For more information about energy_performance_preference, see https://www.kernel.org/doc/html/v4.19/admin-guide/pm/intel_pstate.html#energy-vs-performance-hints

Resolves https://github.com/redhat-performance/tuned/issues/563
Supersedes https://github.com/redhat-performance/tuned/pull/565

This PR adds `energy_performance_preference` hints to profiles in TuneD.

Setting EPP hints on systems that support `intel-pstate` and `amd-pstate` drivers can significantly reduce power consumption. For example, `power-profiles-daemon` supports energy_performance_preference hints, but currently suffers from an issue where multiple drivers cannot be used at once, therefore `platform-profile` takes precedence and power savings are lost.

Note: I had made a previous PR on this subject, but didn't read `CONTRIBUTING.md` on how to sign off the commit message. The commit msg should have the proper format now.

### DIFF
--- a/profiles/balanced/tuned.conf
+++ b/profiles/balanced/tuned.conf
@@ -12,6 +12,7 @@ cpufreq_conservative=+r
 priority=10
 governor=conservative|powersave
 energy_perf_bias=normal
+energy_performance_preference=balance_performance
 
 [audio]
 timeout=10
--- a/profiles/powersave/tuned.conf
+++ b/profiles/powersave/tuned.conf
@@ -8,6 +8,7 @@ summary=Optimize for low power consumption
 [cpu]
 governor=ondemand|powersave
 energy_perf_bias=powersave|power
+energy_performance_preference=power
 
 [eeepc_she]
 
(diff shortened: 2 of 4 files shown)


zacikpa:
This looks good. Are we sure that all of the values (`power`, `balance_power`, `performance`, `balance_performance`) are always available in both AMD and Intel? If not, it would be nice to specify fallback values using the '|' syntax, similarly as in #573.

ryanabx:
> This looks good. Are we sure that all of the values (`power`, `balance_power`, `performance`, `balance_performance`) are always available in both AMD and Intel? If not, it would be nice to specify fallback values using the '|' syntax, similarly as in #573.

I believe intel-pstate and amd-pstate both implement the same values.

ryanabx:
@zacikpa I have an intel device for testing now (i5 8250u) and I checked the available energy performance preferences:

image

Looks like we have the same preferences for AMD and Intel!

zacikpa (approved):

yarda:
LGTM, thanks.

PR #583, opened 2023-12-20
Expand sysctl plugin options when reporting overrides

zacikpa:
When applying system sysctl options, we log when the applied option overrides an option from a TuneD config. The overrided option may be provided as a variable, so we have to expand it before comparing it with the applied option.

Resolves: RHEL-18972


yarda:
Thanks, LGTM.

PR #587, opened 2024-01-11
plugin_cpu: Support cstate settings of pm_qos_resume_latency_us

zacikpa:
The `force_latency` option already supports using C-state IDs/names. This commit extends the support to the `pm_qos_resume_latency_us` option, together with the ability to set fallback values (including the special "n/a" value which disables all C-states).

Resolves: RHEL-21129


yarda:
Thanks, LGTM.

PR #589, opened 2024-01-24
Do not report errors on systems with no wifi

jmencak:
This change does not attempt to `_wifi_set_power_level()` unless `/proc/net/wireless` exists.

Other changes: fix a typo.

### DIFF
--- a/profiles/functions
+++ b/profiles/functions
@@ -356,13 +356,16 @@ _wifi_set_power_level() {
 	# 6    disable power savings
 	level=$1
 
+	# do not report errors on systems with no wireless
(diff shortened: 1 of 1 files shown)


jmencak:
Greetings, @zacikpa / @yarda!  I've noticed another unnecessary error when using the `powersave` profile. That one comes from the  `plugin_video.py` plug-in in `_get_radeon_powersave()` directly.  Not sure about the best way of fixing that, so leaving that perhaps as future work those more familiar with the code.

yarda:
LGTM, thanks. The rawhide CI error is caused by regression in the kernel.

PR #590, opened 2024-01-30
Add intel uncore plugin

sgruszka:
Add simplistic intel uncore plugin to so that uncore maximum and minimum frequency can be set.

For example lowering uncore maximum frequency can have significant impact on the total power consummation of SoC.

Refer to the following link to get details:
https://docs.kernel.org/admin-guide/pm/intel_uncore_frequency_scaling.html

When system with TPMI (Topology Aware Register and PM Capsule Interface) is detected only devices configured via TPMI are used, since configuring via deprecated (package/die) interface create conflict on new systems.

Plugin allow to configure frequency per individual uncore units, example profile config options can look like this:

[intel_uncore_all0] # power save
type=intel_uncore
devices=uncore0*
max_freq_khz=2000000

[intel_uncore_10] # performance
type=intel_uncore
devices=uncore10
min_freq_khz=4000000

[intel_uncore] # for all others save power with little performance cost
max_freq_khz=4600000

For systems where maximum/minimum configured frequency is above/below the system limits, values are capped to the limit. Wrong configuration i.e. maximum frequency below minimum frequency is skipped and errors are logged.

TODO:

- Create matching functions in (tuned/profiles/functions) to allow to identify cpus belongs to uncore unit as well uncore TPMI unit based on power domain/package/die.
- Add uncore frequency configuration to basics or specialistic profiles.

### DIFF
--- a/tuned/plugins/plugin_intel_uncore.py
+++ b/tuned/plugins/plugin_intel_uncore.py
@@ -1,0 +1,174 @@
+from . import hotplug
+from .decorators import *
+import tuned.logs
+from tuned.utils.commands import commands
+
+import os
+import fnmatch
+
+log = tuned.logs.get()
+cmd = commands()
+
+SYSFS_DIR = "/sys/devices/system/cpu/intel_uncore_frequency/"
+
+IS_MIN = 0
+IS_MAX = 1
+
+class IntelUncorePlugin(hotplug.Plugin):
+	"""
+	`intel_uncore`::
+
+	`max_freq_khz, min_freq_khz`:::
+	Limit the maximum and minumum uncore frequency.
+	====
+	----
+	[intel_uncore]
+	max_freq_khz=4000000
+	----
+	Using this options *TuneD* will limit maximum frequency of all uncore units
+	in the system to 4 GHz
+	====
+	"""
+
+	def _init_devices(self):
+		self._devices_supported = True
+		self._assigned_devices = set()
+		self._free_devices = set()
+		self._is_tpmi = False
+
+		try:
+			devices = os.listdir(SYSFS_DIR)
+		except OSError:
+			return
+
+		# For new TPMI interface use only uncore devices
+		tpmi_devices = fnmatch.filter(devices, 'uncore*')
+		if len(tpmi_devices) > 0:
+			self._is_tpmi = True  # Not used at present but can be usefull in future
+			devices = tpmi_devices
+
+		for d in devices:
+			self._free_devices.add(d)
+
+		log.debug("devices: %s", str(self._free_devices))
+
+	def _instance_init(self, instance):
+		instance._has_static_tuning = True
+		instance._has_dynamic_tuning = False
+
+	def _instance_cleanup(self, instance):
+		pass
+
+	def _device_module_name(self, device):
+		return "intel_uncore_frequency"
+
+	def _get(self, dev_dir, file):
+		sysfs_file = SYSFS_DIR + dev_dir + "/" + file
+		value = cmd.read_file(sysfs_file)
+		if len(value) > 0:
+			return int(value)
+		return None
+
+	def _set(self, dev_dir, file, value):
+		sysfs_file = SYSFS_DIR + dev_dir + "/" + file
+		if cmd.write_to_file(sysfs_file, "%u" % value):
+			return value
+		return None
+
+	@classmethod
+	def _get_config_options(cls):
+		return {
+			"max_freq_khz": None,
+			"min_freq_khz": None,
+		}
+
+	def _validate_value(self, device, min_or_max, value):
+		try:
+			freq_khz = int(value)
+		except ValueError:
+			log.error("value '%s' is not integer" % value)
+			return None
+
+		try:
+			initial_max_freq_khz = self._get(device, "initial_max_freq_khz")
+			initial_min_freq_khz = self._get(device, "initial_min_freq_khz")
+			max_freq_khz = self._get(device, "max_freq_khz")
+			min_freq_khz = self._get(device, "min_freq_khz")
+		except (OSError, IOError):
+			log.error("fail to read inital uncore frequency values")
+			return None
+
+		if min_or_max == IS_MAX:
+			if freq_khz < min_freq_khz:
+				log.error("%s max_freq_khz %d value below min_freq_khz %d" % (device, freq_khz, min_freq_khz))
+				return None
+
+			if freq_khz > initial_max_freq_khz:
+				log.info("%s: max_freq_khz %d above initial_max_freq_khz - capped to %d" % (device, freq_khz, initial_max_freq_khz))
+				freq_khz = initial_max_freq_khz
+
+		elif min_or_max == IS_MIN:
+			if freq_khz > max_freq_khz:
+				log.error("%s: min_freq_khz %d value above max_freq_khz %d" % (device, freq_khz, max_freq_khz))
+				return None
+
+			if freq_khz < initial_min_freq_khz:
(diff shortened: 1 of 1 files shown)


tuned/plugins/plugin_intel_uncore.py
```diff
@@ -0,0 +1,174 @@
...
+				log.info("%s: min_freq_khz %d below initial_max_freq_khz - capped to %d" % (device, freq_khz, initial_min_freq_khz))
+				freq_khz = initial_min_freq_khz
+
+		else:
+			return None
+
+		return freq_khz
+
+	@command_set("max_freq_khz", per_device = True)
+	def _set_max_freq_khz(self, value, device, sim):
```
yarda, line 125: Please rebase and update all `command_set` to new internal API signature, i.e. the `remove` parameter was added:
```
def _set_max_freq_khz(self, value, device, sim, remove):
```

tuned/plugins/plugin_intel_uncore.py
```diff
@@ -0,0 +1,174 @@
...
+		log.debug("devices: %s", str(self._free_devices))
+
+	def _instance_init(self, instance):
+		instance._has_static_tuning = True
+		instance._has_dynamic_tuning = False
+
+	def _instance_cleanup(self, instance):
+		pass
+
+	def _device_module_name(self, device):
```
yarda, line 62: IMHO this method is unused, so probably it can be dropped. I guess it's copy&paste from the audio plugin, where it is used.

tuned/plugins/plugin_intel_uncore.py
```diff
@@ -0,0 +1,174 @@
...
+			log.error("fail to read inital uncore frequency values")
+			return None
+
+		if min_or_max == IS_MAX:
+			if freq_khz < min_freq_khz:
+				log.error("%s max_freq_khz %d value below min_freq_khz %d" % (device, freq_khz, min_freq_khz))
+				return None
+
+			if freq_khz > initial_max_freq_khz:
+				log.info("%s: max_freq_khz %d above initial_max_freq_khz - capped to %d" % (device, freq_khz, initial_max_freq_khz))
```
yarda, line 107: Maybe add the word 'value', to be consistent with other outputs.

tuned/plugins/plugin_intel_uncore.py
```diff
@@ -0,0 +1,174 @@
...
+			initial_min_freq_khz = self._get(device, "initial_min_freq_khz")
+			max_freq_khz = self._get(device, "max_freq_khz")
+			min_freq_khz = self._get(device, "min_freq_khz")
+		except (OSError, IOError):
+			log.error("fail to read inital uncore frequency values")
+			return None
+
+		if min_or_max == IS_MAX:
+			if freq_khz < min_freq_khz:
+				log.error("%s max_freq_khz %d value below min_freq_khz %d" % (device, freq_khz, min_freq_khz))
```
yarda, line 103: Maybe use '%s:' for the device to be consistent with other outputs.

tuned/plugins/plugin_intel_uncore.py
```diff
@@ -0,0 +1,174 @@
...
+				log.info("%s: max_freq_khz %d above initial_max_freq_khz - capped to %d" % (device, freq_khz, initial_max_freq_khz))
+				freq_khz = initial_max_freq_khz
+
+		elif min_or_max == IS_MIN:
+			if freq_khz > max_freq_khz:
+				log.error("%s: min_freq_khz %d value above max_freq_khz %d" % (device, freq_khz, max_freq_khz))
+				return None
+
+			if freq_khz < initial_min_freq_khz:
+				log.info("%s: min_freq_khz %d below initial_max_freq_khz - capped to %d" % (device, freq_khz, initial_min_freq_khz))
```
yarda, line 116: Maybe add the word 'value' to be consistent with other outputs.

tuned/plugins/plugin_intel_uncore.py
```diff
@@ -0,0 +1,174 @@
...
+		self._free_devices = set()
+		self._is_tpmi = False
+
+		try:
+			devices = os.listdir(SYSFS_DIR)
+		except OSError:
+			return
+
+		# For new TPMI interface use only uncore devices
+		tpmi_devices = fnmatch.filter(devices, 'uncore*')
```
yarda, line 45: Maybe you could use udev for listing of the devices? I.e. similar way how the cpu or disk plugin does. In such way the hotplug will also start working. It seems udev knows about uncores, but I wasn't able to deeper test it. I did very simple udev query for all available devices on my system and I can see:
```
...
Device('/sys/devices/uncore_arb')
Device('/sys/devices/uncore_cbox_0')
Device('/sys/devices/uncore_cbox_1')
Device('/sys/devices/uncore_imc')
...
```

yarda (changes requested):
The main blocker is the `command_set` internal API change (see inline comments).

Maybe you could try to use udev for the enumeration of uncores (inline comment).

Other than that and some minor things (in inline comments) LGTM.

tuned/plugins/plugin_intel_uncore.py
yarda: The `remove` is set to True, if the method is called due to the removal event. In such case the device files may be missing and all E_NOENT errors should be silently ignored.

yarda: If you don't mind race conditions when you can get E_NOENT errors in the log on the deplug events you you can ignore the `remove` parameter in the code, but it has to be defined.

tuned/plugins/plugin_intel_uncore.py
sgruszka: Hi, thanks for review!

> Maybe you could use udev for listing of the devices? I.e. similar way how the cpu or disk plugin do. In such 
case the hotplug will also start working.

Uncore frequency devices are just entries in sysfs directory, not "class" devices that generate udev events out of the box and have udev DEVPATH= . The intel_uncore_frequency.ko drivers support hotplug in that sense it will create sysfs entries when new physical processor is added or remove them when processor box is removed, however FWICT it will not generate any user-space events. I think this have to be improved on the driver side or at least validated.

> It seems udev knows about uncores, but I wasn't able to deeper test it. I did very simple udev query for all available devices on my system and I got:
> 
> ```
> ...
> Device('/sys/devices/uncore_arb')
> Device('/sys/devices/uncore_cbox_0')
> Device('/sys/devices/uncore_cbox_1')
> Device('/sys/devices/uncore_imc')
> ...
> ```

Those are performance monitoring devices, not strictly related to uncore_frequency.

sgruszka:
> The main blocker is the `command_set` internal API change (see inline comments).
> 
> Maybe you could try to use udev for the enumeration of uncores (inline comment).
> 
> Other than that and some minor things (in inline comments) LGTM.

Updated with minor fixes and added remove parameter with no implementation (since hotplug events have to be tested).

yarda:
Thanks for the info. I think it's good enough for the merge. Currently, it isn't used in any upstream profile, thus we can e.g. say it's technology preview in the new TuneD release and improve it later (by e.g. hotplug support).

yarda (approved):

yarda:
@sgruszka what do you think?

sgruszka:
Sorry for late answer, I was on vacation. 

> I thought a bit about it and usually in the past when we added intel specific feature other CPU vendors usually followed sooner or later. So maybe rename the plugin to just "uncore"? It's just RC.1 release now, thus we can still do it quite cheap.

Yes, sure, make sense. I can prepare the name change.

sgruszka:
@yarda crated PR #606 for plugin rename .

PR #591, opened 2024-01-31
tests/beakerlib: Add new test which covers build-in functions.

RHEmployee:


yarda:
@RHEmployee could you fix the CodeQL found issues and could you fix the test?

RHEmployee:
Rhel7 or centos7 is not supported by this test which is reflected in provided metadata.

One issue is preset: wrong detection of centos.

PR #592, opened 2024-01-31
tests/beakerlib: Add new test which covers sysctl options.

RHEmployee:


yarda:
@RHEmployee the test is failing, could you fix it?

PR #593, opened 2024-01-31
tests/beakerlib: Add new test which covers sysctl differently.

RHEmployee:


yarda:
@RHEmployee the test is failing, could you fix it?

RHEmployee:
> @RHEmployee the test is failing, could you fix it?

What if we found some real issue? Should I fix test to ignore it?

This particular test is troublemaker and  I would like to talk with upstream about this.
It seems that sometimes right values are not set. But this seems much better than my local testing.

Is there possibility of race condition in test? Would you please take a look?

Issue #599, opened 2024-02-09
Manually installing tuned from source doesn't work on Fedora 39


superm1:
I originally tried to build an RPM, but I failed due to https://github.com/redhat-performance/tuned/issues/597 and then the follow on issue after that one is solved.  So I shifted over to trying to install from source instead.

It installs fine:
```
$ git checkout https://github.com/redhat-performance/tuned/
$ cd tuned
$ sudo make install 
mkdir -p /usr/local/lib/python3.12/site-packages
mkdir -p /usr/lib/tuned
mkdir -p /var/lib/tuned
mkdir -p /var/log/tuned
mkdir -p /run/tuned
mkdir -p /usr/share/doc/tuned
mkdir -p /etc
mkdir -p /usr/lib/tuned/recommend.d
mkdir -p /etc/tuned/recommend.d
# library
cp -a tuned /usr/local/lib/python3.12/site-packages
# binaries
install -Dm 0755 tuned.py /usr/sbin/tuned && sed -i -r -e '1s|^#!/usr/bin/\<python3\>|#!/usr/bin/python3|' /usr/sbin/tuned && touch -r tuned.py /usr/sbin/tuned;
install -Dm 0755 tuned-adm.py /usr/sbin/tuned-adm && sed -i -r -e '1s|^#!/usr/bin/\<python3\>|#!/usr/bin/python3|' /usr/sbin/tuned-adm && touch -r tuned-adm.py /usr/sbin/tuned-adm;
install -Dm 0755 tuned-gui.py /usr/sbin/tuned-gui && sed -i -r -e '1s|^#!/usr/bin/\<python3\>|#!/usr/bin/python3|' /usr/sbin/tuned-gui && touch -r tuned-gui.py /usr/sbin/tuned-gui;
install -Dpm 0755 systemtap/diskdevstat /usr/sbin/diskdevstat;  install -Dpm 0755 systemtap/netdevstat /usr/sbin/netdevstat;  install -Dpm 0755 systemtap/scomes /usr/sbin/scomes;
install -Dm 0755  systemtap/varnetload  /usr/sbin/varnetload && sed -i -r -e '1s|^#!/usr/bin/\<python3\>|#!/usr/bin/python3|'  /usr/sbin/varnetload && touch -r  systemtap/varnetload  /usr/sbin/varnetload;
# glade
install -Dpm 0644 tuned-gui.glade /usr/share/tuned/ui/tuned-gui.glade
# tools
[...]

PR #600, opened 2024-02-09
Drop `-s` from the launch script shebangs

superm1:
Running `make install` as suggested by `INSTALL`  will place all the python modules in /usr/local which won't be included in the search path when `-s` is used in the shebang.

Closes: https://github.com/redhat-performance/tuned/issues/599

### DIFF
--- a/tuned-ppd.py
+++ b/tuned-ppd.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3 -Es
+#!/usr/bin/python3 -E
 import sys
 import os
 import dbus
--- a/tuned.py
+++ b/tuned.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3 -Es
+#!/usr/bin/python3 -E
 #
 # tuned: daemon for monitoring and adaptive tuning of system devices
 #


yarda:
This seems like python bug. According to the [1]:
> -s Don’t add the [user site-packages directory](https://docs.python.org/3/library/site.html#site.USER_SITE) to [sys.path](https://docs.python.org/3/library/sys.html#sys.path).

From the [2]:
>  Default value is ~/.local/lib/pythonX.Y/site-packages

On my system (fedora 39):
```
import site

print(site.USER_SITE)
```
prints:
```
/home/yarda/.local/lib/python3.12/site-packages
```

I think we generally want the `-s`. I will probably open python bug to sort it out and if it's wanted behavior I will ask them to fix the python documentation.

[1] https://docs.python.org/3/using/cmdline.html#cmdoption-s
[2] https://docs.python.org/3/library/site.html#site.USER_SITE

yarda:
You can workaround it by installation into system sitelib, e.g.:
```
# make install PYTHON_SITELIB=/usr/lib/python3.12/site-packages
```
Or whatever the system sitelib on your distro is.

superm1:
> I think we generally want the -s. I will probably open python bug to sort it out and if it's wanted behavior I will ask them to fix the python documentation.

It's also a bit weird that make install has some stuff ignoring PREFIX entirely and putting things in `/usr` and `/etc`

> You can workaround it by installation into system sitelib, e.g.:

Yeah that works (Fedora 39).

yarda:
> > I think we generally want the -s. I will probably open python bug to sort it out and if it's wanted behavior I will ask them to fix the python documentation.
> 
> It's also a bit weird that make install has some stuff ignoring PREFIX entirely and putting things in `/usr` and `/etc`
>
For `/usr` it's probably bug, that should be fixed, for `etc` it probably isn't. The default prefix for user builds is `/usr/local` for distro builds it's `/usr`. For prefix addition to `/etc` you need to use DESTDIR.
 
> > You can workaround it by installation into system sitelib, e.g.:
> 
> Yeah that works (Fedora 39).

rpmbuild/mock does some magic to set the prefix automatically to `/usr` when packages are built through it.

PR #601, opened 2024-02-11
Add a plugin for amdgpu tuning of the `panel_power_savings` attribute

superm1:
This attribute accepts a range from 0 through 4 where larger values will also have larger panel power savings.

Using this has a trade off for color accuracy, and it is only applied when the system is currently operating on battery.

The plugin uses upower to get a signal when battery changes so that it will react instantly.

Intentionally the plugin will check what values are already programmed to the sysfs file to avoid unnecessary writes.  Writing the sysfs file will cause a modeset which isn't necessary if writing the same value twice.

The default values are applied to the profiles that are used in power-profiles-daemon compatbility.  They also match the values used in that software.

### DIFF
--- a/tuned/plugins/plugin_video.py
+++ b/tuned/plugins/plugin_video.py
@@ -115,6 +115,9 @@ class VideoPlugin(base.Plugin):
 	@command_get("radeon_powersave")
 	def _get_radeon_powersave(self, device, ignore_missing = False):
 		sys_files = self._files(device)
+		if not os.path.exists(sys_files["method"]):
+			log.debug("radeon_powersave is not supported on '%s'" % device)
+			return None
 		method = self._cmd.read_file(sys_files["method"], no_error=ignore_missing).strip()
 		if method == "profile":
 			return self._cmd.read_file(sys_files["profile"]).strip()
--- a/tuned/plugins/plugin_video.py
+++ b/tuned/plugins/plugin_video.py
@@ -69,7 +69,7 @@ class VideoPlugin(base.Plugin):
 	def _instance_cleanup(self, instance):
 		pass
 
-	def _radeon_powersave_files(self, device):
+	def _files(self, device):
 		return {
 			"method" : "/sys/class/drm/%s/device/power_method" % device,
 			"profile": "/sys/class/drm/%s/device/power_profile" % device,
@@ -78,7 +78,7 @@ class VideoPlugin(base.Plugin):
 
 	@command_set("radeon_powersave", per_device=True)
 	def _set_radeon_powersave(self, value, device, sim, remove):
-		sys_files = self._radeon_powersave_files(device)
+		sys_files = self._files(device)
 		va = str(re.sub(r"(\s*:\s*)|(\s+)|(\s*;\s*)|(\s*,\s*)", " ", value)).split()
 		if not os.path.exists(sys_files["method"]):
 			if not sim:
@@ -114,7 +114,7 @@ class VideoPlugin(base.Plugin):
 
 	@command_get("radeon_powersave")
 	def _get_radeon_powersave(self, device, ignore_missing = False):
-		sys_files = self._radeon_powersave_files(device)
+		sys_files = self._files(device)
 		method = self._cmd.read_file(sys_files["method"], no_error=ignore_missing).strip()
 		if method == "profile":
 			return self._cmd.read_file(sys_files["profile"]).strip()
--- a/tuned/plugins/plugin_video.py
+++ b/tuned/plugins/plugin_video.py
@@ -82,7 +82,7 @@ class VideoPlugin(base.Plugin):
 		va = str(re.sub(r"(\s*:\s*)|(\s+)|(\s*;\s*)|(\s*,\s*)", " ", value)).split()
 		if not os.path.exists(sys_files["method"]):
 			if not sim:
-				log.warn("radeon_powersave is not supported on '%s'" % device)
+				log.debug("radeon_powersave is not supported on '%s'" % device)
 				return None
 		for v in va:
 			if v in ["default", "auto", "low", "mid", "high"]:
--- a/tuned/plugins/plugin_video.py
+++ b/tuned/plugins/plugin_video.py
@@ -11,9 +11,10 @@ log = tuned.logs.get()
 class VideoPlugin(base.Plugin):
 	"""
 	`video`::
-	
-	Sets various powersave levels on video cards. Currently, only the
-	Radeon cards are supported. The powersave level can be specified
+
+	Sets various power saving features on video cards.
+	Radeon cards are supported.
+	The powersave level can be specified
 	by using the [option]`radeon_powersave` option. Supported values are:
 	+
 	--
@@ -40,17 +41,27 @@ class VideoPlugin(base.Plugin):
 	radeon_powersave=high
 	----
 	====
+
+	Mobile hardware with amdgpu driven eDP panels can be configured
+	with the [option]`panel_power_savings` option.
+	This accepts a value range from 0 to 4, where 4 is the highest power savings
+	but will trade off color accuracy.
 	"""
+
+	def __init__(self, *args, **kwargs):
+		super(VideoPlugin, self).__init__(*args, **kwargs)
 
 	def _init_devices(self):
 		self._devices_supported = True
 		self._free_devices = set()
 		self._assigned_devices = set()
 
-		# FIXME: this is a blind shot, needs testing
-		for device in self._hardware_inventory.get_devices("drm").match_sys_name("card*").match_property("DEVTYPE", "drm_minor"):
-			self._free_devices.add(device.sys_name)
-
+		# Add any radeon and amdgpu hardware with /any/ supported attributes present
+		for device in self._hardware_inventory.get_devices("drm").match_sys_name("card*-*"):
+			attrs = self._files(device.sys_name)
+			for attr in attrs:
+				if os.path.exists(attrs[attr]):
+					self._free_devices.add(device.sys_name)
 		self._cmd = commands()
 
 	def _get_device_objects(self, devices):
@@ -60,6 +71,7 @@ class VideoPlugin(base.Plugin):
 	def _get_config_options(self):
 		return {
 			"radeon_powersave" : None,
+			"panel_power_savings": None,
 		}
 
 	def _instance_init(self, instance):
@@ -73,8 +85,30 @@ class VideoPlugin(base.Plugin):
 		return {
 			"method" : "/sys/class/drm/%s/device/power_method" % device,
 			"profile": "/sys/class/drm/%s/device/power_profile" % device,
-			"dpm_state": "/sys/class/drm/%s/device/power_dpm_state" % device
+			"dpm_state": "/sys/class/drm/%s/device/power_dpm_state" % device,
+			"panel_power_savings": "/sys/class/drm/%s/amdgpu/panel_power_savings" % device,
 		}
+
+	def apply_panel_power_saving_target(self, device, target, sim=False):
+		"""Apply the target value to the panel_power_savings file if it doesn't already have it"""
+
+		# if we don't have the file, we might be radeon not amdgpu
+		if not os.path.exists(self._files(device)["panel_power_savings"]):
+			return None
+
+		# make sure the value is different (avoids unnecessary kernel modeset)
+		current = int(self._get_panel_power_savings(device))
+		if current == target:
+			log.info(
+				"panel_power_savings for %s already %s" % (device, target)
+			)
+			return target
+
+		# flush it out
+		log.info("%s panel_power_savings -> %s" % (device, target))
+		if sim or self._cmd.write_to_file(self._files(device)["panel_power_savings"], target):
+			return target
+		return None
 
 	@command_set("radeon_powersave", per_device=True)
 	def _set_radeon_powersave(self, value, device, sim, remove):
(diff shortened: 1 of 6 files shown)


tuned/plugins/plugin_amdgpu.py
```diff
@@ -0,0 +1,124 @@
...
+            return None
+
+        # determine value to use (only apply if on battery, otherwise set to 0)
+        if self._on_battery:
+            target = self.target_value
+        else:
+            target = 0
+
+        # make sure the value is different (avoids unnecessary kernel modeset)
+        current = int(self._get_panel_power_savings(device))
```
zacikpa, line 91: This fails when the file `/sys/class/drm/card1-eDP-1/amdgpu/panel_power_savings` does not exist.

yarda:
> I'm not sure what I think about enabling this by default for the main stock profiles, mainly due to the new `upower` dependency and the fact that it basically performs a kind of "dynamic tuning", which TuneD has now [abandoned by default](https://github.com/redhat-performance/tuned/pull/603). @yarda, what are your thoughts on this?

Regarding "dynamic tuning" it's not a blocker, because I think we need per plugin configuration, e.g. something like:
```
tuned-main.conf:
dynamic_tuning=no, yes:amdgpu, yes:scheduler
```

This was discussed earlier, but never implemented :). This is for another PR. Regarding the `scheduler` plugin, the dynamic tuning should do the same what currently the `runtime=1` hack do.

yarda:
I think it should reflect the `enable_dbus`  main conf and `--no-dbus` command line option and also work without DBus (e.g. with limited functionality or with plugin disabled), because there are products which runs by default without DBus, e.g. OCP and this shouldn't break it.

yarda:
> I'm not sure what I think about enabling this by default for the main stock profiles, mainly due to the new `upower` dependency

Regarding the upower, it should detect whether upower is available and if yes use it. This is what other plugin do (e.g. disk plugin with hdparm) then we could softdep upower and it will not break consumers without upower.

yarda:
Also what about adding this functionality into the `video` plugin?

superm1:
> > I'm not sure what I think about enabling this by default for the main stock profiles, mainly due to the new `upower` dependency
> 
> Regarding the upower, it should detect whether upower is available and if yes use it. This is what other plugin do (e.g. disk plugin with hdparm) then we could softdep upower and it will not break consumers without upower.

If it's acceptable I would prefer to make power mandatory for the plugin. If it's not available or masked then fail the plugin initialization/load or make it a noop.

I'll double check behavior for the plugin with it masked to make sure I got it right.

superm1:
> Also what about adding this functionality into the `video` plugin?

I was thinking dedicated amdgpu plugin with the idea that we could extend it to other amdgpu stuff later maybe.

superm1:
> I think it should reflect the `enable_dbus`  main conf and `--no-dbus` command line option and also work without DBus (e.g. with limited functionality or with plugin disabled), because there are products which runs by default without DBus, e.g. OCP and this shouldn't break it.

I think plugin should be disabled for this case.

tuned/plugins/plugin_amdgpu.py
superm1: But doesn't the plugin fail to add any devices?

zacikpa: Oh, right, that should suffice when implemented correctly. Currently, you do:
```python
self._hardware_inventory.get_devices("drm")
            .match_sys_name("card*eDP*")
            .match_sys_name("amdgpu")
            .match_sys_name("panel_power_savings")
```
Applying `match_sys_name` multiple times yields all devices that match at least one of the names (i.e., it's a union, not an intersection). In my case, `card1-eDP-1` gets matched and added to plugin devices.

superm1: Ah! Got it.  I've pushed a change to only match the eDP and add the device if the sysfs file is found for it.

superm1:
> > I think it should reflect the `enable_dbus`  main conf and `--no-dbus` command line option and also work without DBus (e.g. with limited functionality or with plugin disabled), because there are products which runs by default without DBus, e.g. OCP and this shouldn't break it.
> 
> I think plugin should be disabled for this case.

I believe the modifications I've made will effectively cause this behavior if upower can't be accessed for any reason (such as dbus is disabled).  Let me know if you disagree.

zacikpa (commented):
Thanks for the update. Now, this seems to work as expected (i.e., no-op) on machines without AMD GPUs or without upower on the dbus. I'll try to find an AMD GPU machine to test it fully.

yarda:
> > > Also what about adding this functionality into the `video` plugin?
> > 
> > 
> > I was thinking dedicated amdgpu plugin with the idea that we could extend it to other amdgpu stuff later maybe.
> 
> The current functionality of the `video` plugin is also AMD-specific.

I think it could be good extension of the video plugin, because it currently doesn't have dynamic tuning and maybe easier and more user friendly to write more generic:
```
[video]
...
```
than HW specific:
```
[video]
...
[amdgpu]
...
```

superm1:
For now I've force pushed using the old configuration option name and kept the existing style for that plugin.  In general though, I think it would be better to reformat it with `black` if you're open to it.

superm1:
ping?

superm1:
The important part is that it is a mobile design with Vega or newer. I tested it on Phoenix and Rembrandt but I would expect it to be exposed on Renoir or newer.

superm1:
One more thing:

Backport these 2 kernel patches from drm-next to 6.8 kernels to enable access to ABM:

* https://gitlab.freedesktop.org/agd5f/linux/-/commit/63d0b87213a0ba241b3fcfba3fe7b0aed0cd1cc5
* https://gitlab.freedesktop.org/agd5f/linux/-/commit/040fdcde288a2830edc31dd507963d6aadf990d2

They'll be part of 6.9-rc1.

tuned/plugins/plugin_video.py
```diff
@@ -124,3 +199,27 @@ def _get_radeon_powersave(self, device, ignore_missing = False):
...
+		try:
+			value = int(value, 10)
+		except ValueError:
+			if not sim:
+				log.warn("Invalid value %s for panel_power_savings" % value)
+			return None
+		if value in range(0, 5):
+			self.target_value = value
+			if not sim:
+				return self.apply_panel_power_saving_target(device)
```
zacikpa, line 215: IIUC, this will set up the signaling when handling the first device and subsequently apply the settings for *all* devices because of the loop in `upower_changed`. For the remaining devices, it's then essentially a no-op because they are already configured.

I think that's quite inelegant, but it shouldn't cause issues. Alternatively, you could remove the manual call to `upower_changed` and [not return ](https://github.com/redhat-performance/tuned/pull/601/files#diff-d503f51fb415d704c70263a4c6d4eaf2c2def141d988346452ab610d242a9da5R125)after setting up the signalling (which I guess may require an additional error check).

tuned/plugins/plugin_video.py
```diff
@@ -124,3 +199,27 @@ def _get_radeon_powersave(self, device, ignore_missing = False):
 			return "dpm-" + self._cmd.read_file(sys_files["dpm_state"]).strip()
 		else:
 			return None
+
+	@command_set("panel_power_savings", per_device=True)
+	def _set_panel_power_savings(self, value, device, sim, remove):
```
zacikpa, line 204: With `sim=True`, the function should do precisely what it would do with `sim=False` except for actually overwriting the file. It can still print warnings and more importantly, it should return the same value. Currently, it always returns `None` if `sim` is `True`.

zacikpa (commented):
I still haven't been able to test this, because I can't easily patch the kernel on the AMD GPU machines that I can access.

Nonetheless, I have a couple of comments.

tuned/plugins/plugin_video.py
superm1: Fixed

tuned/plugins/plugin_video.py
superm1: I did test it on a two GPU system where both GPUs can connect to a mux for eDP and that's why it has the logic it has now; to ensure that you don't have a corner case.  Let me know if you still want it changed.

superm1:
> I still haven't been able to test this, because I can't easily patch the kernel on the AMD GPU machines that I can access.

The patches are going to be in 6.9-rc1 if that makes it easier.

Here is what it looks like when you boot with power plugged in, unplug and then replug:

```
 sudo ./tuned.py  2>&1 | grep amdgpu
2024-03-18 07:53:49,163 INFO     tuned.plugins.base: instance amdgpu: assigning devices card0-eDP-1
2024-03-18 07:53:49,537 INFO     tuned.plugins.plugin_amdgpu: panel_power_savings already 0 [🔋: False]


2024-03-18 07:53:54,324 INFO     tuned.plugins.plugin_amdgpu: panel_power_savings -> 3 [🔋: True]


2024-03-18 07:53:58,661 INFO     tuned.plugins.plugin_amdgpu: panel_power_savings -> 0 [🔋: False]
```

tuned/plugins/plugin_video.py
zacikpa: I believe it can stay as it is now.

tuned/plugins/plugin_video.py
```diff
@@ -69,20 +108,53 @@ def _instance_init(self, instance):
...
+		"""Apply the target value to the panel_power_savings file if it doesn't already have it"""
+
+		# if we don't have a proxy, we can't tell if we're on battery
+		if not self.proxy:
+			self.setup_battery_signaling()
+			return None
+
+		# if we don't have the file, we might be radeon not amdgpu
+		if not os.path.exists(self._files(device)["panel_power_savings"]):
+			return None
```
zacikpa, line 129: I would swap the first two checks to make sure we don't setup the signaling at all if no device supports `panel_power_savings`: 
```suggestion
		# if we don't have the file, we might be radeon not amdgpu
		if not os.path.exists(self._files(device)["panel_power_savings"]):
			return None

		# if we don't have a proxy, we can't tell if we're on battery
		if not self.proxy:
			self.setup_battery_signaling()
			return None
```

zacikpa (changes requested):
Thanks for the update. I've got one more change request and when that is fixed, this looks mergeable to me.

tuned/plugins/plugin_video.py
```diff
@@ -124,3 +199,24 @@ def _get_radeon_powersave(self, device, ignore_missing = False):
...
+			self.target_value = value
+			return self.apply_panel_power_saving_target(device, sim)
+		else:
+			log.warn("Invalid value %s for panel_power_savings" % value)
+		return None
+
+	@command_get("panel_power_savings")
+	def _get_panel_power_savings(self, device, ignore_missing=False):
+		"""Get the current panel_power_savings value"""
+		fname = self._files(device)["panel_power_savings"]
```
zacikpa, line 221: Check that the file exists the same way as when attempting to set it. This is necessary for the `tuned-adm verify` command to succeed when the device does not support `panel_power_savings`.
```suggestion
		"""Get the current panel_power_savings value"""
		if not os.path.exists(self._files(device)["panel_power_savings"]):
			log.debug("panel_power_savings is not supported on '%s'" % device)
			return None
		fname = self._files(device)["panel_power_savings"]
```

zacikpa (commented):
Managed to find one other thing still.

tuned/plugins/plugin_video.py
superm1: Thx, looks good to me.

zacikpa (approved):
Thanks. LGTM, so approving, but leaving the final merge decision on @yarda.

superm1:
@yarda ping?

zacikpa (changes requested):
As discussed in #616, the response to battery/AC changes should be handled by tuned-ppd and thus moved to that PR.

tuned/plugins/plugin_video.py
```diff
@@ -5,15 +5,21 @@
...
 import errno
 import re
+import dbus
 
 log = tuned.logs.get()
 
+UPOWER_DBUS_NAME = "org.freedesktop.UPower"
+UPOWER_DBUS_PATH = "/org/freedesktop/UPower"
+UPOWER_DBUS_INTERFACE = "org.freedesktop.UPower"
+
```
zacikpa, line 15:
```suggestion
```

tuned/plugins/plugin_video.py
```diff
@@ -5,15 +5,21 @@
 import os
 import errno
 import re
+import dbus
```
zacikpa, line 8:
```suggestion
```

tuned/plugins/plugin_video.py
```diff
@@ -40,17 +46,28 @@ class VideoPlugin(base.Plugin):
...
+
+	Mobile hardware with amdgpu driven eDP panels can be configured
+	with the [option]`panel_power_savings` option.
+	This accepts a value range from 0 to 4, where 4 is the highest power savings
+	but will trade off color accuracy.
 	"""
 
+	def __init__(self, *args, **kwargs):
+		super(VideoPlugin, self).__init__(*args, **kwargs)
+		self.proxy = None
```
zacikpa, line 58:
```suggestion
```

zacikpa (changes requested):
Thanks, please delete the remnants of the dbus functionality according to the comments.

zacikpa (approved):

superm1:
> Thanks, please remove the upower mention from the commit https://github.com/redhat-performance/tuned/commit/9d108694d623fb7d32c7dde40881041e215ff6ce description:

Rebased on main and done now.

yarda (approved):

yarda:
Thanks.

PR #606, opened 2024-02-22
Uncore rename

sgruszka:
Rename the plugin name as so can be used as placeholder for Uncore
settings for other CPU vendors


yarda:
Thanks, LGTM, I think it will deserve tuned-2.22.1 release.

Please next time squash commits.

yarda:
Regarding CI problem on f40 it's probably some CI issue.

PR #607, opened 2024-02-22
Fix running the CPU plugin with procfs from pypi

superm1:
Using the latest procfs from pypi fails because it doesn't have a cpuinfo() member.  Catch these exceptions and try to more gracefully handle them instead of a traceback from the daemon.

Link: https://pypi.org/project/procfs/

### DIFF
--- a/tuned/plugins/plugin_cpu.py
+++ b/tuned/plugins/plugin_cpu.py
@@ -246,7 +246,11 @@ class CPULatencyPlugin(hotplug.Plugin):
 			# Possible other x86 vendors (from arch/x86/kernel/cpu/*):
 			# "CentaurHauls", "CyrixInstead", "Geode by NSC", "HygonGenuine", "GenuineTMx86",
 			# "TransmetaCPU", "UMC UMC UMC"
-			cpu = procfs.cpuinfo()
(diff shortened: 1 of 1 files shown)


superm1:
BTW - this was on Ubuntu 22.04, I ran into this problem.  I didn't  have anything natively installed for procfs which is why I went to pypi.

If I `sudo apt install python3-linux-procfs` that version seems to work.
```
$ apt policy python3-linux-procfs
python3-linux-procfs:
  Installed: 0.6.3-1.1build1
  Candidate: 0.6.3-1.1build1
  Version table:
 *** 0.6.3-1.1build1 500
        500 http://us.archive.ubuntu.com/ubuntu jammy/universe amd64 Packages
        100 /var/lib/dpkg/status
```

But nonetheless it's better to be defensive against this problem IMO.

evgeni:
FWIW, the `tuned` package in Debian (and Ubuntu) depends on `python3-linux-procfs`.
I don't think `procfs` on pypi is related to the one in that package at all.

yarda:
I think if we want to handle this, we should detect the wrong library at the import and log error or warning that it's incompatible library.

PR #609, opened 2024-03-05
Add the option to specify custom directories with user-defined profiles

zacikpa:
The new default directory for user-defined profiles is `/etc/tuned/profiles/`; other directories can be specified via the `custom_profile_dirs` option in `tuned-main.conf`.

When updating the TuneD package, existing profiles in `/etc/tuned/` are automatically copied to `/etc/tuned/profiles/`, unless there are conflicts.

GUI also needed to be updated - I only did the bare minimum for it to work, more major refactoring deserves a separate PR.

### DIFF
--- a/doc/manual/modules/performance/con_the-location-of-tuned-profiles.adoc
+++ b/doc/manual/modules/performance/con_the-location-of-tuned-profiles.adoc
@@ -6,10 +6,19 @@
 *TuneD* stores profiles in the following directories:
 
 [filename]`/usr/lib/tuned/`::
-Distribution-specific profiles are stored in the  directory. Each profile has its own directory. The profile consists of the main configuration file called `tuned.conf`, and optionally other files, for example helper scripts.
+Distribution-specific profiles are stored in the [filename]`/usr/lib/tuned/` directory. Each profile has its own directory. The profile consists of the main configuration file called `tuned.conf`, and optionally other files, for example helper scripts.
 
 [filename]`/etc/tuned/`::
-If you need to customize a profile, copy the profile directory into the  directory, which is used for custom profiles. If there are two profiles of the same name, the custom profile located in [filename]`/etc/tuned/` is used.
+If you need to customize a profile, copy the profile directory into the [filename]`/etc/tuned/` directory, which is used for custom profiles, and then adjust it. If there is a system profile and a custom profile of the same name, the custom profile located in [filename]`/etc/tuned/` is used.
+
+.User-defined profile directories
+====
+If you want to make TuneD load profiles from a directory other than [filename]`/usr/lib/tuned/` and [filename]`/etc/tuned/`, you can list it in [filename]`/etc/tuned/tuned-main.conf` as follows:
+----
+profile_dirs=/usr/lib/tuned,/etc/tuned,/my/custom/profiles
+----
+In this example, profiles are loaded also from [filename]`/my/custom/profiles/`. If two directories contain profiles with the same names, the one that is listed later takes precedence.
+====
 
 [role="_additional-resources"]
 .Additional resources
--- a/tuned-main.conf
+++ b/tuned-main.conf
@@ -82,3 +82,8 @@ log_file_max_size = 1MB
 # - not_on_exit: rollbacks are always performed on a profile
 #   switch, but not on any kind of TuneD process exit
 # rollback = auto
+
+# Directories to search for profiles separated by , or ;
+# In case of conflicts in profile names, the later directory
+# takes precedence
+# profile_dirs = /usr/lib/tuned,/etc/tuned
--- a/tuned/consts.py
+++ b/tuned/consts.py
@@ -13,7 +13,7 @@ DBUS_INTERFACE = "com.redhat.tuned.control"
 DBUS_OBJECT = "/Tuned"
 DEFAULT_PROFILE = "balanced"
 DEFAULT_STORAGE_FILE = "/run/tuned/save.pickle"
-LOAD_DIRECTORIES = ["/usr/lib/tuned", "/etc/tuned"]
+SYSTEM_PROFILE_DIR = "/usr/lib/tuned"
 PERSISTENT_STORAGE_DIR = "/var/lib/tuned"
 PLUGIN_MAIN_UNIT_NAME = "main"
 # Magic section header because ConfigParser does not support "headerless" config
@@ -122,6 +122,7 @@ CFG_UNIX_SOCKET_PERMISIONS = "unix_socket_permissions"
 CFG_UNIX_SOCKET_CONNECTIONS_BACKLOG = "connections_backlog"
 CFG_CPU_EPP_FLAG = "hwp_epp"
 CFG_ROLLBACK = "rollback"
+CFG_PROFILE_DIRS = "profile_dirs"
 
 # no_daemon mode
 CFG_DEF_DAEMON = True
@@ -171,6 +172,8 @@ CFG_DEF_UNIX_SOCKET_CONNECTIONS_BACKLOG = "1024"
 CFG_FUNC_UNIX_SOCKET_CONNECTIONS_BACKLOG = "getint"
 # default rollback strategy
 CFG_DEF_ROLLBACK = "auto"
+# default profile directories
+CFG_DEF_PROFILE_DIRS = [SYSTEM_PROFILE_DIR, "/etc/tuned"]
 
 PATH_CPU_DMA_LATENCY = "/dev/cpu_dma_latency"
 
--- a/tuned/gtk/gui_profile_loader.py
+++ b/tuned/gtk/gui_profile_loader.py
@@ -60,7 +60,7 @@ class GuiProfileLoader(object):
 
         profilePath = self._locate_profile_path(profile_name)
 
-        if profilePath == tuned.consts.LOAD_DIRECTORIES[1]:
+        if profilePath != tuned.consts.SYSTEM_PROFILE_DIR:
             file_path = profilePath + '/' + profile_name + '/' + tuned.consts.PROFILE_FILE
             config_parser = ConfigParser(delimiters=('='), inline_comment_prefixes=('#'), strict=False)
             config_parser.optionxform = str
@@ -127,7 +127,15 @@ class GuiProfileLoader(object):
         self._load_all_profiles()
 
     def save_profile(self, profile):
-        path = tuned.consts.LOAD_DIRECTORIES[1] + '/' + profile.name
+        # save the new profile to a non-system directory with the highest priority
+        path = None
+        for d in reversed(self.directories):
+            if d != tuned.consts.SYSTEM_PROFILE_DIR:
+                path = os.path.join(d, profile.name)
+                break
+        if path is None:
+            raise managerException.ManagerException('Cannot save profile to a system directory')
+
         config = {
             'main': collections.OrderedDict(),
             'filename': path + '/' + tuned.consts.PROFILE_FILE,
@@ -160,7 +168,7 @@ class GuiProfileLoader(object):
             raise managerException.ManagerException('Profile: '
                     + old_profile_name + ' is not in profiles')
 
-        path = tuned.consts.LOAD_DIRECTORIES[1] + '/' + profile.name
+        path = os.path.join(self._locate_profile_path(old_profile_name), profile.name)
 
         if old_profile_name != profile.name:
             self.remove_profile(old_profile_name, is_admin=is_admin)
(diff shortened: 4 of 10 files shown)


tuned/consts.py
```diff
@@ -122,6 +122,7 @@
 CFG_UNIX_SOCKET_CONNECTIONS_BACKLOG = "connections_backlog"
 CFG_CPU_EPP_FLAG = "hwp_epp"
 CFG_ROLLBACK = "rollback"
+CFG_CUSTOM_PROFILE_DIRS = "custom_profile_dirs"
```
jmencak, line 125: As mentioned above, it might be a good idea to change `_dirs` to `_paths` due to consistency with other `tuned-main.conf` options.

jmencak (commented):
Thank you for this PR, Pavol, I very much appreciate it!  I tested the following worked::
* profile migration script
* specifying custom directories in `tuned-main.conf` -- profiles applied from custom dirs

I did not test the GUI though.

The code looks clean to me.  I found a couple of nits and the only real suggestion for improvement I have to add the new option commented out in `tuned-main.conf`

tuned.spec
```diff
@@ -329,6 +329,13 @@ if [ -r "%{_sysconfdir}/default/grub" ]; then
     %{_sysconfdir}/default/grub
 fi
 
+# move all user-defined profiles from /etc/tuned/ to /etc/tuned/profiles/
```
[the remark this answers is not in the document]
zacikpa: Agreed,  changed.

doc/manual/modules/performance/con_the-location-of-tuned-profiles.adoc
```diff
@@ -8,8 +8,18 @@
...
-[filename]`/etc/tuned/`::
-If you need to customize a profile, copy the profile directory into the  directory, which is used for custom profiles. If there are two profiles of the same name, the custom profile located in [filename]`/etc/tuned/` is used.
+[filename]`/etc/tuned/profiles`::
+If you need to customize a profile, copy the profile directory into the  directory, which is used for custom profiles. If there are two profiles of the same name, the custom profile located in [filename]`/etc/tuned/profiles` is used.
+
+.User defined profile directories
+====
+If you want to make TuneD load customized profiles from directories other than [filename]`/etc/tuned/profiles`, you can list them in [filename]`/etc/tuned/tuned-main.conf` as follows:
+----
+custom_profile_dirs=/my/custom/profiles,/etc/tuned/profiles/
```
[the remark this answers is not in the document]
zacikpa: Thanks, fixed.

doc/manual/modules/performance/con_the-location-of-tuned-profiles.adoc
```diff
@@ -8,8 +8,18 @@
...
 Distribution-specific profiles are stored in the  directory. Each profile has its own directory. The profile consists of the main configuration file called `tuned.conf`, and optionally other files, for example helper scripts.
 
-[filename]`/etc/tuned/`::
-If you need to customize a profile, copy the profile directory into the  directory, which is used for custom profiles. If there are two profiles of the same name, the custom profile located in [filename]`/etc/tuned/` is used.
+[filename]`/etc/tuned/profiles`::
+If you need to customize a profile, copy the profile directory into the  directory, which is used for custom profiles. If there are two profiles of the same name, the custom profile located in [filename]`/etc/tuned/profiles` is used.
+
+.User defined profile directories
+====
+If you want to make TuneD load customized profiles from directories other than [filename]`/etc/tuned/profiles`, you can list them in [filename]`/etc/tuned/tuned-main.conf` as follows:
```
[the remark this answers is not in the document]
zacikpa: Done.

tuned.spec
```diff
@@ -329,6 +329,13 @@ if [ -r "%{_sysconfdir}/default/grub" ]; then
...
+for f in %{_sysconfdir}/tuned/*; do
+  if [ -f "$f/tuned.conf" ] && [ ! -d "%{_sysconfdir}/tuned/profiles/`basename $f`" ]; then
```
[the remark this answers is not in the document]
zacikpa: I agree with all three points, updated to use just `-e`.

tuned/consts.py
zacikpa: Here I would disagree. The documentation/manual pages always refer to (custom) profile directories, not paths, so directories seem more clear to me.

zacikpa:
Thanks, @jmencak, for the thorough review. I now updated the PR.

jmencak:
Hi @zacikpa , thank you for the effort, I really appreciate it.  I've just realized a blocker for this from OpenShift side.  While the migration script will work on RHEL and on upgrades, it will not work in container environments where the profiles are still expected in /etc/tuned and the supervising process is unpacking them to /etc/tuned.  An FDP release of this would immediately break all OCP versions which used FDP (OCP 4.10+).  There is unfortunately no way to adjust the supervising process which extracts TuneD profiles all the way down to 4.10.  I do not have a good solution for this at the moment apart from still supporting the obsoleted /etc/tuned for the profiles.  One idea I have could be to ship with the following `tuned-main.conf` file:

```
custom_profile_dirs=/etc/tuned/profiles,/etc/tuned
```

While I do understand this is not clean from the TuneD/RHEL perspective, I currently do not have a better solution to this for the OCP environment.  Thoughts? cc @yarda

zacikpa:
A more extreme solution would be to keep `/etc/tuned/` as the default directory but still merge the changes that would allow users to configure it.

jmencak:
> A more extreme solution would be to keep `/etc/tuned/` as the default directory but still merge the changes that would allow users to configure it.

Yes, this is also a possibility with some INFO (or even WARNING) message that the support for profile directory '/etc/tuned' will be dropped in future versions. Not sure which approach is better to be honest.  I'll let you/Jarda to decide.

yarda:
There will be probably no further RHEL-8 TuneD release apart from the FDP, thus we cannot change the default directory in RHEL-8. I am also afraid it could break things.

For RHEL-9, I am also afraid it could break things for some people. 

I.e. I would change the profile location in upstream, i.e. in this PR and in further RHELs (e.g. RHEL-10). For older RHELs I would just backport the option from upstream, but wouldn't change the defaults. @jmencak if you want to change it in OCP you could reconfigure TuneD.

tuned.spec
```diff
@@ -329,9 +329,9 @@ if [ -r "%{_sysconfdir}/default/grub" ]; then
     %{_sysconfdir}/default/grub
 fi
 
-# move all user-defined profiles from /etc/tuned/ to /etc/tuned/profiles/
+# migrate all user-defined profiles from /etc/tuned/ to /etc/tuned/profiles/
 for f in %{_sysconfdir}/tuned/*; do
-  if [ -f "$f/tuned.conf" ] && [ ! -d "%{_sysconfdir}/tuned/profiles/`basename $f`" ]; then
+  if [ -e "$f/tuned.conf" ] && [ ! -e "%{_sysconfdir}/tuned/profiles/`basename $f`" ]; then
```
yarda, line 334: IMHO the current system profile directory is `/usr/lib/tuned/`, not `/usr/lib/tuned/profiles/`.

But I think you should migrate all profiles, because there could be custom profiles overrides in `/etc/tuned/` with the same name as profiles in `/var/lib/tuned/`. It's quite common to e.g. do:
```
# custom throughput-performance profile in /etc/tuned overriding setting of throughput-performance profile in /var/lib/tuned
[main]
# the stock one
include=throughput-performance

# customization
...
```

jmencak:
> There will be probably no further RHEL-8 TuneD release apart from the FDP, thus we cannot change the default directory in RHEL-8. I am also afraid it could break things.
> 
> For RHEL-9, I am also afraid it could break things for some people.
> 
> I.e. I would change the profile location in upstream, i.e. in this PR and in further RHELs (e.g. RHEL-10). For older RHELs I would just backport the option from upstream, but wouldn't change the defaults. @jmencak if you want to change it in OCP you could reconfigure TuneD.

Reconfiguring TuneD is a problem for very old but still supported OCP releases I'm afraid. As soon as you merge this and create and FDP out of it (as is), things will break in OCP.

yarda:
> > There will be probably no further RHEL-8 TuneD release apart from the FDP, thus we cannot change the default directory in RHEL-8. I am also afraid it could break things.
> > For RHEL-9, I am also afraid it could break things for some people.
> > I.e. I would change the profile location in upstream, i.e. in this PR and in further RHELs (e.g. RHEL-10). For older RHELs I would just backport the option from upstream, but wouldn't change the defaults. @jmencak if you want to change it in OCP you could reconfigure TuneD.
> 
> Reconfiguring TuneD is a problem for very old but still supported OCP releases I'm afraid. As soon as you merge this and create and FDP out of it (as is), things will break in OCP.

We will just add the option and don't change the defaults in older RHELs.  With the proposed:
```
profile_dirs = /usr/lib/tuned/profiles;/etc/tuned/profiles
```
You could reconfigure it yourself in OCP as you would like, e.g.:
```
profile_dirs = /usr/lib/tuned;/usr/lib/tuned/profiles;/etc/tuned;/etc/tuned/profiles
```
With such config all dirs will be supported and the upstream default dirs will have higher priority to override.

yarda:
Also the new upstream dirs will be seen only in OCP, not in standard RHEL-8/9, which will have the same config as it have now.

tuned.spec
zacikpa: The second check is actually only there to make sure that we don't overwrite existing directories in `/etc/tuned/profiles/` - though they have no reason to exist before the migration, and I could also do that more cleanly using `mv -n`.

zacikpa:
> @zacikpa I think the system profiles should be also moved to `/usr/lib/tuned/profiles`. Because, currently there can be the same problem with the conflicting profile names and special files/configuration files in this directory.

Ok, I will do the migration of both custom and system profiles in a separate PR. 

> Maybe global configuration option for it would be also good, but I think we don't need to auto-migrate them (the system profiles).
> 
> I am not sure whether two configuration options (custom/system profiles) have any benefit, maybe one configuration option for all? E.g.:
> 
> ```
> profile_dirs = /usr/lib/tuned/profiles;/etc/tuned/profiles
> ```
> 
> With the priority from left to right? The rightmost, the most prio. Opinions?

I think having it in a single option would be quite clean. Tuned-gui needs to explicitly distinguish the system profiles from the custom ones, however, so I'll have to think about how to resolve that.

zacikpa:
I'm also in favor of specifying the default values via `tuned.consts` and not directly via `tuned-main.conf` because the migration might then break TuneD for users that already use a modified version of `tuned-main.conf.`

yarda:
> I'm also in favor of specifying the default values via `tuned.consts` and not directly via `tuned-main.conf` because the migration might then break TuneD for users that already use a modified version of `tuned-main.conf.`

I think it may work as with other options. If unspecified or commented out then the default values from the consts should be used. For the auto migration you could check whether it's supported configuration and if not then do not auto-migrate. I think profiles under `/usr/lib/tuned` shouldn't be auto-migrated, just profiles in the '/etc/tuned' should be auto-migrated. I think profiles under `/usr/lib/tuned` should be handled by distro maintainer and if some users put their profiles there it's their responsibility to cope with it.

zacikpa:
The PR now contains only changes related to the new global configuration option, `profile_dirs`. Please re-review. I will open a new PR for the migration to new default profile locations.

tuned-main.conf
```diff
@@ -82,3 +82,8 @@ log_file_max_size = 1MB
 # - not_on_exit: rollbacks are always performed on a profile
 #   switch, but not on any kind of TuneD process exit
 # rollback = auto
+
+# Directories to search for profiles separated by , or ;
+# In case of conflicts in profile names, the later directory
+# takes priority
```
yarda, line 88: I would write "has priority" or "takes precedence", but I am not native speaker.

tuned/gtk/gui_profile_loader.py
```diff
@@ -60,7 +60,7 @@ def set_raw_profile(self, profile_name, config):
 
         profilePath = self._locate_profile_path(profile_name)
 
-        if profilePath == tuned.consts.LOAD_DIRECTORIES[1]:
+        if profilePath != tuned.consts.SYSTEM_PROFILE_DIR:
```
yarda, line 63: I think this could be configurable, maybe a list with system profile dirs? In tuned-gtk.conf? Or tuned-main.conf? It could be subset of profile_dirs that would be treated as as system profile dirs.

Or maybe a better way: Add tuned-gtk.conf and option like profile_save_dir which specifies where to save user profiles.

But this can be handled in later PRs, so LGTM for this PR.

yarda (approved):
Thanks, LGTM. Maybe improve the comment as mentioned earlier.

tuned-main.conf
zacikpa: Thanks, fixed. I mixed up the two.

tuned/gtk/gui_profile_loader.py
zacikpa: I agree. In fact, I have many more refactoring changes prepared for the GUI. To make this PR easily reviewable, I only included the necessary changes that keep the GUI working (more or less) as before.

yarda:
LGTM,thanks.

PR #611, opened 2024-03-11
uncore: Allow to configure frequency limits using percent

sgruszka:
Using direct frequency values in only useful in specialistic profiles
where we know target platform. Allow to configure uncore frequency as
percent of hardware maximum/minimum frequency. Such parameters can be
used in generic profiles like 'balanced' in portable way, for example:

[uncore]
max_freq_khz=90%

### DIFF
--- a/tuned/plugins/plugin_uncore.py
+++ b/tuned/plugins/plugin_uncore.py
@@ -19,17 +19,25 @@ class UncorePlugin(hotplug.Plugin):
 	`uncore`::
 
 	`max_freq_khz, min_freq_khz`:::
-	Limit the maximum and minumum uncore frequency.
+	Limit the maximum and minimum uncore frequency.
 
 	Those options are Intel specific and correspond directly to `sysfs` files
-	exposed by Intel uncore frequency driver.
+	exposed by Intel uncore frequency driver. Values can be specified as kHz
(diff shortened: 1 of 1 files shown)


yarda (changes requested):
Please add ASCII doc inline comments describing plugin parameters like other plugins have. Especially mention what happened in case of conflicting setting (i.e. khZ vs percentage).

Otherwise the code LGTM.

yarda:
What about using unit suffix to distinguish between percentage and kHz? Then we could avoid the possible conflict between kHz and percentage setting. E.g. we used something similar in `readahead` option of disk plugin, where the user can set the value in sectors or kiBs.

sgruszka:
> What about using unit suffix to distinguish between percentage and kHz? Then we could avoid the possible conflict between kHz and percentage setting. E.g. we used something similar in `readahead` option of disk plugin, where the user can set the value in sectors or kiBs.

Sorry  for delay, I was busy with other project. Yes, suffix makes perfect sense, will update.

sgruszka:
@yarda I made the update, please check.  Not sure why test are failing now, I think it's issue with CI.

yarda (approved):
LGTM, the CI failure is unrelated.

PR #612, opened 2024-03-11
plugin_net: do not read monitors if dynamic tuning is disabled

yarda:
Resolves: RHEL-28757


zacikpa (approved):
I  guess this is OK as a quick fix, but similar issues may come up in the disk plugin and the load plugin, which also use monitors.

yarda:
> I guess this is OK as a quick fix, but similar issues may come up in the disk plugin and the load plugin, which also use monitors.

Thanks, it's a good point, I will improve the PR.

yarda:
I added `_instance_init_dynamic` base plugin method and did all the dynamic tuning initialization there.

yarda:
Although the other plugins shouldn't pose problem, I think it's more clean now.

zacikpa (commented):
Looks good, thanks!

PR #615, opened 2024-03-27
Migrate profiles to /etc/tuned/profiles/ and /usr/lib/tuned/profiles/

zacikpa:
This PR changes the default locations of user and system profiles to `/etc/tuned/profiles/` and `/usr/lib/tuned/profiles/`, respectively. In Fedora and RHEL 10, existing user-defined profiles should also be automatically migrated.

Only the last commit of the PR is relevant for review, as it will be rebased onto #609.


zacikpa:
Rebased onto master, since #609 was merged.

PR #616, opened 2024-04-02
tuned-ppd: Detect battery change events

superm1:
power-profiles-daemon has the ability to detect battery change events using upower and to apply different tuned settings based upon whether plugged into power or not.

In PPD this is used specifically to set the energy performance preference differently in the 'balanced' profile, but there is no reason that this concept can't actually apply to all profiles.

Add support for detecting battery change events and apply a profile specified in `ppd.conf` for battery in the different PPD states.

CC @smallorange

### DIFF
--- a/tuned/ppd/config.py
+++ b/tuned/ppd/config.py
@@ -7,12 +7,18 @@ PPD_PERFORMANCE = "performance"
 
 MAIN_SECTION = "main"
 PROFILES_SECTION = "profiles"
+BATTERY_SECTION = "battery"
 DEFAULT_PROFILE_OPTION = "default"
+BATTERY_DETECTION_OPTION = "battery_detection"
 
 
 class PPDConfig:
     def __init__(self, config_file):
         self.load_from_file(config_file)
+
+    @property
+    def battery_detection(self):
+        return self._battery_detection
 
     @property
     def default_profile(self):
@@ -25,6 +31,10 @@ class PPDConfig:
     @property
     def tuned_to_ppd(self):
         return self._tuned_to_ppd
+
+    @property
+    def ppd_to_tuned_battery(self):
+        return self._ppd_to_tuned_battery
 
     def load_from_file(self, config_file):
         cfg = ConfigParser()
@@ -59,3 +69,17 @@ class PPDConfig:
         self._default_profile = cfg[MAIN_SECTION][DEFAULT_PROFILE_OPTION]
         if self._default_profile not in self._ppd_to_tuned:
             raise TunedException("Unknown default profile '%s'" % self._default_profile)
+
+        if BATTERY_DETECTION_OPTION not in cfg[MAIN_SECTION]:
+            raise TunedException("Missing battery detection option in the configuration file '%s'" % config_file)
+        self._ppd_to_tuned_battery = self._ppd_to_tuned
+        self._battery_detection = cfg.getboolean(MAIN_SECTION, BATTERY_DETECTION_OPTION)
+        if self._battery_detection:
+            if BATTERY_SECTION not in cfg:
+                raise TunedException("Missing battery section in the configuration file '%s'" % config_file)
+            for k, _v in dict(cfg[PROFILES_SECTION]).items():
+                if k in cfg[BATTERY_SECTION].keys():
+                    self._tuned_to_ppd = self._tuned_to_ppd | {cfg[BATTERY_SECTION][k]:k}
+            for k, v in dict(cfg[BATTERY_SECTION]).items():
+                if k in cfg[PROFILES_SECTION].keys():
+                    self._ppd_to_tuned_battery = self._ppd_to_tuned_battery | {k:v}
--- a/tuned/ppd/controller.py
+++ b/tuned/ppd/controller.py
@@ -13,6 +13,9 @@ DRIVER = "tuned"
 NO_TURBO_PATH = "/sys/devices/system/cpu/intel_pstate/no_turbo"
 LAP_MODE_PATH = "/sys/bus/platform/devices/thinkpad_acpi/dytc_lapmode"
 
+UPOWER_DBUS_NAME = "org.freedesktop.UPower"
+UPOWER_DBUS_PATH = "/org/freedesktop/UPower"
+UPOWER_DBUS_INTERFACE = "org.freedesktop.UPower"
 
 class PerformanceDegraded(StrEnum):
     NONE = ""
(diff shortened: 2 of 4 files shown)


zacikpa (commented):
Thanks for the PR, @superm1.

I'm a bit confused about the motivation behind this update. As I understand it, it's currently the users of ppd who choose to switch the ppd profile if the system changes from battery to AC or vice versa. For instance, gnome switches to the power-saving profile (using a hold) when below a certain battery level and the system is unplugged from AC. Can we be sure that this new extension won't cause any conflicts with the existing behavior of ppd/tuned-ppd users?

superm1:
Sure let me explain the background of it.  The PPD daemon does use upower to monitor for battery changes and some "PPD drivers" will react to this and make changes "within" their given profile.  GNOME and KDE aren't notified over D-Bus when this happens.  It's all internal to PPD.

The reason for bringing this functionality to tuned-ppd is so that when a distro changes to tuned and tuned-ppd the power consumption on battery doesn't go up.

superm1:
Sorry I realized I didn't answer your question. In terms of what GNOME does at low power (at least with PPD) it's tangential.

I think it's best to think of it as a "subprofile" for a given profile.

tuned/ppd/ppd.conf
```diff
@@ -1,9 +1,13 @@
...
+battery_detection=true
 
 [profiles]
 # PPD = TuneD
 power-saver=powersave
+power-saver-battery=powersave
 balanced=balanced
+balanced-battery=balanced-battery
 performance=throughput-performance
+performance-battery=throughput-performance
```
zacikpa, line 13: I'm not a fan of the "-battery" suffix syntax, as it introduces redundancy and requires many unappealing changes. As of now, it also breaks the `propertyChanged` signal, sending the profile name with the "battery" suffix included. Could we instead make it two sections?
```ini
[profiles]
power-saver=powersave
balanced=balanced
performance=throughput-performance

[battery]
balanced=balanced-battery
```
When the system is unplugged from AC, we could check whether the current profile is in the `battery` section and if yes, apply the TuneD profile from there.

tuned/ppd/controller.py
```diff
@@ -174,16 +200,22 @@ def set_active_profile(self, profile):
...
 
     @exports.property_getter("Profiles")
     def get_profiles(self):
+        profiles = []
+        for profile in ["power-saver", "balanced", "performance"]:
+            if profile in self._config.ppd_to_tuned:
+                profiles.append(profile)
         return dbus.Array(
-            [{"Profile": profile, "Driver": DRIVER} for profile in self._config.ppd_to_tuned.keys()],
+            [{"Profile": profile, "Driver": DRIVER} for profile in profiles],
```
zacikpa, line 218: This, for instance, wouldn't be necessary if using a separate config section for the "on-battery" profiles.

superm1:
> > Sorry I realized I didn't answer your question. In terms of what GNOME does at low power (at least with PPD) it's tangential.
> > I think it's best to think of it as a "subprofile" for a given profile.
> 
> Thanks for the explanation. I'm also wondering whether the dynamic part of #601 shouldn't also be moved here, as part of the "subprofiles".

Yup; that was already on my mind.  If this one gets accepted first I'll modify that one.  If that one gets accepted first I'll modify this one.

tuned/ppd/ppd.conf
superm1: Makes perfect sense.  I'll modify.

tuned/ppd/controller.py
```diff
@@ -182,8 +212,9 @@ def get_active_profile(self):
 
     @exports.property_getter("Profiles")
     def get_profiles(self):
+        profiles = self._config.ppd_to_tuned_battery if self._on_battery else self._config.ppd_to_tuned
         return dbus.Array(
-            [{"Profile": profile, "Driver": DRIVER} for profile in self._config.ppd_to_tuned.keys()],
+            [{"Profile": profile, "Driver": DRIVER} for profile in profiles.keys()],
```
zacikpa, line 217: Same as above.

zacikpa (changes requested):
Thanks for the update. Do I understand it correctly that power-profiles-daemon does not change its profile set when on battery? If so, I'd enforce that here as well, as suggested in one of the comments.

superm1:
> Thanks for the update. 

:+1: 

> Do I understand it correctly that power-profiles-daemon does not change its profile set when on battery? If so, I'd enforce that here as well, as suggested in one of the comments.

Sounds good.  I've made that change and your other suggested ones.

superm1:
> Yup; that was already on my mind. If this one gets accepted first I'll modify that one. If that one gets accepted first I'll modify this one.

I've made the changes to `plugin_video` on another branch that has both PRs merged to stop using upower and to instead have the separate profiles for balanced vs balanced-battery.  It works well.  Here is what that diff looks like: https://github.com/redhat-performance/tuned/commit/30dc23a74b3ccbacf76fe574487afb4d36e40058

tuned/ppd/controller.py
```diff
@@ -101,8 +104,27 @@ def __init__(self, bus, tuned_interface):
...
+    def setup_battery_signaling(self):
+        try:
+            bus = dbus.SystemBus()
+            self.proxy = bus.get_object(UPOWER_DBUS_NAME, UPOWER_DBUS_PATH)
+            self.proxy.connect_to_signal("PropertiesChanged", self.upower_changed)
+            self.upower_changed(None, None, None)
+        except dbus.exceptions.DBusException as error:
+            log.debug(error)
+            return False
+        return True
```
zacikpa, line 126:
```suggestion
```

tuned/ppd/config.py
```diff
@@ -59,3 +69,17 @@ def load_from_file(self, config_file):
         self._default_profile = cfg[MAIN_SECTION][DEFAULT_PROFILE_OPTION]
         if self._default_profile not in self._ppd_to_tuned:
             raise TunedException("Unknown default profile '%s'" % self._default_profile)
+
+        if BATTERY_DETECTION_OPTION not in cfg[MAIN_SECTION]:
+            raise TunedException("Missing battery detection option in the configuration file '%s'" % config_file)
+        self._ppd_to_tuned_battery = self._ppd_to_tuned
+        self._battery_detection = cfg[MAIN_SECTION][BATTERY_DETECTION_OPTION].lower() == "true"
```
zacikpa, line 76:
```suggestion
        self._battery_detection = cfg.getboolean(MAIN_SECTION, BATTERY_DETECTION_OPTION)
```

zacikpa (changes requested):
Thanks, looks good. Just two more nits.

zacikpa:
Perfect, LGTM.

I think the cleanest solution now would be to adjust #601 (keeping just the new video plugin option), and split this PR into two commits, one adding the battery functionality, the other updating `ppd.conf` with the new battery profiles.

Btw, is `panel_power_savings=0` the default value? If yes, I think we don't need to [add it to existing profiles](https://github.com/redhat-performance/tuned/commit/30dc23a74b3ccbacf76fe574487afb4d36e40058#diff-6f31c0584a055245ddf59e4dab78eb89159f9f2af2205ef6975d61508ff97189R25) at all.

superm1:
> Btw, is panel_power_savings=0 the default value? If yes, I think we don't need to [add it to existing profiles](https://github.com/redhat-performance/tuned/commit/30dc23a74b3ccbacf76fe574487afb4d36e40058#diff-6f31c0584a055245ddf59e4dab78eb89159f9f2af2205ef6975d61508ff97189R25) at all.

What does default mean?  I want to make sure that no matter what value it is set before starting tuned that it gets corrected to 0 if balanced.

zacikpa:
> > Btw, is panel_power_savings=0 the default value? If yes, I think we don't need to [add it to existing profiles](https://github.com/redhat-performance/tuned/commit/30dc23a74b3ccbacf76fe574487afb4d36e40058#diff-6f31c0584a055245ddf59e4dab78eb89159f9f2af2205ef6975d61508ff97189R25) at all.
> 
> What does default mean? I want to make sure that no matter what value it is set before starting tuned that it gets corrected to 0 if balanced.

Oh, right, that makes sense.

zacikpa (approved):
LGTM.

yarda (approved):
LGTM, thanks.

PR #617, opened 2024-04-02
Propagate unexpected exceptions in the wait exception handler

yarda:


zacikpa:
Huh, took me a while to figure out why there is the call with the extra `False` argument in the first place :) LGTM.

zacikpa (approved):

yarda:
Thanks for the review. When we drop python2 support, this hack can be also dropped.

PR #618, opened 2024-04-04
Update panel power savings for battery-balanced profile

superm1:
This is dependent upon these two merge requests:

https://github.com/redhat-performance/tuned/pull/616
https://github.com/redhat-performance/tuned/pull/601

### DIFF
--- a/profiles/balanced-battery/tuned.conf
+++ b/profiles/balanced-battery/tuned.conf
@@ -8,3 +8,6 @@ include=balanced
 
 [cpu]
 energy_performance_preference=balance_power
+
+[video]
+panel_power_savings=1


superm1:
#601 and #616 are merged, so I've rebased on upstream/master for this one.

yarda:
Thanks, LGTM.

septatrix:
Are there experimental numbers on how much this actually saves? Even just `panel_power_savings=1` leads to a very noticeable reduction in visual quality. Especially colors look washed out and contrast is reduced. It feels like reducing the brightness by 10-15% but also making colors more muted (on my Framework 13 AMD). The consequence was that whenever I was on battery power I simply increased the panel brightness which definitely seems counterproductive. (Nowadays I applied the hack from #699 at the cost of it also overwriting `panel_power_saving` for the `powersave` profile).

I have several colleagues/friends who have AMD Frameworks now. Some of them asked me if I ever also had problems with my display panel looking "bad" and when I showed them the workaorund (#699) they said it was a difference like night and day.

I am all in favor of saving power in places where it is often not noticeable (e.g. changing CPU governor) but `panel_power_savings` results in a very noticeable loss of fidelity. Most people (just like me) will probably just increase the brightness as a response which is surely counterproductive.

PR #619, opened 2024-04-04
daemon: buffer sighup signal

yarda:
Resolves: RHEL-31180

### DIFF
--- a/tuned/daemon/controller.py
+++ b/tuned/daemon/controller.py
@@ -71,10 +71,16 @@ class Controller(tuned.exports.interfaces.ExportableInterface):
 		self._terminate.set()
 
 	def sighup(self):
-		if not self._daemon._sighup_processing.is_set():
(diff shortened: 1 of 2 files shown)


yarda:
There is no need for atomic test & set or semaphores, because the signal handler is running in one thread thus the worst thing that could happen is that the SIGHUP on the edge of the time interval will be ignored. Nevertheless, the buffer should improve the responsiveness to the SIGHUP.

yarda:
It would be generally better and much cleaner to do the reloads in the main loop, but it would require bigger code changes.

jmencak:
Thank you for a quick fix, @yarda !
LGTM
I've also tested this and I haven't seen `TuneD` stuck after it was sent a SIGHUP after cca 800 iterations of our e2e test suite.

zacikpa (approved):
Thanks, looks good to me.

PR #621, opened 2024-04-09
plugin_cpu: decrease the severity of _has_pm_qos==False

adriaan42:
The feature depends on kernel configuration (CONFIG_CPU_IDLE). While this seems to be commonly enabled on x86, we've seen ARM systems where it's disabled, and in this case it's perfectly valid for /dev/cpu_dma_latency to be missing.
Since we cannot opt-out of the PM QoS feature in the cpu plugin, this case should not be an error.


yarda:
LGTM, thanks.

PR #622, opened 2024-04-10
Avoid unnecessary inter-processor interrupts by checking that writes are necessary

zacikpa:
This PR adds a new option `check_diff` to the function `commands.write_to_file`. When set to `True`, the function checks whether the current value is equal to the one being written - if it is, the write is skipped.

The new option is disabled by default, but I enable it for plugin options where the write may cause an inter-processor interrupt even when the value does not change.

### DIFF
--- a/tuned/plugins/plugin_cpu.py
+++ b/tuned/plugins/plugin_cpu.py
@@ -359,7 +359,7 @@ class CPULatencyPlugin(hotplug.Plugin):
 
 	def _set_intel_pstate_attr(self, attr, val):
 		if val is not None:
-			self._cmd.write_to_file("/sys/devices/system/cpu/intel_pstate/%s" % attr, val)
+			self._cmd.write_to_file("/sys/devices/system/cpu/intel_pstate/%s" % attr, val, ignore_same=True)
 
 	def _getset_intel_pstate_attr(self, attr, value):
 		if value is None:
@@ -524,7 +524,7 @@ class CPULatencyPlugin(hotplug.Plugin):
 					log.info("setting governor '%s' on cpu '%s'"
 							% (governor, device))
 					self._cmd.write_to_file("/sys/devices/system/cpu/%s/cpufreq/scaling_governor"
-							% device, governor, no_error = [errno.ENOENT] if remove else False)
+							% device, governor, no_error = [errno.ENOENT] if remove else False, ignore_same=True)
 				break
 			elif not sim:
 				log.debug("Ignoring governor '%s' on cpu '%s', it is not supported"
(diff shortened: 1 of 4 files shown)


jmencak (commented):
Thank you for the PR. I'm wondering why do we need a new option for this and why this is not a default? cc @MarSik

jmencak:
Thank you for the PR, @zacikpa . Testing this right now. So far, I've noticed the following issues.  When the same value is in the governor, we still write the same value.  The `check_diff` seems to be `False`.  Another issue I've noticed is that we'll need to change the TuneD configs.  There are some aliases like "performance" when written to `/sys/devices/system/cpu/cpuX/power/energy_perf_bias` will be translated to `0`.  So our configuration, such as 

```
energy_perf_bias=performance
```

will have to be changed to

```
energy_perf_bias=0
```
so that we take advantage of this new functionality.

Anyways, could you please look into the `scaling_governor` issue?  Thank you.

jmencak:
Also, running

```
x86_energy_perf_policy -r
```

causes quite a few IPIs.  This is pretty "expensive" if we're just checking that `x86_energy_perf_policy` is supported by `TuneD` and then we never use it anyway.  I wonder if we could just check for the presence of the file and whether it is executable?

jmencak:
I've finished testing the code.  In summary, I see these issues that I'd like to see fixed before merging:
* setting the same `scaling_governor` still triggers IPIs
* let's remove `x86_energy_perf_policy -r` check, which causes unnecessary IPIs

Issues that I do not expected fixed by this PR:
* `[rtentsk]` plug-in
* even `reading` EPP/EPB causes IPIs

zacikpa:
Hello @jmencak. I updated the PR so that:

- `check_diff` is `True` also for the `scaling_governor`,
- `x86_energy_perf_policy` is only checked for if we may actually need it, i.e., if we cannot set EPB via sysfs.  

Regarding the reads: if reading EPP is as expensive as writing, then I'd say it doesn't make sense to check the value first.

zacikpa:
@jmencak Regarding the aliases for EPB: can you please open a separate upstream or Jira issue for that?

jmencak:
Thank you for the changes, Pavol.  Great work!  The code now does what I expect it to do.  After testing, I can only see IPIs caused by the following options:

```
[rtentsk]
[cpu]
energy_perf_bias=<value>
energy_performance_preference=<value>
```

At this point this is expected and hard to avoid/work around.

The code changes look sane to me, but deferring to other reviewers for comment.

yarda (commented):
Thanks, the code LGTM. I have two minor comments that aren't blockers and could be handled later. I think this PR needs to be merged into FDP and we have still some time to fine tune it, so I will keep it open for now.

yarda (approved):

PR #624, opened 2024-04-11
functions: added 'log' which helps with debugging

yarda:
E.g.:
[variables]
dummy=${f:log:${f:exec:uname}}

Logs the output of the 'uname' command to the TuneD log.

Also added more verbose debug output which can help with functions debugging. Enable with 'tuned -D'.

### DIFF
--- a/tuned/profiles/functions/function_log.py
+++ b/tuned/profiles/functions/function_log.py
@@ -1,0 +1,19 @@
+import tuned.logs
+from . import base
+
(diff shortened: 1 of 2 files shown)


zacikpa (changes requested):
Thanks, this is quite handy.

Looks good to me, except for the unused `os` import.

yarda:
Thanks, it should be fixed now.

zacikpa (approved):

PR #625, opened 2024-04-14
dump: added new command

yarda:
The dump command dumps active profile. It's good for debugging, because the output is flattened, i.e. it shows the resulting profile after inheritance (profiles merging) is processed.

It keeps naming conventions - for tuned-adm it's named 'dump' action, for API it's named 'dump_profile' method.

The 'priority' value is not output, but output is ordered according to it.

The 'cpuinfo_regex' and 'uname_regex' are not output, but output is rendered according to it.

Variables are not expanded.


yarda:
It's work in progress PoC. I will probably change the API method to use structured output instead of the flat string, also documentation will be updated accordingly.

@jmencak please test and provide feedback.

yarda:
I will probably add `--hide-defaults` option for hiding default values that haven't been changed.

jmencak:
> The dump command dumps active profile. It's good for debugging, because the output is flattened, i.e. it shows the resulting profile after inheritance (profiles merging) is processed.
> 

I like the idea.

> Variables are not expanded.

Yeah, I've noticed. I wonder if support for this planned or at least variable definitions will be shown? They are not at the moment of testing.

Is there a Jira card for this so that I could read about the rationale apart from what I read above (useful for debugging)? Thank you!

yarda:
> > Variables are not expanded.
> 
> Yeah, I've noticed. I wonder if support for this planned or at least variable definitions will be shown? They are not at the moment of testing.
>
This is a good idea, I will add list of variables. It shouldn't be hard. For actual values, it would require extra storage (currently the values aren't stored anywhere and always re-calculated) and refactor of the current code. It's doable, but it depends whether such feature is needed.
 
> Is there a Jira card for this so that I could read about the rationale apart from what I read above (useful for debugging)? 
>
I think there is no Jira card at the moment, but we could create one. I was told by @MarSik few days ago that such feature is urgently needed.

jmencak:
> I think there is no Jira card at the moment, but we could create one. I was told by @MarSik few days ago that such feature is urgently needed.

Not sure what Martin had in mind, I haven't synced with him about this, but I like the idea of being able to inspect the current configuration.

yarda:
I will also add the `[main]` prefix to the output, because the output should be reusable as new TuneD profile.

PR #627, opened 2024-04-17
Re-check dbus when (re)starting TuneD from tuned-adm profile

### LINKED ISSUE #623: no error message when failing to apply profile while starting tuned
Hello,

When tuned fails to apply a profile while the service is started, we get a legitimate error message and appropriate return code (here I try to isolate 20 cpus but I only have 16):
```
root@w:~# systemctl stop tuned
root@w:~# systemctl start tuned
root@w:~# grep ^isol /etc/tuned/realtime-virtual-host-variables.conf
isolated_cores=1-20
isolate_managed_irq=Y
root@w:~# tuned-adm profile realtime-virtual-host
Unable to switch profile: Cannot load profile(s) 'realtime-virtual-host': Assertion 'isolated_cores contains online CPU(s)' failed.
```

But if the service is stopped for whatever reason (in my case, it's because it's stopped by default on freshly deployed servers, and enabled only on specific ones), then tuned starts the service, fails to apply the profile, but says otherwise:
```
root@w:~# systemctl start tuned
root@w:~# tuned-adm off
root@w:~# tuned-adm active
No current active profile.
root@w:~# systemctl stop tuned
root@w:~# grep ^isol /etc/tuned/realtime-virtual-host-variables.conf
isolated_cores=1-20
isolate_managed_irq=Y
root@w:~# tuned-adm profile realtime-virtual-host
Cannot talk to TuneD daemon via DBus. Is TuneD daemon running?
Trying to (re)start tuned...
TuneD (re)started, changes applied.
root@w:~# tuned-adm active
No current active profile.
root@w:~#
```

As you can see, there is no mention of tuned failing to apply the profile (although it did fail), we get a nice "changes applied" message with a returncode=0 so we could think that the profile was applied.

Thanks.

zacikpa:
When TuneD is off, and a profile is applied using `tuned-adm profile`, the command attempts to start it. Before this change, the profile would always be applied via the `active_profile` file.

When TuneD supports DBus, we can change the profile via a DBus call instead, and thus get information about the result. After the restart, I therefore added a new check which determines whether TuneD is on the dbus. If it's not, we fallback to the original profile activation.

The only disadvantage of this approach is that TuneD is restarted twice in the unlikely scenario that both `enable_dbus` and `daemon` are turned on, but TuneD does not become active on the DBus after a restart. This would, however, signalize some misconfiguration because if the TuneD service unit has `type=dbus` set, restarting the service hangs until it becomes active on the DBus. 

Resolves #623.

### DIFF
--- a/tuned-adm.py
+++ b/tuned-adm.py
@@ -125,7 +125,8 @@ if __name__ == "__main__":
 	result = False
 
 	profile_dirs = config.get_list(consts.CFG_PROFILE_DIRS, consts.CFG_DEF_PROFILE_DIRS)
-	dbus = config.get_bool(consts.CFG_DAEMON, consts.CFG_DEF_DAEMON)
+	daemon = config.get_bool(consts.CFG_DAEMON, consts.CFG_DEF_DAEMON)
+	dbus = daemon and config.get_bool(consts.CFG_ENABLE_DBUS, consts.CFG_DEF_ENABLE_DBUS)
 
 	try:
 		admin = tuned.admin.Admin(profile_dirs, dbus, debug, asynco, timeout, log_level)
--- a/tuned/admin/admin.py
+++ b/tuned/admin/admin.py
@@ -32,13 +32,16 @@ class Admin(object):
 		self._log_token = None
 		self._log_level = log_level
 		self._profile_recommender = ProfileRecommender()
-		if self._dbus:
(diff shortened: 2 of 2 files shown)


yarda:
Thanks, LGTM, but the changed output messages will require update of the Beaker tests.

PR #629, opened 2024-04-19
Bug fixes in tuned-ppd

zacikpa:
Having used tuned-ppd by default for a couple of months, I've discovered a couple of bugs that should be resolved by this PR.


yarda (approved):
Thanks, LGTM.

PR #630, opened 2024-04-25
cpu-partitioning-powersave no_turbo fix

christophefontaine:
Fix 'no_turbo' from bool (true) to int (1).
The current codes throws an error as it doesn't understand a string bool value, only an int.


christophefontaine:
Will rework the profile more globally, rather than just fixing that value.

PR #637, opened 2024-05-30
hotplug: wait for device initialization

yarda:
This is the first shot trying to resolve the potential problem with hotplug. It needs thorough testing. I am not sure whether it's a good thing to use it also for the remove events, maybe it would require refactor.

### DIFF
--- a/tuned/hardware/inventory.py
+++ b/tuned/hardware/inventory.py
@@ -54,6 +54,14 @@ class Inventory(object):
 		if not device.subsystem in self._subscriptions:
 			return
 
+		retry = consts.HOTPLUG_WAIT_FOR_DEV_INIT_RETRIES
+		while not device.is_initialized and retry > 0:
+			log.debug("Device '%s' is uninitialized, waiting '%.2f' seconds for its initialization." % (device, consts.HOTPLUG_WAIT_FOR_DEV_INIT_DELAY))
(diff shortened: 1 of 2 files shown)


tuned/hardware/inventory.py
```diff
@@ -54,6 +54,14 @@ def _handle_udev_event(self, event, device):
 		if not device.subsystem in self._subscriptions:
 			return
 
+		retry = consts.HOTPLUG_WAIT_FOR_DEV_INIT_RETRIES
+		while not device.is_initialized and retry > 0:
+			log.debug("Device '%s' is uninitialized, waiting '%d' seconds for its initialization." % (device, consts.HOTPLUG_WAIT_FOR_DEV_INIT_DELAY))
```
zacikpa, line 59: Maybe print this as float, the current log would just say 0 seconds in the default configuration.
```suggestion
			log.debug("Device '%s' is uninitialized, waiting '%.2f' seconds for its initialization." % (device, consts.HOTPLUG_WAIT_FOR_DEV_INIT_DELAY))
```

zacikpa (commented):
The changes look good to me, but I did no thorough testing. Just one nit:

tuned/hardware/inventory.py
yarda: Thanks, fixed.

zacikpa (approved):

yarda:
Centos 7 CI failure is caused by dropped EPEL-7, it's being worked on.

PR #638, opened 2024-06-10
CNF-13015: Add helper function to run `lscpu` and parse result using regex parameters

fontivan:
- This is similar to the cpuinfo function that reads from /proc/cpuinfo
- Unfortunately that function was insufficient for our needs as we need to identify both the Vendor and Architecture reliably

### DIFF
--- a/tuned/profiles/functions/function_lscpu_check.py
+++ b/tuned/profiles/functions/function_lscpu_check.py
@@ -21,7 +21,8 @@ class lscpu_check(base.Function):
 	def execute(self, args):
 		if not super(lscpu_check, self).execute(args):
 			return None
-		lscpu = self._cmd.execute("lscpu")
+		# Stdout is the 2nd result from the execute call
+		_, lscpu = self._cmd.execute("lscpu")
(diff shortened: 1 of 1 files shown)


tuned/profiles/functions/function_lscpu_check.py
```diff
@@ -0,0 +1,33 @@
...
+	"""
+	def __init__(self):
+		# unlimited number of arguments, min 2 arguments
+		super(lscpu_check, self).__init__("lscpu_check", 0, 2)
+
+	def execute(self, args):
+		if not super(lscpu_check, self).execute(args):
+			return None
+		# Stdout is the 2nd result from the execute call
+		_, lscpu = self._cmd.execute("lscpu")
```
fontivan, line 25: I could also check the return code here and potentially raise an exception though I'm not sure if its required

yarda:
Thanks, I think we can have this function. I am now @ DevConf, I will handle this during next week.

fontivan:
@yarda any update for this?

yarda (approved):

PR #640, opened 2024-06-15
chore: remove use of deprecated Logger.warn()

adriaan42:
Always use warning(), as warn() has been removed from Python with v3.13.


yarda:
Thanks for the fix.

PR #641, opened 2024-06-18
update RECOMMEND_DIRECTORIES const for profile migration

evgeni:
The installation was updated, but the consts were missing.

Fixes: 6d3aa949b3a7019bbb70f9bd3b34cb4a9381e8fb

### DIFF
--- a/tuned/consts.py
+++ b/tuned/consts.py
@@ -18,7 +18,7 @@ PERSISTENT_STORAGE_DIR = "/var/lib/tuned"
 PLUGIN_MAIN_UNIT_NAME = "main"
 # Magic section header because ConfigParser does not support "headerless" config
 MAGIC_HEADER_NAME = "this_is_some_magic_section_header_because_of_compatibility"
-RECOMMEND_DIRECTORIES = ["/usr/lib/tuned/recommend.d", "/etc/tuned/recommend.d"]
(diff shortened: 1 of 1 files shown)


tuned/consts.py
```diff
@@ -18,7 +18,7 @@
 PLUGIN_MAIN_UNIT_NAME = "main"
 # Magic section header because ConfigParser does not support "headerless" config
 MAGIC_HEADER_NAME = "this_is_some_magic_section_header_because_of_compatibility"
-RECOMMEND_DIRECTORIES = ["/usr/lib/tuned/recommend.d", "/etc/tuned/recommend.d"]
+RECOMMEND_DIRECTORIES = ["/usr/lib/tuned/profiles/recommend.d", "/etc/tuned/recommend.d"]
```
evgeni, line 21: I am not 100% sure this path is correct tho, as I kinda expected this to remain *outside* the `profiles/` folder as this is not a profile.

evgeni: related to https://github.com/redhat-performance/tuned/issues/642, I guess

evgeni:
Closing, as that was not supposed to be moved, according to https://github.com/redhat-performance/tuned/issues/642#issuecomment-2177861637

PR #646, opened 2024-06-21
Tuned ppd adjustments

zacikpa:
This PR introduces small changes in and related to `tuned-ppd`:
- The daemon now has the option to log output into a file (the default is `/var/log/tuned/tuned-ppd.log`).
- The PID option was removed from the systems service file: the daemon currently does not write its PID anywhere.
- The spec file is adjusted to include standard systems scriptlet macros according to [Fedora guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_systemd).
- When swapping `power-profiles-daemon` for `tuned-ppd`, scriptlets now check whether PPD is running and if yes, `tuned-ppd` is also started right away.

### DIFF
--- a/tuned.spec
+++ b/tuned.spec
@@ -338,6 +338,11 @@ for f in %{_sysconfdir}/tuned/*; do
 done
 %endif
 
+
+%post ppd
+%systemd_post tuned-ppd.service
+
+
 %preun
 %systemd_preun tuned.service
 if [ "$1" == 0 ]; then
@@ -346,6 +351,10 @@ if [ "$1" == 0 ]; then
 # clear temporal storage
   rm -f /run/tuned/*
 fi
+
+
+%preun ppd
+%systemd_preun tuned-ppd.service
 
 
 %postun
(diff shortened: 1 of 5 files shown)


tuned.spec
```diff
@@ -403,6 +424,15 @@ if [ -d %{_sysconfdir}/grub.d ]; then
 fi
 
 
+%posttrans ppd
+# if power-profiles-daemon was active before installing tuned-ppd,
+# start tuned-ppd right away
+if [ -f %{_localstatedir}/lib/rpm-state/tuned/ppd-active ]; then
+  systemctl start tuned-ppd
+  rm -rf %{_localstatedir}/lib/rpm-state/tuned
```
yarda, line 432: This will probably trigger rpmlint warning, thus it's better to do `rm -f` and then `rmdir`, but this is minor problem that can be ignored/waived or fixed later.

yarda (approved):
Thanks, it seems OK.

yarda:
Rawhide CI failure is unrelated.

PR #647, opened 2024-06-21
Add support for controlling amd-pstate core performance boost

superm1:
Core performance boost control is added to kernel 6.11. This sets up the policy to turn it off while in the power saver profile.

Link: https://lore.kernel.org/linux-pm/1a78eeaa-fadd-4734-aaeb-2fe11e96e198@amd.com/T/#m4a0c8917ea8fb033504055bd61512c80c85410c8


yarda:
Thanks, LGTM.

PR #650, opened 2024-06-24
Enable controller reset for audio by default

wangwillian0:
It fixes Nvidia suspension for laptops.

https://linrunner.de/tlp/settings/audio.html#sound-power-save-on-ac-bat


zacikpa (approved):
Looks good to me, thanks!

The Rawhide test failure in the CI is unrelated to the change (it seems to be related to the recent addition of TuneD to systemd presets).

yarda:
Thanks, LGTM.

PR #651, opened 2024-06-25
Beakerlib tests adjustments

zacikpa:
Some minor test adjustments:
- All other tests run `tunedDisableSystemdRateLimitingStart` without `rlRun`, so stick to it in `variables-support-in-profiles`, too.
- Add a missing `rlServiceRestore` to the `error-messages` test. This fixes [recent test failures](https://artifacts.dev.testing-farm.io/e67c22b9-4ea9-4be6-8cf1-ad060d839fd3/) in Rawhide, most likely triggered by the recent [addition of TuneD to systemd presets](https://src.fedoraproject.org/rpms/fedora-release/pull-request/335).

### DIFF
--- a/tests/beakerlib/variables-support-in-profiles/runtest.sh
+++ b/tests/beakerlib/variables-support-in-profiles/runtest.sh
@@ -24,7 +24,7 @@ rlJournalStart
     rlPhaseStartSetup
         rlAssertRpm $PACKAGE
         rlImport "tuned/basic"
-        rlRun "tunedDisableSystemdRateLimitingStart"
+        tunedDisableSystemdRateLimitingStart
         rlRun "TmpDir=\$(mktemp -d)" 0 "Creating tmp directory"
         rlRun "pushd $TmpDir"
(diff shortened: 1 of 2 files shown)


tests/beakerlib/variables-support-in-profiles/runtest.sh
```diff
@@ -24,7 +24,7 @@ rlJournalStart
     rlPhaseStartSetup
         rlAssertRpm $PACKAGE
         rlImport "tuned/basic"
-        rlRun "tunedDisableSystemdRateLimitingStart"
+        tunedDisableSystemdRateLimitingStart
```
yarda, line 27: I think all commands (including functions) should be started through `rlRun`, so it's easy to spot possible test problems in a consistent way. But I will let it on @RHEmployee to eventually make such changes. For now, I think it's good to be consistent, thus let's go with this change.

yarda (approved):
Thanks, LGTM.

PR #658, opened 2024-07-09
fix: expand variables in plugin_irq

adriaan42:
Forgot to expand variables in the new `irq` plugin.

Side question: is there a reason that variables in options are not automatically expanded, e.g. right after loading the profile? They are automatically expanded for `@command` invocations, but not for the "general" non-command tuning.

### DIFF
--- a/tuned/plugins/plugin_irq.py
+++ b/tuned/plugins/plugin_irq.py
@@ -94,7 +94,7 @@ class IrqPlugin(hotplug.Plugin):
 		instance._has_static_tuning = True
 		instance._has_dynamic_tuning = False
 
-		affinity = instance.options.get("affinity")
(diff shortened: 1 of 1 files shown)


yarda:
Thanks, LGTM.

> Side question: is there a reason that variables in options are not automatically expanded, e.g. right after loading the profile? They are automatically expanded for `@command` invocations, but not for the "general" non-command tuning.

I think it's because during the expansion various functions can be called which could have side effects. E.g. the `${f:exec:}` can alter the system state. Thus the variables are expanded and the functions executed only when needed and in the order the `@command`s are called.

But it could be improved.

yarda:
CentOS 7 CI failure is unrelated and being worked on by the TFT team.

PR #663, opened 2024-07-25
realtime: remove leftover use of tuna

### LINKED ISSUE #594: tuna?  You really want to verify tuna?
PopOS 22.04LTS. 
I edited /etc/tuned/realtime-virtual-host-variables.conf to isolate 4 cores like this:  
`isolated_cores=20-23`

After changing the active profile to realtime-virtual-host, I ran the verification:  
`sudo tuned-adm verify`  
which produced  

`Verification failed, current system settings differ from the preset profile.
    You can mostly fix this by restarting the Tuned daemon, e.g.:
    systemctl restart tuned`
 
which did not fix the error.  It persisted through the restart.  

cat /var/log/tuned/tuned.log:
```ERROR    tuned.plugins.plugin_script: script '/usr/lib/tuned/realtime/script.sh' error output: '/usr/lib/tuned/realtime/script.sh: line 14: tuna: command not found'```  

My first thought was, what?  tuna?  So I went and looked at the realtime/script.sh and sure enough, it says "verify() {tuna....".    

I didn't have anything to do with the "realtime" profile. I only tweaked the "realtime-virtual-host-variables.conf" file. 

For the record, I went back into the conf file and commented out my isolated_cores statement.  I then went back and made "balanced" the active profile (which was the default before I even thought about changing anything).  The restart and subsequent verify steps produced errors (on the purely default configuration).  A look at the logs indicated that tuned is trying to manipulate my grub configuration, but hello:  this machine doesn't use grub.  Grub is not even installed, but your stuff doesn't even check to see if Grub is the bootloader.  It just assumes...... It is also showing errors about various device hosts 'alpm' = 'None' expected 'medium_power'. There are a lot of those.  In all, there are a lot of errors in the default configuration, and my one little change resulted in this new tuna adventure. 

As it stands, the default tuned configuration never verifies as a success.  It fails directly from the default settings, and then gets really funky about tuna when I try to follow the example in the /etc/tuned/realtime-virtual-host-variables.conf to isolate a few cores for when I spin up any virtual machines.

adriaan42:
Before the `scheduler` plugin, `tuna` was used for CPU isolation. There is a leftover use in the verification of the `realtime` profile. Removing that, the dependency on the `tuna` package, and now-empty scripts from the realtime profiles.

closes #594


yarda:
Thanks, good catch.

PR #665, opened 2024-07-26
spec: used macro for profiles path and other fixes

yarda:
- improved conditionals for better readability
- fixed tuned-ppd not to run install twice
- used macro for profiles path

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,9 +47,11 @@ $(error Failed to determine python library directory)
 endif
 KERNELINSTALLHOOKDIR = /usr/lib/kernel/install.d
 TUNED_SYSTEM_DIR = /usr/lib/tuned
-TUNED_PROFILES_DIR = $(TUNED_SYSTEM_DIR)/profiles
+TUNED_SYSTEM_PROFILES_DIR = $(TUNED_SYSTEM_DIR)/profiles
+TUNED_CFG_DIR = $(SYSCONFDIR)/tuned
+TUNED_USER_PROFILES_DIR = $(TUNED_CFG_DIR)/profiles
 TUNED_RECOMMEND_DIR = $(TUNED_SYSTEM_DIR)/recommend.d
-TUNED_USER_RECOMMEND_DIR = $(SYSCONFDIR)/tuned/recommend.d
+TUNED_USER_RECOMMEND_DIR = $(TUNED_CFG_DIR)/recommend.d
 BASH_COMPLETIONS = $(DATADIR)/bash-completion/completions
 
 copy_executable = install -Dm 0755 $(1) $(2)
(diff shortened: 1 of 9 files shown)


zacikpa:
The changes look fine to me and the RPM build/installation works well. The failed CI tests do not seem related to the changes, though I'm not sure what's causing them. Approving.

zacikpa (approved):

yarda:
Rebased and hopefully fixed not to break the CI now :)

yarda:
It was a bit more tricky than I initially thought :). It seems I will also have to fix tests.

yarda:
It also requires: https://github.com/beakerlib/tuned/pull/10

yarda:
@zacikpa please re-review.

tests/beakerlib/Traceback-caused-by-scheduler-plugin-with/runtest.sh
```diff
@@ -29,33 +28,36 @@ rlJournalStart
...
+        PROFILE_DIR=$(tunedGetProfilesBaseDir)
+
+        rlRun "mkdir -p ${PROFILE_DIR}/test-profile"
+        rlServiceStop "tuned"
         sleep 1
-	rlFileBackup "/var/log/tuned/tuned.log"
-	rlRun "rm -rf /var/log/tuned/tuned.log"
+        rlFileBackup "/var/log/tuned/tuned.log"
+        rlRun "rm -rf /var/log/tuned/tuned.log"
+        rlServiceStart "tuned"
```
zacikpa, line 41: Why separate the service stop? I'd back up the logs first and then run rlServiceStart (which restarts TuneD if it's already running).

Makefile
```diff
@@ -161,26 +163,36 @@ install: install-dirs
...
+	# Update profiles location
+	sed '/^\s*#\?\s*profile_dirs\s*=\s*/ s|\(profile_dirs\s*=\s*\).*$$|\1$(TUNED_SYSTEM_PROFILES_DIR),$(TUNED_USER_PROFILES_DIR)|' \
+		tuned-main.conf > tuned-main.conf.tmp
+	touch -r tuned-main.conf tuned-main.conf.tmp
+	sed -i '/^\s*USER_PROFILES_DIR\s*=\s*/ s|\(USER_PROFILES_DIR\s*=\s*\).*$$|\1"$(TUNED_USER_PROFILES_DIR)"|' \
+		$(DESTDIR)$(PYTHON_SITELIB)/tuned/consts.py
+	sed -i '/^\s*SYSTEM_PROFILES_DIR\s*=\s*/ s|\(SYSTEM_PROFILES_DIR\s*=\s*\).*$$|\1"$(TUNED_SYSTEM_PROFILES_DIR)"|' \
+		$(DESTDIR)$(PYTHON_SITELIB)/tuned/consts.py
+	install -Dpm 0644 tuned-main.conf.tmp $(DESTDIR)$(TUNED_CFG_DIR)/tuned-main.conf
+	rm -f tuned-main.conf.tmp
```
zacikpa, line 175: Can you please add a comment here, explaining why it's necessary? Took me quite a while to figure out what's happening here.

I'm not a big fan of modifying the code like this, but I guess it's the best option at this point. Another way would be to add new options into `tuned-main.conf`, specifying the system/user profile directories. But since we already have the `profile_dirs` option there, it could get confusing...

zacikpa (commented):
Thanks, this looks good, I only have a couple of comments.

tests/beakerlib/Traceback-caused-by-scheduler-plugin-with/runtest.sh
yarda: It's taken from the RHEL dist-git upstream Beaker test. I wanted to sync to their version. IMHO it's trying to avoid possible race if the daemon is simultaneously logging while the file is removed/backed-up. Without it IMHO the worst case that could happen is that for multiline logs there would be some lines missing in the log (IMHO it uses line buffering). But the calls in the test could be reordered.

Makefile
yarda: Yep, probably better ways are usage of the m4 macro pre-processor or template file. It could also include generated file, but I didn't want to complicate things just for two variables.

yarda:
I improved the comment, but feel free to propose some more text.

Makefile
yarda: I improved the comment, but feel free to propose some more text.

tests/beakerlib/Traceback-caused-by-scheduler-plugin-with/runtest.sh
zacikpa: Ok, let's keep the tests in sync.

zacikpa (approved):

PR #666, opened 2024-07-27
Clear plugin repository when stopping tuning

zacikpa:
Not clearing it before a profile switch would
result in never deleting the original plugin
objects - those would accumulate in memory with
each profile switch.

TBH, I believe we could get rid of the plugin repository
class altogether, I'm not quite sure what's its responsibility
and why we need it when we have the plugin manager.


yarda:
Thanks, good catch. Please link the corresponding Jira ticket. I think it's worth to get it into the stable release.

yarda (approved):
LGTM, thanks.

PR #669, opened 2024-08-09
More tuned-ppd adjustments

zacikpa:
This PR brings another round of tuned-ppd adjustments and bug fixes. The individual commit messages explain why the changes are being made.

The most prominent change is dropping reverse profile mapping, i.e., from TuneD to PPD profiles. Instead, the daemon now keeps track of the currently active profile and the so-called base profile. This improves tuned-ppd in two ways:

1. Profile mappings now don't have to be injective.
2. When tuned-ppd is restarted, the initial profile is not the last active profile as inferred from the restored TuneD profile, but the last base profile, as saved into a file by tuned-ppd. The active and the base profile may differ due to profile holds being active before the restart. These should arguably be released when tuned-ppd is restarted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -178,6 +178,7 @@ install: install-dirs
 	echo -n > $(DESTDIR)$(TUNED_CFG_DIR)/active_profile
 	echo -n > $(DESTDIR)$(TUNED_CFG_DIR)/profile_mode
 	echo -n > $(DESTDIR)$(TUNED_CFG_DIR)/post_loaded_profile
+	echo -n > $(DESTDIR)$(TUNED_CFG_DIR)/ppd_base_profile
 	install -Dpm 0644 bootcmdline $(DESTDIR)$(TUNED_CFG_DIR)/bootcmdline
 	install -Dpm 0644 modules.conf $(DESTDIR)$(SYSCONFDIR)/modprobe.d/tuned.conf
 
--- a/tuned.spec
+++ b/tuned.spec
@@ -504,6 +504,7 @@ fi
 %config(noreplace) %verify(not size mtime md5) %{_sysconfdir}/tuned/active_profile
 %config(noreplace) %verify(not size mtime md5) %{_sysconfdir}/tuned/profile_mode
 %config(noreplace) %verify(not size mtime md5) %{_sysconfdir}/tuned/post_loaded_profile
+%config(noreplace) %verify(not size mtime md5) %{_sysconfdir}/tuned/ppd_base_profile
 %config(noreplace) %{_sysconfdir}/tuned/tuned-main.conf
 %config(noreplace) %verify(not size mtime md5) %{_sysconfdir}/tuned/bootcmdline
 %verify(not size mtime md5) %{_sysconfdir}/modprobe.d/tuned.conf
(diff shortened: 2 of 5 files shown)


zacikpa:
CI failures fixed.

zacikpa:
Rebased onto master because the issue resolved by the first commit was fixed by c082797fd03e89f9bc2319c578c05e324adebd7c.

zacikpa:
Added another change - a new `Version` DBus property, which signifies the level of compatibility with power-profiles-daemon (as suggested in https://github.com/redhat-performance/tuned/pull/684#issuecomment-2343883405).

zacikpa:
Moved the first commit into a separate PR: #697. That one should ideally land in Fedora 41 before the final freeze. The rest are more complex and will likely require some discussion.

velsinki:
Just chiming in here to say that technically this is api-incompatible still with PPD. The original PPD cannot return an `"unknown"` profile, whereas tuned-ppd can.

This can cause a crash in gnome-control-center: see https://bugzilla.redhat.com/show_bug.cgi?id=2324518.

Source: I'm the upstream (co)maintainer of gnome-control-center.

zacikpa:
Thanks, @velsinki. I'll adjust the code not to perform [this check](https://github.com/redhat-performance/tuned/pull/669/files#diff-79ef8f44396af6db79d6f8476e7a24d278b8e5f408f221538b88b36c3904afaeR331).

velsinki:
Thanks for the reply.

I'm not sure that's ideal either, as it would mean tuned-ppd is lying about which profile is active? We'll anyways fix the crash in gnome-control-center, not against keeping "unknown". I don't know what other consumers of PPD are doing though, GNOME Shell should have a similar issue.

velsinki:
> What's your proposed behavior of gnome-control-center if tuned-ppd returns "unknown"?

We currently show three radio buttons (performance, balanced, power-saver). If tuned-ppd would lie about a profile, we would start the UI with that radio button enabled, as we don't know anything else. That would mean that radio button can't actually be clicked anymore, as it's already active. If however we get "unknown", we can start the UI with no radio buttons selected instead.

Currently we don't handle "unknown", and crash, but that's not hard to support.

zacikpa:
> If however we get "unknown", we can start the UI with no radio buttons selected instead.

This sounds like very reasonable behavior.

> Currently we don't handle "unknown", and crash, but that's not hard to support.

In that case, I'd prefer to keep this as is and check in with the other consumers.

Sorry for the troubles and thanks for looking into this!

zacikpa:
Major update pushed now:
- After user feedback, I decided to stick with enforcing bijective PPD-to-TuneD profile mapping. We can thus translate TuneD profiles back to PPD profiles if the TuneD profile changes, e.g., via `tuned-adm`. This resolves #689. If users wish to simulate non-injective profile mapping, they'll have to create TuneD profile copies.
- Instead of actively polling for changes in the files related to performance degradation, the daemon now uses `inotify` to handle the changes.
- `inotify` is also used to watch the ACPI platform profile, which can optionally serve as an alternative way of switching PPD profiles on Thinkpad machines which support Fn+L/M/H function keys. This resolves #710.

yarda (approved):
Thanks, the code looks good.  I didn't run functionality tests, I don't own specific HW to test all the supported features.

PR #626, opened 2024-04-15
Refactor dynamic tuning and add per-plugin global configuration

zacikpa:

PR #678, opened 2024-08-21
disk plugin: ignore loop devices

panlinux:
It doesn't make sense to apply tuning parameters to loop devices, since they point at files that live in other block devices, also subject to tuning.

I checked the `device` structure looking for a telltale of a loop device. I thought that the `loop/backing_file` attribute would be a good choice, but I had some that had no such attribute. I could also inspect the name of the device, and assume "loop*" is a loop device. I ended up picking the major number, 7, and was a bit annoyed that there was no straight attribute with that number, so there is some ugly parsing of the `dev` attribute.

Let's use this PR to plant the idea of ignoring loop devices. If that sounds ok and I didn't miss a use case where we *want* loop devices, we can iterate over the best way to do it.

Thanks!


panlinux:
Thanks to Athos Ribeiro, here is a much simpler version that I force-pushed.

yarda:
Thanks. I think the change makes sense, but I am afraid it will break several Beaker functionality tests that use loop devices to safely test disk plugin features. That's why I would prefer it to be configurable, e.g. in `/etc/tuned/tuned-main.conf` and by default the loopdev tuning can be disabled. In such case if there is anybody who was using it they could easily re-enable it.

I think the loopdev can be detected if the device has the udev `disk/by-loop*` attribute(s), because IMHO if there is a loopdev there should be associated inode on the filesystem, so the `disk/by-loop-inode` should be there. But hardcoding the devid 7 is probably also OK.

yarda:
> Thanks. I think the change makes sense, but I am afraid it will break several Beaker functionality tests that use loop devices to safely test disk plugin features. That's why I would prefer it to be configurable, e.g. in `/etc/tuned/tuned-main.conf` and by default the loopdev tuning can be disabled. In such case if there is anybody who was using it they could easily re-enable it.
>
Or per plugin configuration when the #626 is merged.

PR #679, opened 2024-08-22
Parse no_turbo cpu plugin option using commands.get_bool

zacikpa:
The no_turbo knob only recognizes 0 or 1, so we have to convert the option value defined in the profile to an integer if necessary.

Related: #630
Resolves: RHEL-51760


yarda:
LGTM, thanks.

PR #680, opened 2024-08-23
drop support for cpuspeed

evgeni:
cpuspeed has not been in Fedora since 2011 [1] and last time in RHEL 6.

it should be fairly safe to drop support for it now

[1] https://bugzilla.redhat.com/show_bug.cgi?id=713572


yarda:
Thanks, it makes sense.

Issue #683, opened 2024-09-02
Advertise p-p-d compatibility D-Bus interface under UPower namespace


jpetso:
UPower [starting with release 0.20](https://gitlab.freedesktop.org/upower/power-profiles-daemon/-/releases/0.20) is advertising itself on D-Bus as 'org.freedesktop.UPower.PowerProfiles' following its move into the UPower umbrella. The previous 'net.hadess.PowerProfiles' interface is still kept for compatibility reasons, but might be removed in the future - e.g. [openSUSE tried to remove it immediately, but couldn't](https://bugzilla.suse.com/show_bug.cgi?id=1219956).

I'm just a bypasser looking out for compatibility concerns. If I'm reading the tuned-ppd source correctly, the compatibility interface is advertised under the old name only. If the new name isn't advertised, users of the interface will stick to the old one and there isn't any upgrade path, despite the upstream project endorsing the new interface within UPower.

So it would be nice if tuned-ppd also advertised both, or if in doubt, perhaps only the current one instead of the previous one.

Conan-Kudo:
This is actually going to need to be a requirement for tuned to replace ppd in Fedora 41...

yarda:
Good catch, @zacikpa could you handle it?

zacikpa:
Sure, will do.

iyanmv:
This will also affect users updating to KDE Plasma 6.3 (expected final release next February 11)

https://invent.kde.org/plasma/powerdevil/-/commit/223b927a8351f35d7c2c2a70ea6d78568051ba3c
https://bugs.kde.org/show_bug.cgi?id=498590

PR #684, opened 2024-09-03
tuned-ppd: Support the new UPower PPD namespace

zacikpa:
Since version 0.20, power-profiles-daemon exposes
its DBus interface under "org.freedesktop.UPower.PowerProfiles" alongside the previous name. To be compatible with PPD, this commit implements the same change into tuned-ppd.

Resolves: #683

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,7 @@ TUNED_USER_PROFILES_DIR = $(TUNED_CFG_DIR)/profiles
 TUNED_RECOMMEND_DIR = $(TUNED_SYSTEM_DIR)/recommend.d
 TUNED_USER_RECOMMEND_DIR = $(TUNED_CFG_DIR)/recommend.d
 BASH_COMPLETIONS = $(DATADIR)/bash-completion/completions
+PPD_BUS_NAMES = org.freedesktop.UPower.PowerProfiles net.hadess.PowerProfiles
 
 copy_executable = install -Dm 0755 $(1) $(2)
 rewrite_shebang = sed -i -r -e $(SHEBANG_REWRITE_REGEX) $(1)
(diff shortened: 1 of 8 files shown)


superm1:
Please be careful with this, the PPD namespace has other changes that are in flight.
See these:
*  https://gitlab.freedesktop.org/upower/power-profiles-daemon/-/merge_requests/210
* https://gitlab.freedesktop.org/upower/power-profiles-daemon/-/merge_requests/212

zacikpa:
Thanks for the heads up, @superm1. I just had a look at the proposed power-profiles-daemon changes, but I don't see anything that should be an immediate problem for us.

IIUC, you're adding some new functionality to the existing PPD namespace. As of now, tuned-ppd does not aim to cover the entire functionality - we're mostly interested in what's actively used by power management daemons such as powerdevil or gnome-control-center. The new functionality may or may not be added in the future.

zacikpa:
> Maybe it's best to advertise a compatibility level as well? I recall PPD has a version property. Maybe you can use "0.23" to indicate you match the API from 0.23 release?

Thanks, that's a nice suggestion, I added it into another open PR.

Conan-Kudo:
Can we please get this landed? Final Freeze is on Tuesday, and this needs to be in before then as things are listening on the UPower namespace exclusively already.

Conan-Kudo:
Thanks!

yarda:
https://bodhi.fedoraproject.org/updates/FEDORA-2024-ceee3c1021 it seems stuck in testing.

zacikpa:
Done.

Conan-Kudo:
What's holding this up to be merged? It's been shipped in Fedora for a while and it works fine.

Pointedstick:
It would be helpful to KDE to have this merged sooner rather than later (and also a new release with it made) because the upcoming Plasma 6.3 release needs this change to properly detect that `tuned-ppd` is installed.

The final release is on February 11th, so ideally it would be packaged and released in the various rolling release distros before then.

Thanks folks!

yarda (approved):

yarda:
The upstream release was delayed, @zacikpa will backport to Fedora.

yarda:
> It would be helpful to KDE to have this merged sooner rather than later (and also a new release with it made) because the upcoming Plasma 6.3 release needs this change to properly detect that `tuned-ppd` is installed.
> 
> The final release is on February 11th, so ideally it would be packaged and released in the various rolling release distros before then.
> 
> Thanks folks!

Sorry for the delay, we initially planned TuneD upstream release to December, but I didn't make it. Due to other deadlines the stable version has to be released in January, which means RC version should happen during next week. The 11th February should be fine and it should also give some time to distro packagers.

PR #685, opened 2024-09-04
refactor the code base to promote readability and maintainability in the future

iaacornus:
this is a **draft** proposal. I made this so I can know whether it would be
worth proceeding or not. Please kindly inform me if the maintainers would like
this work to proceed or not. Here are my proposed changes:

1. Comply to pep 0008 by:
     a. Convert all tabs into spaces
     b. sort imports
     c. improve the conditions in if statements
     d. remove like use of ""
     e. shorten functions
     f. remove unused constants
     g. simplify some expressions
     h. trim and shorten long lines/statements
     ...
2. Improve the logic of code and decrease its size

here are those only i can think at the moment. although it certainly extends deeper
with aims to increase the readability, promote maintainability and sustainability of
the code base in the future. since as of the moment, with all due respect, it is crappy
and shitty at best.

regards,
james aaron erang

### DIFF
--- a/tuned/admin/admin.py
+++ b/tuned/admin/admin.py
@@ -15,470 +15,470 @@ import threading
 import logging
 
 class Admin(object):
-	def __init__(self, profile_dirs, dbus = True, debug = False,
-			asynco = False, timeout = consts.ADMIN_TIMEOUT,
-			log_level = logging.ERROR):
-		self._dbus = dbus
-		self._debug = debug
-		self._async = asynco
-		self._timeout = timeout
-		self._cmd = commands(debug)
-		self._profiles_locator = profiles_locator(profile_dirs)
-		self._daemon_action_finished = threading.Event()
-		self._daemon_action_profile = ""
-		self._daemon_action_result = True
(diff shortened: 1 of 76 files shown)


iaacornus:
> Personally, I'm not against code reformatting, but if we do decide to go down that road, I'd be inclined towards using an automatic code formatter such as [black](https://github.com/psf/black) and [enforcing its rules in the CI](https://black.readthedocs.io/en/stable/integrations/github_actions.html).

oh yes, sure. would it be fine if i'll apply code formatting using some formatter out there? however, id also emphasize that there's more to its formatting. I've seen quite a lot to refactor, not meaning to offend here, which can somehow, although i doubt significantly, impact the code's performance. Well, its more of maintaining it easier, so it would not be hard to understand and read.

superm1:
I think that it would make sense to do it a few steps (multiple PRs); first do black on the codebase, enforce it from CI then do manual fixups that make sense layered on top of that.

iaacornus:
breaking this pull request into smaller one. closing it for now @zacikpa @superm1

PR #687, opened 2024-09-11
fix: expand variables in Plugin._verify_all_device_commands

adriaan42:
Need to expand variables before calling the `@command` function. This is actually done for `_execute_all[_non]_device_commands` and `_verify_all_non_device_commands`, but not in `_verify_all_device_commands`.
This adds the missing expansion.


zacikpa (approved):
Good catch! LGTM, thanks.

adriaan42:
@yarda ping

yarda:
LGTM, thanks.

PR #688, opened 2024-09-18
Explicitly support vm.dirty_(bytes|ratio) and vm.dirty_background_(bytes|ratio)

zacikpa:
The options interact within these two pairs: when `dirty_bytes` is set, `dirty_ratio` is automatically set to 0 and vice versa. Rollback thus does not work out of the box when setting these options via the sysctl plugin.

This PR adds explicit support for the options into the VM plugin: the new custom commands check whether the value set in the profile is permitted and the rollback is adjusted: the option that was not 0 before the profile switch is reverted.

@jmencak, feel free to test this.

### DIFF
--- a/tuned/plugins/plugin_vm.py
+++ b/tuned/plugins/plugin_vm.py
@@ -38,11 +38,26 @@ class VMPlugin(base.Plugin):
 			"transparent_hugepages" : None,
 			"transparent_hugepage" : None,
 			"transparent_hugepage.defrag" : None,
+			"dirty_bytes" : None,
+			"dirty_ratio" : None,
+			"dirty_background_bytes" : None,
+			"dirty_background_ratio" : None
 		}
+
+	@staticmethod
+	def _check_conflicting_dirty_options(instance, first, second):
+		if instance.options[first] is not None and instance.options[second] is not None:
+			log.error("Conflicting options '%s' and '%s', this may cause undefined behavior." % (first, second))
+
+	@staticmethod
+	def _proc_sys_vm_option_path(option):
+		return os.path.join("/proc/sys/vm", option)
 
 	def _instance_init(self, instance):
 		instance._has_static_tuning = True
 		instance._has_dynamic_tuning = False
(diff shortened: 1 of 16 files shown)


jmencak:
> @jmencak, feel free to test this.

Thank you! Planning to take a look today, tomorrow at the latest.

tuned/plugins/plugin_vm.py
```diff
@@ -119,3 +141,84 @@ def _get_transparent_hugepage_defrag(self):
 			return cmd.get_active_option(cmd.read_file(sys_file))
 		else:
 			return None
+
+	def _check_twice_pagesize(self, option, int_value):
+		min_bytes = 2 * int(cmd.getconf("PAGESIZE"))
+		if int_value >= min_bytes:
+			return True
+		log.error("The value of '%s' must be at least twice the page size (%s)." % (option, min_bytes))
```
jmencak, line 149: Nit: This is a matter of opinion/style, but I typically tend to do the "[Happy path](https://en.wikipedia.org/wiki/Happy_path)".  I.e. I deal with issues (like returning errors first) and then continue to solve the problem (here return True) in the end.  Similarly elsewhere in the code.  Not a deal breaker or a reason not to merge this though.

jmencak (commented):
Thank you for the PR, Pavol.  I've tested the code and it resolves https://issues.redhat.com/browse/RHEL-58820.  I've also checked the changes in TuneD profiles. They seem fine.  Also had a quick look at the code and apart from a minor nit with the style (which is a matter of opinion anyway) it looks fine to me.

tuned/plugins/plugin_vm.py
zacikpa: To be honest, that sounds like a nice practice... Updated.

jmencak:
Thank you for the changes.  I believe the whole PR is cleaner this way.  I've also retested the scenario in RHEL-58820.
LGTM

yarda (approved):
It looks good, thanks.

zacikpa:
Done.

PR #691, opened 2024-09-30
scheduler: Do not assume that perf events have type attribute

zacikpa:
The `perf.evlist.read_on_cpu` function may return instances of classes that do not have the `type` attribute, e.g. `perf.lost_event`. Make sure we do not read the `type` attribute if we are not sure that it exists.

### DIFF
--- a/tuned/plugins/plugin_scheduler.py
+++ b/tuned/plugins/plugin_scheduler.py
@@ -1098,10 +1098,13 @@ class SchedulerPlugin(base.Plugin):
 						event = instance._evlist.read_on_cpu(cpu)
 						if event:
 							read_events = True
-							if event.type == perf.RECORD_COMM or \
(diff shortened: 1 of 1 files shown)


tuned/plugins/plugin_scheduler.py
```diff
@@ -1098,10 +1098,13 @@ def _thread_code(self, instance):
 						event = instance._evlist.read_on_cpu(cpu)
 						if event:
 							read_events = True
-							if event.type == perf.RECORD_COMM or \
-								(self._perf_process_fork_value and event.type == perf.RECORD_FORK):
+							if isinstance(event, perf.comm_event) or (
```
yarda, line 1101: There could be explicit sanity check that `event.type == perf.RECORD_COMM`, but taking into account the perf interface is stable for years it's unlikely that some future perf update would break this "optimization".

yarda (approved):
LGTM, thanks.

PR #693, opened 2024-10-09
Automatic generation of plugin docs from their docstrings

zacikpa:
This PR follows #337, which introduced AsciiDoc docstrings for plugin classes. The docstrings were cleaned up and the relevant module of the manual is now auto-generated from them.


yarda (approved):
This is pretty cool, thanks.

PR #698, opened 2024-10-16
tuned-ppd: Unify polkit policy with power-profiles-daemon

zacikpa:
Rename HoldProfile/ReleaseProfile polkit actions to hold-profile/release-profile.

Add the switch-profile action and check against it when switching the profile using the ActiveProfile DBus property.

The changes require some refactoring of the DBusExporter class, mainly because we need to be able to supply a custom polkit action name instead of deriving it from the DBus method name.


yarda:
@zacikpa please resolve conflict

yarda:
Thanks, the CI failure seems unrelated.

PR #700, opened 2024-10-26
Do not initialize kvm low latency if kvm not present

lack:
On systems mising the kvm module, skip any attempts to initialize kvm.

Signed-off-by: Jim Ramsay <jramsay@redhat.com>


yarda:
Thanks, LGTM.

The `functions` are there mostly for the backward compatibility, we are trying to get rid of it and prefer the plugins written in python.

PR #703, opened 2024-11-03
powertop2tuned: use default user profile path from the consts

yarda:
It obeys defaults from the consts.

It could be also improved to obey TuneD main configuration, but it's not a problem at the moment, because the path is configurable in the powertop2tuned. Support of the TuneD main configuration would require changes in the main configuration file, because at the moment we do not explicitly mark there the user profile directory.


zacikpa (approved):
Works and LGTM.

PR #707, opened 2024-11-06
tuned-ppd: Disallow releasing profile holds of other processes

zacikpa:
Add a check which raises an error if the caller of ReleaseProfile
is not identical to the caller of HoldProfile.

Do not number cookies from 0, use pseudo-random integers.

Do not allow null cookies, which may cause issues in clients.
E.g., in gnome-settings-daemon, a null cookie means that it
currently has no active hold.


yarda:
Thanks LGTM, the CI failure seems unrelated.

PR #715, opened 2024-12-02
plugin_cpu: allow raw energy_performance_preference values

ruihe774:
Besides values specified in `energy_performance_available_preferences`, `energy_performance_preference` also accepts numeric values to allow finer controlling. This PR allows valid numeric values for `energy_performance_preference` in plugin `cpu`.

### DIFF
--- a/tuned/plugins/plugin_cpu.py
+++ b/tuned/plugins/plugin_cpu.py
@@ -798,7 +798,13 @@ class CPULatencyPlugin(hotplug.Plugin):
 			if not sim:
 				avail_vals = set(self._cmd.read_file(self._pstate_preference_path(cpu_id, True)).split())
 				for val in vals:
-					if val in avail_vals:
+					try:
+						val = int(val)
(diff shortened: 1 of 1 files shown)


yarda:
Thanks, LGTM.

superm1:
> Besides values specified in energy_performance_available_preferences, energy_performance_preference also accepts numeric values to allow finer controlling. This PR allows valid numeric values for energy_performance_preference in plugin cpu.

FWIW this is only true on intel-pstate at the moment.  But this field is actually used for any CPU that provides `energy_performance_preference`.  So you need to be careful if any profiles start using numbers instead of strings because they will only work on intel-pstate.

There is a discussion on the mailing list that may change this in the future to allow other drivers to take raw EPP values, or there may be another string for other drivers (TBD).

You can follow along here: https://lore.kernel.org/linux-pm/6909eef3-20aa-4341-9177-a42323a0d5c6@huawei.com/

But I just want to provide this warning now to avoid future bugs.

PR #716, opened 2024-12-02
plugin_scheduler: make perf support optional

yarda:
Requested in: https://src.fedoraproject.org/rpms/tuned/pull-request/8

### DIFF
--- a/tuned/plugins/plugin_scheduler.py
+++ b/tuned/plugins/plugin_scheduler.py
@@ -8,7 +8,12 @@ import tuned.logs
 import re
 from subprocess import *
 import threading
-import perf
+# perf is optional
+try:
+	import perf
+except ModuleNotFoundError:
+# if perf is unavailable, it will be disabled later
+	pass
 import select
 import tuned.consts as consts
 import procfs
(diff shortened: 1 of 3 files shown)


travier (approved):
Thanks, LGTM

tuned/plugins/plugin_scheduler.py
```diff
@@ -532,6 +545,7 @@ def _instance_init(self, instance):
 					instance._evlist.mmap(pages = perf_mmap_pages)
 			# no perf
 			except:
+				log.warn("python-perf unavailable, disabling perf support, you can try to (re)install python(3)-perf package")
```
zacikpa, line 548: I believe `logging.warn` is deprecated in favor of `logging.warning`. We recently moved to using the latter. Also, it would be nice if the warning said that we are disabling runtime tuning.

zacikpa (changes requested):
Works well, checked the functionality with `python3-perf` uninstalled.

Just one nit, see my comment.

tuned/plugins/plugin_scheduler.py
yarda: It seems I was used to it for very long :). Fixed and updated the string.

zacikpa (approved):
Looks good now, thanks!

PR #717, opened 2024-12-06
scheduler: Log process info when its affinity cannot be changed and re-check blacklisted cgroups

zacikpa:
Include the current affinity of the process, its cmdline, and its cgroups in the logs. This may help identify issues with tuning, e.g., in cases when the process disappears soon after affining it fails.

Resolves: RHEL-69933

Update: It may happen that a process gets moved into a blacklisted cgroup
after the initial filtering of processes according to the blacklist.
When setting its affinity fails, re-check whether that is the case,
and if yes, ignore the error.

Resolves: RHEL-72981

### DIFF
--- a/tuned/plugins/plugin_scheduler.py
+++ b/tuned/plugins/plugin_scheduler.py
@@ -644,13 +644,22 @@ class SchedulerPlugin(base.Plugin):
 	def _is_kthread(self, process):
 		return process["stat"]["flags"] & procfs.pidstat.PF_KTHREAD != 0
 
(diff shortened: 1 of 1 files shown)


zacikpa:
One more update per discussion with @MarSik: do the cgroup blacklist check as close to the affinity setting as possible to minimize the race window. I removed the initial blacklist filtering, but feel free to comment on that; we can keep it for efficiency.

yarda (approved):
LGTM.

PR #718, opened 2024-12-11
doc: enable documentation generation on RHEL with asciidoc

yarda:
With this patch the documentation will be also generated on RHEL with now obsoleted asciidoc tool. The styling is a bit lame, but it's usable.


zacikpa:
Cool! I think the asciidoc style is timeless :) 

What's worse is that it cannot handle syntax such as [option]\`skip_grub_config=VALUE\`. TBH, I'm not sure what's the purpose of those in the manual anyway. Is that used in Red Hat product docs?

yarda:
> Cool! I think the asciidoc style is timeless :)
> 
> What's worse is that it cannot handle syntax such as [option]`skip_grub_config=VALUE`. TBH, I'm not sure what's the purpose of those in the manual anyway. Is that used in Red Hat product docs?

I don't know :). In the worst case we could drop it. Is it causing any problem now?

zacikpa:
It displays literally when compiled by asciidoc:
image

zacikpa:
Agreed.

zacikpa (approved):

PR #725, opened 2025-01-12
scheduler: updated sched knobs for kernels 6.6+

yarda:
sched_wakeup_granularity_ns - removed
sched_min_granularity_ns    - now sched_base_slice_ns, same units and scale
sched_latency_ns            - removed

### DIFF
--- a/tuned/plugins/plugin_scheduler.py
+++ b/tuned/plugins/plugin_scheduler.py
@@ -401,7 +401,7 @@ class SchedulerPlugin(base.Plugin):
 	with hierarchy-ID 8 and controller-list blkio.
 	====
 
-	Recent kernels moved some `sched_` and `numa_balancing_` kernel run-time
+	Kernels 5.13 and newer moved some `sched_` and `numa_balancing_` kernel run-time
 	parameters from `/proc/sys/kernel`, managed by the `sysctl` utility, to
 	`debugfs`, typically mounted under `/sys/kernel/debug`.  TuneD provides an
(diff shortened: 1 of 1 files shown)


zacikpa:
The change looks functionally OK and works fine also on the newest kernels.

I wonder whether it's necessary to differentiate the case when a knob once existed, but does not exist now (reporting only a debug message). Wouldn't it be cleaner to unify the behavior to _always_ check for the existence of the knobs before accessing them and report a warning if they don't exist?

yarda:
@zacikpa thanks for the review.

> I wonder whether it's necessary to differentiate the case when a knob once existed, but does not exist now (reporting only a debug message). Wouldn't it be cleaner to unify the behavior to _always_ check for the existence of the knobs before accessing them and report a warning if they don't exist?

If I understand your comment correctly this would output warnings on 6.6 kernels, i.e. on RHELs newer than RHEL-9 until we drop the sched tunings from the stock profiles. And if we remove the sched tunings from the stock profiles than we will have regression on older kernels. I think the compatibility layer should work transparently. I.e. it should try the pre 5.13 knobs, then the 5.13 knobs and finally the 6.6 knobs. If the knob is known as removed without replacement then I think it's OK to do nothing.

I am open to discussion.

zacikpa (approved):

yarda:
I think it could happen only with custom configured/build kernels and chance of hitting such specific configuration just between this two kernel versions is probably low. Let's assume persons running such configurations know what they are doing :).

PR #727, opened 2025-01-14
tuned-ppd: Fix handling of profile change signals from TuneD

zacikpa:
Two adjustments needed:
- If multiple profile changes happen too quickly, we must handle the signal only from the last one - i.e., the one that matches the current TuneD profile.
- We must not clear profile holds with each signal from TuneD - the signals come even if the profile change was initiated by tuned-ppd itself.


yarda (approved):
LGTM.

PR #730, opened 2025-01-18
plugin_scheduler: fixed API

yarda:


yarda:
It happened because fix for the RHEL-60906 didn't conflict with previous fix and fix for RHEL-60906 wasn't rebased which broke API.

yarda:
Related to #728 and #725.

yarda:
And fixed docs generation on CentOS/RHEL-7.

yarda:
Also fixed optional perf handling on CentOS/RHEL-7.

PR #733, opened 2025-01-20
utils.commands: Fix CPU online detection when not present

marcan:
Not all platforms can offline CPUs. On those that can't, there is no "online" attribute. Default to online in that case.

### DIFF
--- a/tuned/utils/commands.py
+++ b/tuned/utils/commands.py
@@ -280,7 +280,7 @@ class commands:
 	def is_cpu_online(self, cpu):
 		scpu = str(cpu)
 		# CPU0 is always online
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

yarda:
Thanks, LGTM. Maybe it could log some info in such case, but IMHO in debug mode it will be visible in the logs even as it is.

I guess, it's also targeting 2.25.0 stable, right?

yarda (approved):

Conan-Kudo:
> I guess, it's also targeting 2.25.0 stable, right?

Yes. We'd like to see this shipped in Fedora 41+.

PR #737, opened 2025-01-23
Fix races occuring when two threads expand variables at the same time

zacikpa:
While fixing this, I got confused by the naming of the `PluginLoader` class. In the first commit, I propose to rename it to `ClassLoader`, and fix one of its methods so that it truly behaves like an abstract class.

For the race fix itself, see the second commit.

Resolves: RHEL-75773

### DIFF
--- a/compile_plugin_docs.py
+++ b/compile_plugin_docs.py
@@ -2,13 +2,13 @@
 import argparse
 import os
 import inspect
-from tuned.utils.plugin_loader import PluginLoader
+from tuned.utils.class_loader import ClassLoader
 from tuned.plugins.base import Plugin
 
 
-class DocLoader(PluginLoader):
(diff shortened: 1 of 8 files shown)


jmencak:
Thank you for the PR, Pavol, much appreciated!  The basic tests with NTO on OCP with this change pass.  I'm currently running a test that's trying to reproduce the original issue.  I had success reproducing the issue _with the old code_ without this fix.

So far, I wasn't able to reproduce the issue _with the new code._  I want to test this fix at least over night.  I'll report tomorrow morning how it went.

jmencak:
No race found with overnight testing. @yarda , it would be great to have this in FDP 25.A.

zacikpa:
Thanks for the thorough testing, @jmencak!

PR #738, opened 2025-01-27
bootloader: export Grub variables to make them available in submenus

ycongal-smile:
Without exporting the grub variables, they are only available in the main grub menu[0]. When descending in a submenu (like in default Debian for example), tuned_* variable are expanded to the empty string and, so, tuned kernel cmdline tuning is effectively disabled.

With the export, variables are available in submenus and tuned tuning is applied as expected.

[0]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1094201#10


yarda:
LGTM, thanks.

PR #729, opened 2025-01-16
makefile: added support for installation to custom $BINDIR/$SBINDIR

yarda:
This is required by Fedora 42 /usr/bin /usr/sbin merge.

PR #741, opened 2025-02-01
Makefile: Add support for installation to custom $LIBEXECDIR

SunRed:
Similar to #729 I would like to add this minor change to make packaging TuneD a little bit easier on distros that don't use `libexec`.


yarda:
Thanks, LGTM, the rawhide CI failure seems unrelated.

PR #745, opened 2025-02-04
profiles: use `med_power_with_dipm` for SATA ALPM

retrixe:
Using `min_power` can lead to data loss, and `med_power_with_dipm`
delivers similar power savings without the chance of data loss. [1]

`med_power_with_dipm` is the default setting[2] in upstream Linux for
SATA Link Power Management, and should provide significant power
savings on laptops using SATA drives over `medium_power`.

[1] https://hansdegoede.livejournal.com/18412.html
[2] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/drivers/ata/Kconfig?id=5433f0e7427ae4f5b128d89ec16ccaafc9fef5ee

**Note:** Testing on a SATA system would be appreciated to confirm this patch is setting `med_power_with_dipm` as expected, since unfortunately I'm not sure when I'll be able to get back to my desktop at home

### DIFF
--- a/doc/manual/modules/performance/con_inheritance-between-tuned-profiles.adoc
+++ b/doc/manual/modules/performance/con_inheritance-between-tuned-profiles.adoc
@@ -22,14 +22,14 @@ If the [replaceable]_parent_ profile is updated, such as after a *TuneD* upgrade
 
 .A power-saving profile based on balanced
 ====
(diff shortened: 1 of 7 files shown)


yarda:
Thanks LGTM. I tested it with the RHEL-8 kernel (probably the oldest supported configuration)  and it's supported there, so we don't need to add fallback mechanism (like e.g. `alpm = med_power_with_dipm | min_power`). If there is a need for it, support can be added later.

PR #747, opened 2025-02-11
network_latency: Set non-zero rcutree.nohz_full_patience_delay

zacikpa:
`rcutree.nohz_full_patience_delay` is a new kernel parameter added in 6.11. For more details, see https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=68d124b0999919015e6d23008eafea106ec6bb40.


yarda:
Thanks, LGTM. There is probably no need to drop it on older kernels, because unknown kernel argument should be ignored there, so we can probably use it as it is.

PR #753, opened 2025-02-26
tuned-ppd: Fix inotify watch for performance degradation

msizanoen1:
See individual commits for more details.

### DIFF
--- a/tuned/ppd/controller.py
+++ b/tuned/ppd/controller.py
@@ -293,11 +293,11 @@ class Controller(exports.interfaces.ExportableInterface):
         if self._no_turbo_supported:
             self._inotify_watches |= self._watch_manager.add_watch(path=os.path.dirname(NO_TURBO_PATH),
                                                                    mask=pyinotify.IN_MODIFY,
-                                                                   proc_fun=PerformanceDegradedEventHandler(NO_TURBO_PATH, self))
+                                                                   proc_fun=PerformanceDegradedEventHandler(self, NO_TURBO_PATH))
         if self._lap_mode_supported:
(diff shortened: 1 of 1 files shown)


zacikpa:
Oh, wow, good catch, thanks!

zacikpa (approved):

zacikpa:
@msizanoen1, I'm not able to reproduce the inode caching issue without the last commit applied. Could you please clarify how you're reproducing it and on what system? I tried writing "3" to `/proc/sys/vm/drop_caches`, but  inotify events were still delivered properly to tuned-ppd after that.

msizanoen1:
I was able to reproduce this on a ThinkPad X1 Carbon 9th gen running Fedora 41, against the `dytc_lapmode` trigger specifically. I just put it on my lap, wait until `dytc_lapmode` is 1, run `sysctl vm.drop_caches=3`, close the lid, move it to a stable surface, then reopen it. After reopening the lid `dytc_lapmode` should be `0` but GNOME Settings still reports that the lap sensor is triggered and `/var/log/tuned/tuned-ppd.log` doesn't report any changes.

The specific problematic kernel line of code is https://github.com/torvalds/linux/blob/76b6905c11fd3c6dc4562aefc3e8c4429fefae1e/fs/kernfs/file.c#L926 (should be `kernfs_get_inode(info->sb, kn)` instead) but it's arguable whether this is a kernel bug considering that using inotify on sysfs is not officially supported and changes to attributes should be listened to with `poll()` instead.

msizanoen1:
Please note that the `dytc_lapmode` file must *not* be open by any process on the system (check with `lsof` *before* running `sysctl vm.drop_caches=3`) for this to be reproducible.

Issue #710, opened 2024-11-19
tuned should track firmware initiated changes to the platform_profiles


ocrete:
On my ThinkPad X1 Carbon, when I press Fn-L/M/H, the firmware changes the profile profile, and this is reflected in /sys/firmware/acpi/platform_profile. The old power-profiles-daemon would monitor this file and update its state based on it. It seems this is missing from tuned.

ocrete:
As a user, I think that this would be the expected behaviour. Changing the profile through the keyboard shortcut should just change the whole tuned profile.

zacikpa:
The functionality is now implemented as part of #669. However, I'd consider it "experimental" at this point since we have to differentiate between when the file is changed by TuneD vs. when changed by the keypress (which is more difficult than expected). We'll likely make the functionality disabled by default, at least initially.

It will be possible to easily enable it by adding a single line (e.g., `thinkpad_function_keys=True`) to `/etc/tuned/ppd.conf`.

ocrete:
I though this didn't work.. But then I realized it wasn't enabled by default. I enabled it and now it works

PR #755, opened 2025-02-28
Fix thinkpad_function_keys feature

mrhpearson:
The thinkpad_function_keys feature is incorrectly named, so rename it to sysfs_acpi_monitor to be more correct.

It is also a useful feature - if profiles are updated on Fedora 41 outside of tuned, then they are not reflected in tuned or gnome-settings (or likely KDE, though I didn't test that). Enable the feature by default.

Tested on multiple devices. As a note - I expect this feature to be useful to any vendor supporting platform profiles, not just Thinkpads (though it does mean our function keys work correctly again which is nice)

Thanks
Mark

### DIFF
--- a/tuned/ppd/config.py
+++ b/tuned/ppd/config.py
@@ -11,7 +11,7 @@ PROFILES_SECTION = "profiles"
 BATTERY_SECTION = "battery"
 DEFAULT_PROFILE_OPTION = "default"
 BATTERY_DETECTION_OPTION = "battery_detection"
-THINKPAD_FUNCTION_KEYS_OPTION = "thinkpad_function_keys"
+SYSFS_ACPI_MONITOR_OPTION = "sysfs_acpi_monitor"
 
 
 class ProfileMap:
@@ -75,13 +75,13 @@ class PPDConfig:
         return self._tuned_to_ppd
 
     @property
-    def thinkpad_function_keys(self):
+    def sysfs_acpi_monitor(self):
         """
         Whether to react to changes of ACPI platform profile
         done via function keys (e.g., Fn-L) on newer Thinkpad
         machines. Experimental feature.
         """
-        return self._thinkpad_function_keys
+        return self._sysfs_acpi_monitor
 
     def load_from_file(self, config_file):
         """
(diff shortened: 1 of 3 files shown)


zacikpa:
Thanks, @mrhpearson.

While I agree that the original name was poorly chosen, some users are already using it (see #710), so we should at least make the old name an alias of the new one.

mrhpearson:
Hi @zacikpa 

Good point - though I'd argue that as the 2nd commit enables the feature by default the user won't see any impact will they? The feature will continue working for them correctly.
The issue would only be if anybody was disabling the feature in ppd.conf, but I doubt anybody is doing that.

Happy to change the commits if needed, just not sure if it's actually necessary. Let me know if I'm missing something.

Thanks for the review!
Mark

zacikpa:
The problem is that the configuration file is defined as "noreplace" in the spec file, and thus it won't be replaced if the user has already made changes to it. I believe it works similarly in other distros.

mrhpearson:
Right - is that a problem? You have a config file with an extra useless line in it is all.

I don't want to make more of this than is needed, so I'm happy to remove the naming change if that's preferred, and just enable the feature by default. Just figured it was a good opportunity to make it a bit more accurate.

I'll get that ready but just wanted to be sure it was really required.

tuned/ppd/config.py
```diff
@@ -141,4 +141,4 @@ def load_from_file(self, config_file):
         self._ppd_to_tuned = ProfileMap(profile_dict_ac, profile_dict_dc)
         self._tuned_to_ppd = ProfileMap({v: k for k, v in profile_dict_ac.items()}, {v: k for k, v in profile_dict_dc.items()})
 
-        self._thinkpad_function_keys = cfg.getboolean(MAIN_SECTION, THINKPAD_FUNCTION_KEYS_OPTION, fallback=False)
+        self._sysfs_acpi_monitor = cfg.getboolean(MAIN_SECTION, SYSFS_ACPI_MONITOR_OPTION, fallback=False)
```
zacikpa, line 144:
```suggestion
        self._sysfs_acpi_monitor = cfg.getboolean(MAIN_SECTION, SYSFS_ACPI_MONITOR_OPTION, fallback=True)
```

zacikpa: Please update to default to True even if the option is not explicitly defined.

zacikpa (changes requested):
Fair enough. Please update as requested and this should be good to go.

tuned/ppd/config.py
mrhpearson: Sorry - I might have got the work flow wrong....I'm used to doing kernel commits, I don't play in user space often.

Could I just have accepted the commit suggestion above without having to make the change in my tree? Hope the below works and I haven't messed things up.

zacikpa: Both approaches work fine :)

zacikpa:
Thanks! If you could just squash (fixup) the last change into the second commit, that would be perfect.

yarda:
LGTM, resolved conflict.

yarda:
CI failure seems unrelated.

PR #760, opened 2025-03-19
variables: Add an option to prepend child variables before parent ones

zacikpa:
When a child profile overwrites an option defined by its parent,
the option stays in its place in the option dictionary. Until now, this
prevented the child profile from using any _new_ variables in the option
definition, because those would get added to the end of the option
dictionary. As a workaround (see, e.g., redhat-performance#239), such variables would have
to be pre-defined (empty) in the parent profile. However, there is no
suitable workaround if we do not want to touch the parent profile.

This commit adds a new `prepend` option available in variable units.
With this option set to true, new variables are prepended before
parent variables instead of being added to the end. The location of
overwritten variables is not changed, i.e., they remain in their place.

As part of this change, we now allow multiple instances of variable
units. If multiple instances are defined, they must have unique names.
Instances that are not named "variables" must contain the option
`type=variables`. This is useful in scenarios when some variables
need to be prepended and some appended.

The following is a simple example of what this allows.

Parent profile:
```
[variables]
var_x="defined by parent"
var_y="also defined by parent"
```
Child profile:
```
[main]
include=parent

[variables_prepend]
type=variables
prepend=True
var_z="defined by child"
var_x=${var_z}

[variables_append]
type=variables
var_y_copy=${var_y}
```
When the child profile is loaded, all four variables will be defined with the values:
```
var_x="defined by child"
var_y="also defined by parent"
var_z="defined by child"
var_y_copy="also defined by parent"
```
Resolves redhat-performance#751.

### DIFF
--- a/tuned/consts.py
+++ b/tuned/consts.py
@@ -21,6 +21,7 @@ USER_PROFILES_DIR = "/etc/tuned/profiles"
 SYSTEM_PROFILES_DIR = "/usr/lib/tuned/profiles"
 PERSISTENT_STORAGE_DIR = "/var/lib/tuned"
 PLUGIN_MAIN_UNIT_NAME = "main"
+PLUGIN_VARIABLES_UNIT_NAME = "variables"
 # Magic section header because ConfigParser does not support "headerless" config
 MAGIC_HEADER_NAME = "this_is_some_magic_section_header_because_of_compatibility"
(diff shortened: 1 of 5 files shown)


yarda:
LGTM, thanks, CI failure is unrelated.

PR #761, opened 2025-03-26
perf: dropped from irqbalance and improved detection in scheduler

yarda:
Now the irqbalance plugin will work even if perf is unavailable. Also improved perf detection in the scheduler plugin. If the perf.cpu_map() fails for some reason, perf is disabled and the perf library will not segfault.

### DIFF
--- a/tuned/plugins/plugin_irqbalance.py
+++ b/tuned/plugins/plugin_irqbalance.py
@@ -3,7 +3,6 @@ from .decorators import command_custom
 from tuned import consts
 import tuned.logs
 import errno
-import perf
 import re
 
 log = tuned.logs.get()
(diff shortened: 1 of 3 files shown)


zacikpa:
LGTM, tested that no traceback is shown when `python3-perf` is not installed.

zacikpa (approved):

jmencak:
> LGTM, tested that no traceback is shown when `python3-perf` is not installed.

Yes, also tested, no traceback.

jmencak:
LGTM, with this fix, NTO's CI test that was failing passes now.

yarda:
CI failure is unrelated.

PR #763, opened 2025-04-02
hotplug: fixed device remove race condition

yarda:
During profile loading if the device is assigned to the instance but it isn't processed yet by the main thread and the remove event fires, TuneD may try to remove the device from the free list, because it wrongly assumes that the device is unassigned. With this workaround it shouldn't traceback in such cases, but there may be still some errors when it tries to tune the removed device in the main thread. We currently cannot do much about it.

The add/rename udev events should be mostly safe, because there are two threads - the main thread and the udev callback thread. The udev callback thread is processing udev events sequentially one by one and it shouldn't break the main thread data when processing the add/rename events. It would be better to have one worker thread for it, but it would require quite a big redesign.

Also the cleanup phase was improved, because now it firstly unsubscribes the udev events and then it continues with the cleanup which should minimise problems of udev events firing during the profile unload.

Resolves: RHEL-83827


yarda:
Te be nearly 100% safe it would also require https://github.com/redhat-performance/tuned/pull/762#issuecomment-2760838228 for NTO.

zacikpa (approved):
LGTM, thanks.

yarda:
Fedora rawhide CI error seems unrelated.

PR #764, opened 2025-04-07
Docs and Web: Add to the master branch and build with each push

zacikpa:
In this PR, I propose to move the webpage files from the `gh-pages`  to the master branch. The webpage (preview at https://zacikpa.github.io/tuned/) now contains a link to the documentation (preview at https://zacikpa.github.io/tuned/docs/manual.html), which would be automatically deployed with each push to master.

The documentation of plug-ins and built-in functions is built from code docstrings.

The PR contains some additional refactoring of TuneD built-in functions, see the individual commits.


zacikpa:
@yarda, technically, the static webpage files could be kept in the `gh-pages` branch, but the workflow itself should be in master. Let me know what you think is appropriate. Personally, I would prefer to have everything in master and get rid of the large files (i.e., fonts).

zacikpa:
I did some minor changes to the docs themselves, e.g., updates from `yum` to `dnf`. I also noticed that [downstream RHEL docs](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/10/html/monitoring_and_managing_system_status_and_performance/optimizing-system-performance-with-tuned) diverged considerably from these - I don't think it's a blocker for merging this, but we should likely discuss the future of these upstream docs. I'd be personally inclined towards making them distinct from the RHEL docs and generating as much doc content as possible from the source code.

yarda:
> I did some minor changes to the docs themselves, e.g., updates from `yum` to `dnf`. I also noticed that [downstream RHEL docs](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/10/html/monitoring_and_managing_system_status_and_performance/optimizing-system-performance-with-tuned) diverged considerably from these - I don't think it's a blocker for merging this, but we should likely discuss the future of these upstream docs. I'd be personally inclined towards making them distinct from the RHEL docs and generating as much doc content as possible from the source code.

In the past the RHEL docs were generated from the TuneD upstream, but currently it probably isn't the case. Also the content for RHEL seems a bit outdated and it would require some work on the doc team side, so I am OK with the current PR. In the future, if there is no more collaboration with the doc team, we could clean it up and move to the generated content.

PR #766, opened 2025-04-15
beakerlib: Use ncat explicitly instead of nc

zacikpa:
This PR should fix recent test failures in Fedora Rawhide (for some reason, Rawhide has netcat installed by default).


RHEmployee:
Seems good.
Hope it will not break dependencies :).
But it should not.

yarda:
LGTM, thanks.

PR #769, opened 2025-05-02
scsi: Do not set ALPM on external SATA ports

zacikpa:
See the commit message for details and reasoning.

An alternative solution could be to ignore the external ports altogether during the initialization of the plugin. Since ALPM is currently the only option of the plugin, it would have the same effect; but if we wanted to add more tuning options in the future, the current solution is probably better. Let me know what you think.

Resolves: RHEL-79913


yarda:
LGTM, thanks, CI failure is unrelated.

PR #771, opened 2025-05-06
tuned-main.conf: added startup_udev_settle_wait option

yarda:
Timeout in seconds to maximally wait until udev settles. If set to value bigger than 0, it will either wait until udev settles or the timeout occurs. Disable with 0.

Resolves: RHEL-88238

### DIFF
--- a/tuned-main.conf
+++ b/tuned-main.conf
@@ -77,7 +77,7 @@ log_file_max_size = 1MB
 # connections_backlog = 1024
 
 # TuneD daemon rollback strategy. Supported values: auto|not_on_exit
-# - auto: rollbacks are always performed on a profile switch or 
+# - auto: rollbacks are always performed on a profile switch or
 #   graceful TuneD process exit
 # - not_on_exit: rollbacks are always performed on a profile
 #   switch, but not on any kind of TuneD process exit
@@ -87,3 +87,13 @@ log_file_max_size = 1MB
 # In case of conflicts in profile names, the later directory
 # takes precedence
 # profile_dirs = /usr/lib/tuned/profiles,/etc/tuned/profiles
+
+# Timeout in seconds to wait until udev settles. Value 0 disables the wait.
+# With non-zero wait TuneD startup is delayed until there are no pending
+# udev events or the specified timeout occurs.
+# The delay can help in situations where there are a lot of udev events
+# during TuneD startup and TuneD is too slow to handle them.
+# In environments where systemd is utilised for starting TuneD
+# it is better to keep this feature disabled and rely on systemd
+# functionality (systemd-udev-settle).
+startup_udev_settle_wait = 0
--- a/tuned/consts.py
+++ b/tuned/consts.py
@@ -146,6 +146,7 @@ CFG_UNIX_SOCKET_CONNECTIONS_BACKLOG = "connections_backlog"
 CFG_CPU_EPP_FLAG = "hwp_epp"
 CFG_ROLLBACK = "rollback"
 CFG_PROFILE_DIRS = "profile_dirs"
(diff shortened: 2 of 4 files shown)


yarda:
@jmencak on FDP I would go with 10 s, but probably not more than 30 s.

yarda:
Tests are failing probably due to recent @RHEmployee changes to the beakerlib TuneD lib related to profile directory detection. @RHEmployee could you fix it?

RHEmployee:
> Tests are failing probably due to recent @RHEmployee changes to the beakerlib TuneD lib related to profile directory detection. @RHEmployee could you fix it?

Hi. I reverted changes in beakerlib.
It has been plain sync with downstream library version.
However, `tunedGetSystemProfilesBaseDir` seems to be missing in upstream and some tests using that function, which is now removed from downstream.

yarda:
> > Tests are failing probably due to recent @RHEmployee changes to the beakerlib TuneD lib related to profile directory detection. @RHEmployee could you fix it?
> 
> Hi. I reverted changes in beakerlib. It has been plain sync with downstream library version. However, `tunedGetSystemProfilesBaseDir` seems to be missing in upstream and some tests using that function, which is now removed from downstream.

OK, please PR update.

yarda:
CI test rawhide:  /tests/beakerlib/Expose-TuneD-API-to-the-Unix-Domain-Socket took timeout
Maybe the test or the socket interface implementation will require some changes for rawhide (this needs to be checked), but it doesn't seem to be related to this PR.

tuned-main.conf
```diff
@@ -87,3 +87,13 @@ log_file_max_size = 1MB
...
+
+# Timeout in seconds to maximally wait until udev settles, 0 to disable
+# and don't wait. It delays TuneD startup until there are no pending udev
+# events or the specified timeout occurs.
+# It can help in situations where there are a lot of udev events
+# during TuneD startup and TuneD is too slow to handle them.
+# It's usually used in setups where systemd isn't utilised for starting
+# TuneD (e.g. in OpenShift), otherwise it's better to rely on systemd
+# and disable this feature.
+startup_udev_settle_wait = 0
```
jmencak, line 99: My recommendation would be to rephrase the comment to something like:

```
# Timeout in seconds to wait until udev settles. 0 disables the wait.
# With non-zero wait TuneD startup is delayed until there are no pending
# udev events or the specified timeout occurs.
# The delay can help in situations where there are a lot of udev events
# during TuneD startup and TuneD is too slow to handle them.
# In environments where systemd is utilised for starting TuneD
# it is better to keep this feature disabled and rely on systemd
# functionality (systemd-udev-settle).
```

jmencak:
> @jmencak on FDP I would go with 10 s, but probably not more than 30 s.

I agree this shouldn't be more than 30s.  Personally, I'd start with 20s for FDP though.

jmencak (commented):
Thank you for the PR!

I have a minor recommendation to document this feature.

I've tested this PR and it worked fine.

tuned-main.conf
yarda: Updated. I used "Value 0" instead of "0".

I also explicitly removed the monitor object not to cause any resource leaks.

yarda:
Alternative way is to directly check the udev queue `/run/udev/queue`  like the `udevadm settle` does, it would require less resources, but it would exploit udev/systemd internals. So I am preferring the used solution that utilizes pyudev/udev public API.

zacikpa (approved):
Code looks good. Tested only on a system with low number of devices - works fine. Thanks!

yarda:
Thanks for the review.

PR #772, opened 2025-05-07
tests/beakerlib: remove old API call.

RHEmployee:
Library API has been changed downstream and we need to reflect those changes in upstream.

### DIFF
--- a/tests/beakerlib/error-messages/runtest.sh
+++ b/tests/beakerlib/error-messages/runtest.sh
@@ -24,16 +24,18 @@ rlJournalStart
         rlAssertRpm $PACKAGE
         rlImport "tuned/basic"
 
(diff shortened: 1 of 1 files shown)


yarda:
@RHEmployee tests are still failing, it seems the profile path isn't expanded:
```
:: [ 10:32:31 ] :: [  BEGIN   ] :: Running 'cat /etc/tuned//balanced/tuned.conf | grep alpm='
cat: /etc/tuned//balanced/tuned.conf: No such file or directory
:: [ 10:32:31 ] :: [   FAIL   ] :: Command 'cat /etc/tuned//balanced/tuned.conf | grep alpm=' (Expected 0, got 1)
```

yarda:
The rawhide test failure is probably unrelated.

PR #775, opened 2025-05-16
CI: build and test for c10s

zacikpa:
BTW, does it still make sense to build and test the newest upstream for CentOS 7?


yarda:
> BTW, does it still make sense to build and test the newest upstream for CentOS 7?

IMHO I still saw some support tickets, but it's unlikely we would rebase there (:crossed_fingers:  :smile:), so I am OK to drop it.

yarda:
LGTM, thanks, c10s coverage was needed.

yarda:
CI failure is unrelated.

PR #778, opened 2025-05-20
udev-settle: obey udev buffer size and handle possible tracebacks

yarda:
Resolves: RHEL-92637

### DIFF
--- a/tuned/daemon/controller.py
+++ b/tuned/daemon/controller.py
@@ -62,15 +62,25 @@ class Controller(tuned.exports.interfaces.ExportableInterface):
 		if wait_settle > 0:
 			log.info("waiting for udev to settle")
 			monitor = pyudev.Monitor.from_netlink(pyudev.Context())
+			udev_buffer_size = self._global_config.get_size("udev_buffer_size", consts.CFG_DEF_UDEV_BUFFER_SIZE)
+			try:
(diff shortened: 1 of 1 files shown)


yarda:
Rawhide failure is unrelated.

yarda:
It now uses default TuneD 1MB udev events buffer instead of the 8 kB default udev buffer. The buffer size is configurable in the `tuned-main.conf`. This should prevent filling of the buffer even in case of a storm of udev events. It also contains workaround for the pyudev bug:
https://github.com/pyudev/pyudev/issues/194
Even if the buffer is filled now, the udev queue is no more stuck and legible error is logged.

yarda:
Rebased.

zacikpa (approved):
Right, so whenever the polling fails due to the udev socket filling up, we essentially abort the wait and behave as if it timeouted. That makes sense given that there is nothing else to do in such cases.

yarda:
> Right, so whenever the polling fails due to the udev socket filling up, we essentially abort the wait and behave as if it timeouted. That makes sense given that there is nothing else to do in such cases.

I could drop the context (because without the pyudev bug fix it cannot be reused and it isn't receiving any more events) and restart the wait, but I wouldn't complicate it.

zacikpa:
Yes, let's keep it like this, there is no guarantee that a restart would not lead to another exception.

PR #781, opened 2025-05-29
Dockerfile: New to build on a bootc image

cgwalters:
This is a pretty straightforward wiring up of the `make rpm` flow and then installing it in a bootc container image.

Intended as a starting point both for local development but also wiring up further CI.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+# Avoid rebuilding when this changes
+Dockerfile
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,0 +1,22 @@
+name: Docker Build
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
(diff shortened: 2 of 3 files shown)


yarda:
Thanks.

What about switching to the `ubuntu-latest` from the `ubuntu-24.04`? IMHO this should also allow dropping the podman debian-testing hack:

- ubuntu-latest (currently 25.04): podman-5.4.1+ds1-1
- ubuntu-24.04: podman-4.9.3+ds1-1ubuntu0.1
- debian-testing: podman_5.4.2+ds1-2+b1

yarda:
IMHO the `ubuntu-latest` would also save us from ubuntu version bumps in the workflow.

cgwalters:
> ubuntu-latest (currently 25.04)

No, see https://github.com/actions/runner-images/commit/976232da217887825e7541c2635bf39cbbf22654

yarda:
OK, merging as it is, we can switch to the `ubuntu-latest` later.

PR #782, opened 2025-06-02
Priority inheritance fix

yarda:


zacikpa:
I'm unsure about the behavior w.r.t. the systemd `system.conf` files in different locations.

If I understand the [manpages](https://www.man7.org/linux/man-pages/man5/systemd-system.conf.5.html) correctly, only the first found `system.conf` is loaded. Thus, if there already is some configuration in `/usr/lib/systemd/system.conf` and you create `/etc/systemd/system.conf` with just the new TuneD options, then upon systemd reload only the TuneD options will be loaded.

Secondly, I don't understand why we would want to abort the tuning when neither of  `/usr/lib/systemd/system.conf` and `/etc/systemd/system.conf` exist, but proceed with the tuning when only `/usr/lib/systemd/system.conf` exists.

yarda:
@zacikpa could you please review it?

yarda:
CI failure seems unrelated.

zacikpa (approved):
LGTM, thanks.

PR #783, opened 2025-06-03
Add option to disable gui in tuned

sourabhtk37:
Allow users to control whether they want to install tuned gui or not. 

Having gui in tuned is not beneficial in cloud environment and we can reduce the footprint by removing the files which are not needed. This ensure various Linux distributions to control whether a gui is needed or not.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,8 @@ NAME = tuned
 BUILD = release
 # which config to use in mock-build target
 MOCK_CONFIG = rhel-7-x86_64
+# Allow users to build tuned without gui
+INSTALL_GUI ?= 1
 # scratch-build for triggering Jenkins
 SCRATCH_BUILD_TARGET = rhel-7.5-candidate
 VERSION = $(shell awk '/^Version:/ {print $$2}' tuned.spec)
@@ -74,8 +76,10 @@ release-cp: release-dir
 	cp -a tuned.py tuned.spec tuned.service tuned.tmpfiles Makefile tuned-adm.py \
 		tuned-adm.bash dbus.conf recommend.conf tuned-main.conf 00_tuned \
 		92-tuned.install bootcmdline modules.conf com.redhat.tuned.policy \
-		tuned-gui.py tuned-gui.glade tuned-ppd.py \
-		tuned-gui.desktop functions compile_plugin_docs.py $(VERSIONED_NAME)
+		tuned-ppd.py functions compile_plugin_docs.py $(VERSIONED_NAME)
+	ifeq ($(INSTALL_GUI),1)
+		cp -a tuned-gui.py tuned-gui.glade tuned-gui.desktop $(VERSIONED_NAME)
+	endif
 	cp -a doc experiments libexec man profiles systemtap tuned contrib icons \
 		tests $(VERSIONED_NAME)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -74,8 +76,10 @@ release-cp: release-dir
 	cp -a tuned.py tuned.spec tuned.service tuned.tmpfiles Makefile tuned-adm.py \
 		tuned-adm.bash dbus.conf recommend.conf tuned-main.conf 00_tuned \
 		92-tuned.install bootcmdline modules.conf com.redhat.tuned.policy \
-		tuned-gui.py tuned-gui.glade tuned-ppd.py \
-		tuned-gui.desktop functions compile_plugin_docs.py $(VERSIONED_NAME)
+		tuned-ppd.py functions compile_plugin_docs.py $(VERSIONED_NAME)
+	ifeq ($(INSTALL_GUI),1)
+		cp -a tuned-gui.py tuned-gui.glade tuned-gui.desktop $(VERSIONED_NAME)
+	endif
```
ferdnyc, line 82: Oh, this is gonna break hard. `release-cp` is the target for creating the release tarball, it should NOT be conditional on `INSTALL_GUI` or anything else.

tuned.spec
```diff
@@ -299,7 +302,7 @@ to TuneD from power-profiles-daemon (PPD).
 make html %{make_python_arg}
 
 %install
-make install DESTDIR="%{buildroot}" BINDIR="%{_bindir}" SBINDIR="%{_sbindir}" \
+make install INSTALL_GUI="%{?gui}" DESTDIR="%{buildroot}" BINDIR="%{_bindir}" SBINDIR="%{_sbindir}" \
```
ferdnyc, line 305: You should use the `%(with_foo}` inline macro, with `%bcond_{with,without} foo` and `%{with foo}` / `%{without foo}`.
```suggestion
make install %{!?with_gui:INSTALL_GUI=0} DESTDIR="%{buildroot}" BINDIR="%{_bindir}" SBINDIR="%{_sbindir}" \
```

ferdnyc: Might even be better to have a `DISABLE_GUI` option instead, no default value, and `ifndef` that in the Makefile. `ifeq` is a pain when testing optional configs.

ferdnyc:
The other thing is, the GUI portion of tuned is _insignificant_. It has no meaningful "footprint". The main `tuned` package on Fedora is 1678860 bytes (1.68 MB) installed. `tuned-gtk` is 166601 (167 KB), or under 10% of that size. The primary reason the GTK portion is split out in packaging is because it creates a user-facing application, which distros prefer to avoid unless explicitly requested.

yarda:
I agree:
- the footprint of the GUI is insignificant, only cca. 163 kB
- even with the patch and GUI disabled it still installs `%{python_sitelib}/tuned/gtk/` content
- you can usually drop the GUI and the deps by simply omitting the resulting subpackage (when used with some distro)

Nevertheless, I am not against support in the `Makefile` (although I don't see much benefit), but I would prefer using different targets over the variables, i.e. something like the following (which IMHO is more make-like):
```
install-core: ....
install-profiles: ...
install-gui:...
install: install-core install-profiles install-gui
```

But I think in the future we will need to switch to `toml/setuptools` which is more pythonic and can help to spread the project and we could also get rid of some nasty hacks we currently have in the `Makefile`.

sourabhtk37:
Thank you for the comments. This PR can be closed.

PR #791, opened 2025-07-05
tuned-ppd: Fix degraded performance monitoring

nmraz:
I recently noticed that tuned-ppd never reported `PerformanceDegraded` on my Thinkpad, even when `/sys/devices/platform/thinkpad_acpi/dytc_lapmode` was set.

Turns out the parameters passed to the `PerformanceDegradedEventHandler` were reversed; switching them makes everything work flawlessly on my machine.


nmraz:
Turns out this was already fixed by #753 and I completely missed it :)

PR #792, opened 2025-07-08
vm: Deprecate dirty_ratio in favour of dirty_bytes with %

zacikpa:
Same for dirty_background_ratio and dirty_background_bytes.

Using both the ratio and the bytes is not compatible with the current profile inheritance implementation, because it is not possible for dirty_bytes in a child profile to override dirty_ratio in its parent profile.

Resolves: RHEL-101578


yarda (approved):
The changes make sense, thanks.

PR #805, opened 2025-10-08
cpu-partitioning: autodetect dracut hook directory

yarda:
Wrong hook directory can make the machine unbootable: https://issues.redhat.com/browse/RHEL-119889


profiles/cpu-partitioning/tuned.conf
```diff
@@ -23,6 +23,7 @@ tmpdir=${f:strip:${f:exec:mktemp:-d}}
 isolated_cores_expanded=${f:cpulist_unpack:${isolated_cores}}
 isolated_cpumask=${f:cpulist2hex:${isolated_cores_expanded}}
 not_isolated_cores_expanded=${f:cpulist_invert:${isolated_cores_expanded}}
+not_isolated_cores=${f:cpulist_pack:${not_isolated_cores_expanded}}
```
jmencak: Do we want to add a note this is because of https://issues.redhat.com/browse/RHEL-97580 and that it should be dropped afterwards?  Also, should this be in a separate PR as this has nothing to do with the dracut functionality?

PR #807, opened 2025-10-22
Adjust console logging format when running under systemd

rvl:
Most systemd services write their logs to the systemd journal by default. I think tuned should do likewise, because users would expect this, and it's useful to be able to read tuned logs together with other system logs.

So this PR changes the systemd services to remove the `-l` (log to `/var/log/tuned/tuned.log`) option.

It also adds a systemd-specific log formatter, which is only enabled when running in a systemd service and logging to console.

I chose not to use the `python-systemd` module because the extra dependency can be avoided with very little code.

### DIFF
--- a/tuned.service
+++ b/tuned.service
@@ -9,7 +9,7 @@ Documentation=man:tuned(8) man:tuned.conf(5) man:tuned-adm(8)
 Type=dbus
 PIDFile=/run/tuned/tuned.pid
 BusName=com.redhat.tuned
-ExecStart=/usr/sbin/tuned -l -P
+ExecStart=/usr/sbin/tuned -P
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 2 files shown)


tuned/logs.py
```diff
@@ -116,13 +122,25 @@ def remove_all_handlers(self):
...
+		return os.environ.get("JOURNAL_STREAM", "") != ""
+
 	@classmethod
 	def _setup_console_handler(cls):
 		if cls._console_handler is not None:
 			return
 
 		cls._console_handler = logging.StreamHandler()
-		cls._console_handler.setFormatter(cls._formatter)
+		if cls._detect_systemd_journal_stream():
```
yarda: Shouldn't it also use the "legacy" formatter if the "-l" option is specified? E.g. for the backward compatibility.

PR #811, opened 2025-11-13
profiles: Set boost=1 in *-performance profiles

nmraz:
Commit 068928e ("Add support for controlling amd-pstate core performance boost") started setting the newly-introduced `boost` parameter in the balanced and powersave profiles, but forgot to do the same in the performance profiles. The net result was that transitioning directly from the powersave profile to a performance profile would leave boosting disabled.

Fix this by setting boost=1 explictly in all performance-oriented profiles (i.e., those with a min_perf_pct of 100).


yarda:
Thanks, LGTM.

PR #824, opened 2026-02-05
openshift: include network-throughput profile

yarda:
Update network-throughput profile and drop now duplicate tuning from the openshift profile. Follows up the #809 and cleans the things a bit.

### DIFF
--- a/profiles/openshift/tuned.conf
+++ b/profiles/openshift/tuned.conf
@@ -4,7 +4,7 @@
 
 [main]
 summary=Optimize systems running OpenShift (parent profile)
-include=${f:virt_check:virtual-guest:throughput-performance}
+include=${f:virt_check:virtual-guest:network-throughput}
 
 [selinux]
 avc_cache_threshold=8192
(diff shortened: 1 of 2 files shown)


gheffern:
I like moving these into network-throughput.

I think the addition of net.ipv4.tcp_notsent_lowat=131072 mitigates the biggest issues of increased TCP buffer sizes, those being buffer bloat latency and increased memory usage.

Because of that I almost wonder if it makes sense to reverse the relationship between throughput-performance and network-throughput.  So that throughput-performance also includes network-throughput enhancements.  Modern workloads tend to have a lot of network activity.

I think that would then mean these profiles would get the TCP throughput changes as well:
virtual-guest
virtual-host
accelerator-performance
oracle
mssql 
sap-hana

I think its also worth updating the following comment:

\# Increase kernel buffer size maximums.  Currently this seems only necessary at 40Gb speeds.

To say something like:
\# Increase kernel buffer size maximums to cover a wider range of bandwidth delay products on varying networks

As moderate bandwidth, higher latency links can benefit from these changes as well

profiles/openshift/tuned.conf
```diff
@@ -17,10 +17,6 @@ kernel.pid_max=>4194304
 fs.aio-max-nr=>1048576
 net.netfilter.nf_conntrack_max=1048576
 net.ipv4.conf.all.arp_announce=2
-net.ipv4.tcp_notsent_lowat=131072
-net.ipv4.tcp_slow_start_after_idle=0
```
Fothsid, line 21: These two options are removed completely from the profile, without being moved to network-throughput. Is this correct/intentional?

Fothsid: GitHub decided to show the lines that I didn't select. I meant `net.ipv4.tcp_rmem` and `net.ipv4.tcp_wmem` options.

Fothsid: I see now that those options were already present in network-throughput, so they were not part of the diff. Ignore this comment.

Fothsid (approved):
LGTM.

PR #827, opened 2026-02-27
bootloader: Add bootc loader-entries set-options-for-source support

jmarrero:
On bootc systems with transient /etc, the state file
/etc/tuned/bootcmdline is lost after each reboot. This causes
TuneD to lose track of which kernel arguments it previously
set, leading to kargs stacking up on every reboot cycle.

bootc loader-entries set-options-for-source solves this by
recording kargs ownership directly in the BLS config on /boot,
which persists across reboots regardless of /etc transience.
With --source tuned, bootc automatically removes all previous
kargs from the "tuned" source and replaces them with the new
set, eliminating the need for TuneD to maintain its own state
for diffing.

This commit adds bootc source tracking to the bootloader plugin:

1. Detect set-options-for-source availability by checking
   "bootc loader-entries set-options-for-source --help" exit
   code during plugin init.

2. When available, use "bootc loader-entries
   set-options-for-source --source tuned --options ..." for
   applying kargs and the same command without --options for
   clearing all TuneD-owned kargs on profile removal.

3. The bootc path is a top-level dispatch branch, checked
   before rpm-ostree. Fallback chain: bootc -> rpm-ostree
   legacy --delete/--append -> GRUB2. All existing code paths
   are unchanged.

4. Continue writing to /etc/tuned/bootcmdline as best-effort
   for diagnostics and backward compatibility, but do not
   depend on it for correctness when bootc source tracking
   is available.

5. Ensure the bootloader plugin is always loaded on bootc
   systems, even for profiles without a [bootloader] section.
   Without this, switching from a profile with kargs (e.g.
   network-latency) to one without (e.g. throughput-performance)
   after a reboot would leave stale kargs behind, because the
   bootloader plugin would never be instantiated and its
   unapply/clear logic would never run. The daemon now injects
   a synthetic bootloader unit when bootc source tracking is
   available, and the plugin declares the empty kargs state to
   bootc so stale kargs from a previous profile are cleared.

6. Add 20 unit tests covering: bootc detection (3), apply (3),
   removal (2), init flag derivation (2), top-level dispatch (5),
   stale kargs clearing (2), and daemon bootloader unit
   injection (3).

Requires: bootc with loader-entries set-options-for-source
          (bootc PR #2114) and ostree >= 2026.1
See: https://github.com/bootc-dev/bootc/issues/899
See: https://github.com/bootc-dev/bootc/pull/2114

Assisted-by: OpenCode (Claude Opus 4.6)
Signed-off-by: Joseph Marrero Corchado <jmarrero@redhat.com>

## Summary by CodeRabbit

* **New Features**
  * Bootloader plugin adds support for tracking and managing tuned kernel command-line options on systems using bootc.
  * Daemon now ensures a synthetic bootloader unit is present at startup when needed.
  * Systems using bootc skip initrd-overlay handling to avoid incompatible overlays.

* **Tests**
  * Added comprehensive unit tests covering bootc integration, fallback behavior, and bootloader unit injection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/redhat-performance/tuned/pull/827)

### DIFF
--- a/tests/unit/plugins/test_bootloader.py
+++ b/tests/unit/plugins/test_bootloader.py
@@ -1,0 +1,390 @@
+import unittest
+
+try:
+	from unittest.mock import Mock
+except ImportError:
+	from mock import Mock
+
+import tuned.consts as consts
+
+
+class BootloaderPluginBootcTestCase(unittest.TestCase):
+	"""Tests for the bootc loader-entries set-options-for-source integration
+	in the bootloader plugin.
+
+	These tests exercise the bootc detection, source-based kargs apply,
+	source-based kargs removal, and the dispatch logic that falls back to
+	the rpm-ostree or GRUB2 code paths when bootc is not available.
+
+	The BootloaderPlugin constructor requires /etc/grub.d/00_tuned to exist,
+	so we test the methods directly on a minimal mock rather than
+	instantiating the full plugin.
+	"""
+
+	def _make_plugin_mock(self, bootc_has_source=True):
+		"""Create a mock that has the real methods under test patched onto it.
+
+		This avoids instantiating BootloaderPlugin (which requires GRUB2
+		template files on disk) while still testing the actual method logic.
+
+		Args:
+			bootc_has_source: Whether bootc set-options-for-source is available
+		"""
+		from tuned.plugins.plugin_bootloader import BootloaderPlugin
+
+		mock_plugin = Mock(spec=BootloaderPlugin)
+		mock_plugin._bootc_has_source = bootc_has_source
+		mock_plugin._cmdline_val = ""
+		mock_plugin._cmd = Mock()
+		mock_plugin._cmd.add_modify_option_in_file = Mock(return_value=True)
+
+		# Bind real methods to the mock so we test actual logic
+		mock_plugin._bootc_has_set_options_for_source = \
+			lambda: BootloaderPlugin._bootc_has_set_options_for_source(mock_plugin)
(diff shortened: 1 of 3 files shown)


yarda:
Thanks for the contribution, please fix the CI to pass:
```
ERROR: plugins.test_bootloader (unittest.loader.ModuleImportFailure)
----------------------------------------------------------------------
ImportError: Failed to import test module: plugins.test_bootloader
  File "/usr/lib64/python2.7/unittest/loader.py", line 252, in _find_tests
    module = self._get_module_from_name(name)
  File "/usr/lib64/python2.7/unittest/loader.py", line 230, in _get_module_from_name
    __import__(name)
  File "/builddir/build/BUILD/tuned-2.27.0/tests/unit/plugins/test_bootloader.py", line 324
SyntaxError: Non-ASCII character '\xe2' in file /builddir/build/BUILD/tuned-2.27.0/tests/unit/plugins/test_bootloader.py on line 325, but no encoding declared; see http://www.python.org/peps/pep-0263.html for details
```
It seem it's caused by the UTF-8 character long dash `—` (0xe28094) on the line 324, please use the ASCII dash `-` (0x2d).

Please fix the CodeQL findings or comment if it is wrong.

yarda:
Thanks, I don't see any major problem here. The code LGTM. I did a quick test (without the bootc) and the bootloader plugin still seems to work OK. Maybe the test coverage is too excessive, and probably AI generated, but we can relax it anytime later.

PR #847, opened 2026-04-08
gh-actions: fix doc build

yarda:
The main Makefile needs to be called to correctly export all variables like the PYTHON.


Fothsid (approved):
LGTM.

PR #852, opened 2026-04-22
net: add more coalescing options, filter out unsupported ones

Fothsid:
Ethtool's tx-aggr-* parameters were not known to tuned.

In addition to adding those, checking logic is changed to filter out parameters that are unknown to tuned but supported by ethtool, instead of simply failing to set any options.

Resolves: RHEL-152675

## Summary by CodeRabbit

* **New Features**
  * Added support for more network interface tuning options, including sample-interval and multiple TX aggregation parameters.

* **Bug Fixes**
  * Validation now skips unknown parameters with warnings (won't fail non-feature flows).
  * Feature-specific configs are accepted as provided.
  * Empty or missing device configs are handled gracefully without erroneous returns.

[Review Change Stack](https://app.coderabbit.ai/change-stack/redhat-performance/tuned/pull/852)

### DIFF
--- a/tuned/plugins/plugin_net.py
+++ b/tuned/plugins/plugin_net.py
@@ -332,7 +332,10 @@ class NetTuningPlugin(hotplug.Plugin):
 			"rx-frames-high": None,
 			"tx-usecs-high": None,
 			"tx-frames-high": None,
-			"sample-interval": None
+			"sample-interval": None,
+			"tx-aggr-max-bytes": None,
+			"tx-aggr-max-frames": None,
+			"tx-aggr-time-usecs": None
 			}
 
 	@classmethod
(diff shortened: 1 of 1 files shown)


tuned/plugins/plugin_net.py
```diff
@@ -623,9 +626,8 @@ def _check_parameters(self, context, d):
 				"channels": self._get_config_options_channels }
 		supported = set(supported_getter[context]().keys())
 		if not params.issubset(supported):
-			log.error("unknown %s parameter(s): %s" % (context, str(params - supported)))
-			return False
-		return True
+			log.warning("unknown %s parameter(s): %s" % (context, str(params - supported)))
```
yarda: Maybe add word 'ignoring' i.e. 'ignoring unknown %s parameter(s):....' for users to be sure how it is treated.

yarda (approved):
Maybe improve the text in the warning (see the inline comment) otherwise LGTM, basic functional testing passed.

tuned/plugins/plugin_net.py
Fothsid: Done.

yarda:
@coderabbitai review

yarda (approved):
LGTM

PR #871, opened 2026-08-09
Fix misspellings in user-visible messages

suraj-cmd:
Six messages that reach users contain typos:

    ppd/controller.py       "hold inititated by another process"
    plugins/plugin_systemd  "you may need to manualy run 'dracut -f'"
    daemon/controller.py    "with limited capabilites"
    hardware/inventory.py   "with limited capabilites"
    plugins/plugin_sched.   "Veryfying cgroups affinities"
    plugins/plugin_cpu.py   "no scaling governor fo '%s'"

Two of these are more than cosmetic. The first is the message body of a
`DBusException`, so it is surfaced by any client of tuned-ppd, and the second
is emitted with `log.console()`, which writes to the terminal of whoever ran
`tuned-adm`.

The "limited capabilites" warning is duplicated verbatim in
`daemon/controller.py` and `hardware/inventory.py`, so both copies are
corrected together.

Only message text is changed; no logic is touched.

This touches `tuned/hardware/inventory.py`, which #869 also modifies, but on
different lines — the two merge cleanly in either order.

### DIFF
--- a/tuned/daemon/controller.py
+++ b/tuned/daemon/controller.py
@@ -67,7 +67,7 @@ class Controller(tuned.exports.interfaces.ExportableInterface):
 				monitor.set_receive_buffer_size(udev_buffer_size)
 			except EnvironmentError:
 				log.warning("cannot set udev monitor receive buffer size, we are probably running inside " +
-					 "container or with limited capabilites, TuneD functionality may be limited")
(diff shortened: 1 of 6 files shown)


yarda:
Thanks, LGTM, the CentOS-7 CI failure is unrelated.

yarda:
It will go into tuned-2.29.0.

PR #874, opened 2026-08-13
ppd: map ACPI quiet platform profile to PPD power saver

### LINKED ISSUE #853: [FEATURE REQUEST] Add `quiet` platform profile mapping to `PPD_POWER_SAVER`
## Description

The `tuned-ppd` daemon currently maps "low-power" to `PPD_POWER_SAVER` but leaves "quiet" `unmapped`, causing issues on laptops like ASUS that support an ACPI "quiet" profile.

### Current behavior:

- When users select "quiet" profile in KDE Power Profiles daemon, nothing happens
- The "quiet" option is available in `/sys/firmware/acpi/platform_profile_choices` but has no corresponding `PPD` mapping

### Expected behavior

- "quiet" profile should map to `PPD_POWER_SAVER` similar to "low-power"
- This allows the system to properly handle the quiet profile without requiring manual patching

Current workaround (what the patch does):

```bash
# In /usr/lib/python*/site-packages/tuned/ppd/controller.py
PLATFORM_PROFILE_MAPPING = {
    ...
    "low-power": PPD_POWER_SAVER,
    "quiet": PPD_POWER_SAVER,  # <-- added
    ...
}
```

## Environment

- Fedora 42/43/44
- tuned-ppd package
- ASUS laptops (in my case, Asus TUF Gaming F15) with ACPI platform profile support

Would it be possible to add this mapping upstream so users don't need to patch manually? I already have the [patch](https://github.com/pixincreate/omaforge/blob/master/unix/fedora/install/config/hardware/asus/power.sh) that I'm applying on every update, I can raise the PR with the fixes cherry-picked from my patch.

pixincreate:
## Description

The `powersave` profile already accepts the acpi platform profile `quiet` (`platform_profile=low-power|quiet`), but the hardcoded `PLATFORM_PROFILE_MAPPING` in `tuned/ppd/controller.py` does not know about it. Laptops exposing a `quiet` acpi platform profile (e.g. asus tuf gaming f15) therefore fall back to the `unknown` profile when the user selects quiet mode, ignoring the configured power saver profile.

## Change

Add the missing mapping so that selecting the `quiet` platform profile switches tuned-ppd to the power saver profile, matching the existing `low-power` behavior:

```python
PLATFORM_PROFILE_MAPPING = {
    "low-power": PPD_POWER_SAVER,
    "quiet": PPD_POWER_SAVER,
    "balanced": PPD_BALANCED,
    "performance": PPD_PERFORMANCE
}
```

## Why

This is a user-visible behavior improvement for laptops exposing a `quiet` acpi profile (asus, some lenovo). Users currently have to patch `controller.py` manually after every update to get this working (see the workaround referenced in #853).

## Notes

- `quiet` does not appear in kde's control center. kde only exposes the three standard ppd profiles (power-saver, balanced, performance).
- the `quiet` profile comes from the kernel via the acpi platform profile interface (`/sys/firmware/acpi/platform_profile_choices` on asus tuf laptops). the fn+f5 hardware key cycles `quiet` -> `balanced` -> `performance` at the kernel level, which is the kernel initiated path `PLATFORM_PROFILE_MAPPING` exists for (same mechanism as the thinkpad fn+l handling).
- the ppd to tuned mapping in `ppd.conf` needs no change. `power-saver=powersave` already exists. the only missing piece is the acpi to ppd hop. with `quiet` unmapped, `check_platform_profile()` returns early and keeps the previous profile, so fn+f5 skips power-saver entirely.

Fixes #853

### DIFF
--- a/tuned/ppd/controller.py
+++ b/tuned/ppd/controller.py
@@ -25,6 +25,7 @@ UPOWER_DBUS_INTERFACE = "org.freedesktop.UPower"
 PLATFORM_PROFILE_PATH = "/sys/firmware/acpi/platform_profile"
 PLATFORM_PROFILE_MAPPING = {
     "low-power": PPD_POWER_SAVER,
+    "quiet": PPD_POWER_SAVER,
     "balanced": PPD_BALANCED,
     "performance": PPD_PERFORMANCE
 }


zacikpa:
The original purpose of PLATFORM_PROFILE_MAPPING is to react to changes in ACPI platform profiles initiated by the kernel, e.g. when using the corresponding function keys on new thinkpads. If I understand #853 correctly, KDE now provides the quiet profile also in its control center? If that's the case, I believe the default PPD-to-tuned profile mappings should also be updated alongside this change.

pixincreate:
@zacikpa, to clarify, `quiet` does not appear in kde's control center. kde only exposes the 3 standard ppd profiles (`power-saver`, `balanced`, `performance`). selections there go through the ppd api, not through `PLATFORM_PROFILE_MAPPING`.

the `quiet` profile comes from the kernel via the acpi platform profile interface (`/sys/firmware/acpi/platform_profile_choices` on asus tuf laptops). on my machine, the `fn+f5` hardware key cycles quiet -> balanced -> performance at the kernel level. this is exactly the kernel initiated path `PLATFORM_PROFILE_MAPPING` exists for, the same mechanism as the thinkpad fn+l handling.

that is why the ppd to tuned mapping in `ppd.conf` needs no change. `power-saver=powersave` already exists. the only missing piece is the acpi to ppd hop. with `quiet` unmapped, `check_platform_profile()` hits `'if platform_profile not in PLATFORM_PROFILE_MAPPING: return'` and keeps the previous profile, so `fn+f5` skips `power-saver`` entirely. with the added `"quiet": PPD_POWER_SAVER`, it resolves to `power-saver` -> `powersave`, which matches the expected behavior.

edit: also updated the pr description

zacikpa:
OK, I must have misunderstood
> When users select "quiet" profile in KDE Power Profiles daemon, nothing happens

In that case, this LGTM.

pixincreate:
please let me know if there exist anything else to be done from my end.

thanks in advance!

yarda:
CentOS 7 CI failure is unrelated.

yarda:
Thanks LGTM.

yarda:
It will go into TuneD 2.29.0. If you need it in Fedora sooner, please open Fedora bug.

PR #879, opened 2026-08-20
hardware: do not traceback on invalid devices_udev_regex

lilu5458:
The `devices_udev_regex` profile option (also settable through the `instance_create` D-Bus method) is used verbatim as a regular expression in `DeviceMatcherUdev.match` (`tuned/hardware/device_matcher_udev.py`). An invalid regex raised an unhandled `re.error` that propagated through `assign_free_devices` / `instance_apply_tuning` and broke profile application (and the `instance_create` D-Bus method).

This catches `re.error`, logs it and treats the device as non-matching, mirroring the fix already applied for `lscpu_check` in e83a746 (PR #867).

It also extends `ProfileRecommender.process_config` to catch `re.error`, so an invalid regex in `recommend.conf` (reachable through the unauthenticated `recommend_profile` D-Bus method) no longer crashes profile recommendation.

**Reproducer** (invalid regex):
```
devices_udev_regex=[
```
Before: `re.error: unterminated character set` traceback breaks profile apply. After: error is logged, the device is treated as non-matching.

A unit test `test_invalid_regex_does_not_traceback` is added.

Related: #867 (lscpu_check: do not traceback on invalid regex).

### DIFF
--- a/tests/unit/hardware/test_device_matcher_udev.py
+++ b/tests/unit/hardware/test_device_matcher_udev.py
@@ -34,3 +34,19 @@ class DeviceMatcherUdevTestCase(unittest.TestCase):
 				"/sys/devices/virtual/tty/tty0")
 		self.assertTrue(self.matcher.match("tty.", device))
 		self.assertFalse(self.matcher.match("tty[1-9]", device))
+
+	def test_invalid_regex_does_not_traceback(self):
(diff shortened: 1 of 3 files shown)


yarda:
CentOS 7 CI failure is unrelated.

yarda:
LGTM, thanks, it will go into TuneD 2.29.0.

PR #883, opened 2026-08-22
Fix completions to obey customized profile directories

### LINKED ISSUE #881: Completion scripts should obey the default or customized profile directory
This is related to #872. The bash and later zsh completion scripts should look for TuneD profile in correct directories. E.g. on older RHELs in `/etc/tuned` and `/usr/lib/tuned`, on recent RHELs/Fedora/... in `/etc/tuned/profiles` and `/usr/lib/tuned/profiles` and should also obey user customized paths in `/etc/tuned/tuned-main.conf`, the `profile_dirs` variable.

Speedy1894:
Instead of only the old directories, `/usr/lib/tuned/profiles,/etc/tuned/profiles`, use the `profile_dir` option in the configuration file (the default being the old constant value) or `/usr/lib/tuned,/etc/tuned` if on a version of `tuned` that doesn't support it.

Fixes #881

### DIFF
--- a/tuned-adm.bash
+++ b/tuned-adm.bash
@@ -1,5 +1,36 @@
 # bash completion for tuned-adm
 
+_tuned-adm_get-profiles()
+{
+    local config_file='/etc/tuned/tuned-main.conf'
+    local curV="$(tuned --version)"  # Convert to lowercase to be safe (below)
+    local minV='tuned 2.23.0'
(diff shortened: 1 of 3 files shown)


Speedy1894:
(The completion-adding commit is here because that branch is a parent of this one — sorry if I should've put the second commit on top of it)

Speedy1894:
Fixed the extra commit by rebasing onto current master

Issue #857, opened 2026-06-04
boost verification fails on v2.27


sjpb:
Note: This is not quite the same logic issue as https://github.com/redhat-performance/tuned/issues/842, as far as I can tell.

v2.27 updated profiles: "Set boost=1 in *-performance profiles"

Which means tuned-adm verify fails on VMs with no /sys/devices/system/cpu/cpufreq/policy%s/boost` files:

In https://github.com/redhat-performance/tuned/blob/v2.27.0/tuned/plugins/plugin_cpu.py
- _set_boost(): returns None if that file doens't exist
- _get_boost(): does the same, but presumably this is unconditionally compared to the desired "1" value.

I'm not sure which way around a better fix is TBH, would seem None should pass the verification but maybe that doesn't fix with what you've got elsewhere.

sjpb:
Actually - I think this affects more than just the *-performance profiles, ones with boost=0 (like powersave) will presumably also fail verification on such systems

sjpb:
Ok I think this is stupidity on my end; we'd been applying hpc-performance on VMs in CI as a test; but actually that just isn't appropriate (even though it worked before 2.27). Should use the virtual- profiles instead which don't set cpu requirements.

Issue #695, opened 2024-10-12
Errors when changing to balance on a Ryzen system


alosarjos:
Hi there!

I'm just giving a try to tuned instead of the regular PPD but I'm seeing some errors on the logs, not sure if an issue on my side or the program.

When changing to balance and running a `tuneadm-verify` I'm getting some errors. From the logs:

`2024-10-12 09:42:02,415 ERROR    tuned.utils.commands: Executing 'modprobe -r cpufreq_conservative' error: modprobe: FATAL: Module cpufreq_conservative is builtin.`

```
2024-10-12 09:42:02,419 ERROR    tuned.utils.commands: Writing to file '/sys/devices/system/cpu/cpufreq/policy11/boost' error: '[Errno 22] Invalid argument'
2024-10-12 09:42:02,419 INFO     tuned.plugins.plugin_cpu: Setting boost value '1' for cpu 'cpu11'
2024-10-12 09:42:02,419 ERROR    tuned.utils.commands: Writing to file '/sys/devices/system/cpu/cpufreq/policy9/boost' error: '[Errno 22] Invalid argument'
2024-10-12 09:42:02,419 INFO     tuned.plugins.plugin_cpu: Setting boost value '1' for cpu 'cpu9'
2024-10-12 09:42:02,419 ERROR    tuned.utils.commands: Writing to file '/sys/devices/system/cpu/cpufreq/policy1/boost' error: '[Errno 22] Invalid argument'
2024-10-12 09:42:02,419 INFO     tuned.plugins.plugin_cpu: Setting boost value '1' for cpu 'cpu1'
2024-10-12 09:42:02,419 ERROR    tuned.utils.commands: Writing to file '/sys/devices/system/cpu/cpufreq/policy7/boost' error: '[Errno 22] Invalid argument'
2024-10-12 09:42:02,419 INFO     tuned.plugins.plugin_cpu: Setting boost value '1' for cpu 'cpu7'
2024-10-12 09:42:02,419 ERROR    tuned.utils.commands: Writing to file '/sys/devices/system/cpu/cpufreq/policy14/boost' error: '[Errno 22] Invalid argument'
2024-10-12 09:42:02,419 INFO     tuned.plugins.plugin_cpu: Setting boost value '1' for cpu 'cpu14'
2024-10-12 09:42:02,419 ERROR    tuned.utils.commands: Writing to file '/sys/devices/system/cpu/cpufreq/policy6/boost' error: '[Errno 22] Invalid argument'
[...]

PR #890, opened 2026-08-31
cpu: write global boost node before per-policy, not as a fallback

### LINKED ISSUE #810: Enable overall cpufreq boost before enabling per-policy boost
On a Snapdragon laptop, just using the regular `[cpu] boost=1` without extra system configuration (this is on postmarketOS) I get this:

```
2025-11-11 02:36:34,284 ERROR    tuned.utils.commands: Writing to file '/sys/devices/system/cpu/cpufreq/policy4/boost' error: '[Errno 22] Invalid argument'
2025-11-11 02:36:34,284 INFO     tuned.plugins.plugin_cpu: Setting boost value '1' for cpu 'cpu4'
2025-11-11 02:36:34,284 WARNING  tuned.plugins.plugin_cpu: Unable to set boost on cpu 'cpu10'. Neither per-policy boost nor intel_pstate no_turbo is available.
2025-11-11 02:36:34,284 WARNING  tuned.plugins.plugin_cpu: Unable to set boost on cpu 'cpu6'. Neither per-policy boost nor intel_pstate no_turbo is available.
[…]
```

Well… you *first* need to write `1` to `/sys/devices/system/cpu/cpufreq/boost` before touching the "inner" `/sys/devices/system/cpu/cpufreq/policy*/boost` ones!

wistarabanyukayana:
## Problem

Some drivers/platforms gate per-policy boost control behind the package-wide global toggle: the per-policy `/sys/devices/system/cpu/cpufreq/policyN/boost` node can reject writes with `EINVAL` until the global `/sys/devices/system/cpu/cpufreq/boost` node is enabled first (e.g. amd-pstate in active/EPP mode, or some ARM/Snapdragon laptops). `EINVAL` is already handled gracefully (silenced via the existing `no_error=[errno.EINVAL]`), but nothing currently writes the global node at all, so `boost=1` in a profile silently never takes effect on affected hardware.

#810 reports this exact failure signature on a Snapdragon/ARM laptop and correctly identifies the fix: write the global node *before* touching the per-policy ones, not instead of them.

## Fix

Write the global node first, unconditionally (when it exists), then still attempt the per-policy node afterward - now that the gate is open, it may succeed on hardware with real per-core control, while remaining a harmless no-op (silenced by the existing `no_error` handling) on hardware that doesn't. This replaces an earlier version of this PR that only fell back to the global node *when* per-policy failed, which never actually opened the gate for per-policy to succeed. Applied to both `_set_boost` and `_get_boost`'s write-verify path. The existing `intel_pstate`/`no_turbo` special case is untouched and still runs after both.

## Testing

- `tests/unit/plugins/test_cpu.py`: global node is written before per-policy, per-policy is still attempted (not skipped) after a successful global write, per-policy alone still works when no global node exists, graceful no-op when neither exists, and the `_get_boost` read-side fallback. Full suite (`python -B -m unittest discover tests/unit`) passes, 141/141.
- Verified directly on real `amd-pstate-epp` hardware (Ryzen 7 5800U), with the global node forced to a *different* value first each time (so `ignore_same=True` can't mask a no-op) so the real write syscall is actually exercised:
  ```
  # global forced to 0; per-policy write of 1 fails on every core
  echo 0 > /sys/devices/system/cpu/cpufreq/boost
  echo 1 > /sys/devices/system/cpu/cpufreq/policy0/boost   # write(2) fails, EINVAL
  echo 1 > /sys/devices/system/cpu/cpufreq/policy3/boost   # write(2) fails, EINVAL
  echo 1 > /sys/devices/system/cpu/cpufreq/policy7/boost   # write(2) fails, EINVAL
  echo 1 > /sys/devices/system/cpu/cpufreq/policy15/boost  # write(2) fails, EINVAL

  # global set to 1 first; the identical per-policy write now succeeds
  echo 1 > /sys/devices/system/cpu/cpufreq/boost
  echo 0 > /sys/devices/system/cpu/cpufreq/policy0/boost   # succeeds
  echo 0 > /sys/devices/system/cpu/cpufreq/policy3/boost   # succeeds
  echo 0 > /sys/devices/system/cpu/cpufreq/policy7/boost   # succeeds
  echo 0 > /sys/devices/system/cpu/cpufreq/policy15/boost  # succeeds
  ```
  So on this hardware, per-policy control genuinely exists but is gated behind the global node, exactly as #810 describes - not a categorical absence of per-core control as an earlier version of this PR assumed.

One nuance worth flagging: a commenter on #695 demonstrated a successful raw per-policy write on their own `amd-pstate-epp` machine *without* touching the global node first, suggesting per-core availability (gated or not) is hardware/platform dependent rather than universal. That's fine either way for this fix - the global write is a no-op (`ignore_same=True`) when the per-policy node already works on its own.

Fixes #810

---

I also checked other open/closed boost-related issues before settling on the above: `#857` and `#820` turned out unrelated on closer reading (a VM misconfiguration and an already-fixed Intel `no_turbo` persistence bug, respectively). `#842` is a genuinely adjacent gap in the same functions (`tuned-adm verify` reporting `boost = None` on `intel_pstate` systems), but it's a different root cause - on a pure `intel_pstate` system, neither the per-policy *nor* the global boost file exists at all (per the kernel's own cpufreq docs, `intel_pstate` uses its own `no_turbo` interface instead), so this PR's global-node fallback doesn't reach it. Fixing that properly means reading back `intel_pstate`'s `no_turbo` value specifically, which is a distinct change - flagging it here for visibility rather than bundling it into this patch.


wistarabanyukayana:
@coderabbitai full review

wistarabanyukayana:
@coderabbitai full review

PR #892, opened 2026-09-18
bootloader: check bootc status before using source tracking

jmarrero:
The set-options-for-source help probe succeeds on package-mode systems with bootc installed, causing TuneD to bypass GRUB/BLS.

Require bootc status to report a booted deployment as well. Share the check with the daemon's bootloader unit injection and add regression tests for package-mode hosts and invalid status responses.


jmarrero:
@coderabbitai The replica in `DaemonEnsureBootloaderUnitTestCase` is deliberate. Importing `tuned.daemon.daemon` pulls in runtime-only modules (`procfs` via `profile_recommender`, dbus, pyudev) that are not available in `%check` or minimal unit-test environments, and we do not want to add a `BuildRequires` just for the tests. The replica calls the real `has_set_options_for_source`, which is where the decision logic (including the package-mode host case) lives and is covered by `tests/unit/utils/test_bootc.py`; the daemon side is two lines of glue.

jmarrero:
We can also add the build requires (`%{_py}-linux-procfs` is already a runtime Requires) and actually test the real `Daemon._ensure_bootloader_unit` instead of the replica. Not sure if the tuned team has a preference here. This is the change, verified with `make rpm` (`%check` runs 146 tests, OK) in a CentOS Stream 10 build root; kept local for now:

```diff
diff --git a/tests/unit/plugins/test_bootloader.py b/tests/unit/plugins/test_bootloader.py
index 8b704e8..1942e10 100644
--- a/tests/unit/plugins/test_bootloader.py
+++ b/tests/unit/plugins/test_bootloader.py
@@ -342,34 +342,21 @@ class BootloaderPluginBootcTestCase(unittest.TestCase):
 
 
 class DaemonEnsureBootloaderUnitTestCase(unittest.TestCase):
-	"""Tests for the _ensure_bootloader_unit logic in Daemon.
-
-	The Daemon class cannot be imported directly in unit tests because
-	tuned.daemon.__init__ triggers a chain of imports (pyudev, dbus,
-	procfs, etc.) that are unavailable in the test environment.  Instead,
-	we replicate the _ensure_bootloader_unit logic inline - the method is
-	simple enough that this is reliable and avoids import gymnastics.  The
-	capability check is the real one shared with the bootloader plugin.
-	"""
+	"""Tests for Daemon._ensure_bootloader_unit, run against the real method
+	on a mock daemon."""
 
 	@staticmethod
 	def _ensure_bootloader_unit(daemon_self):
-		"""Replica of Daemon._ensure_bootloader_unit for testing."""
-		import collections
-		from tuned.utils.bootc import has_set_options_for_source
-		if "bootloader" in daemon_self._profile.units:
-			return
-		if not has_set_options_for_source(daemon_self._cmd):
-			return
-		from tuned.profiles.unit import Unit
-		daemon_self._profile.units["bootloader"] = Unit(
-			"bootloader", collections.OrderedDict())
+		from tuned.daemon.daemon import Daemon
+		Daemon._ensure_bootloader_unit(daemon_self)
 
 	def _make_daemon_mock(self, has_bootloader_unit=False, bootc_available=True):
 		"""Create a minimal mock simulating Daemon state."""
 		import collections
+		from tuned.daemon.daemon import Daemon
 
-		daemon = Mock()
+		# spec= so the unbound-method call also works on Python 2
+		daemon = Mock(spec=Daemon)
 		daemon._cmd = Mock()
 		if bootc_available:
 			daemon._cmd.execute = Mock(return_value=(0, "Usage: ...", ""))
diff --git a/tuned.spec b/tuned.spec
index 4910717..adf67d0 100644
--- a/tuned.spec
+++ b/tuned.spec
@@ -84,6 +84,8 @@ BuildRequires: %{_py}-devel
 BuildRequires: %{_py}-mock
 %endif
 BuildRequires: %{_py}-pyudev
+# the daemon unit tests import tuned.daemon, which needs linux-procfs
+BuildRequires: %{_py}-linux-procfs
 Requires: %{_py}-pyudev
 Requires: %{_py}-linux-procfs
 Requires: %{_py}-inotify
```

This is just test code, so we can do this in another PR after solving the issue too.

cgwalters:
The PR upstream here should link to https://redhat.atlassian.net/browse/RHEL-261119 I think.

I am confused, the PR description is talking about "package mode" but isn't the case here rpm-ostree-but-not-bootc?

The distinction between those two cases is *VERY* subtle (when bootc is using the ostree backend) and so I really want to better understand the chain of events here.

jmarrero:
Well there is two things happening, in very little blast radius for both I think.  

1) In package mode systems that decide to install bootc and tuned. Tuned will try to use bootc to set the kargs and fail.
2) In rpm-ostree/bootc systems that have a new bootc but old ostree. Tuned will call bootc successfully but bootc will call ostree, fail and log it.

cgwalters:
> In package mode systems that decide to install bootc and tuned. Tuned will try to use bootc to set the kargs and fail.

Yeah that's clearly worth fixing.

> In rpm-ostree/bootc systems that have a new bootc but old ostree. Tuned will call bootc successfully but bootc will call ostree, fail and log it.

Ah. Wouldn't the most likely fix here be to bootc to detect when we have an older ostree? I think that should be possible?

jmarrero:
Should be possible IIRC on the original PR my inclination to not also check for ostree was because we already do calls to bootc and rpm-ostree. I was thinking that I would publish ostree and bootc in tandem. Which did not happen.

cgwalters:
For sure missing release automation + version skew is a big root problem here; not any one person's fault.

But I think the thing is for the work you've been doing at a practical level we basically need the version detection anyways for cases like CI upstream, like we want to be able to be able to land a PR to bootc that conditionally uses a new ostree feature without requiring that ostree actually land in *every supported stream* right? (We actually CI check Fedora 43 right now which is soon to be EOL and which I want to just turn off)

IOW it seems to me the more right fix again is version detection in bootc and not trying to blast a newer ostree everywhere?

cgwalters:
This is a VERY complex topic, and I could be wrong here but:

I believe this PR will fix the reported symptoms, and it should be relatively safe.

HOWEVER - I think the code is based on an incorrect premise that the problem is "bootc vs ostree". I don't think it's that, it's basically "too old ostree in target OS".

Or to flip it around, which is that in theory this "fix karg handling for tuned" should actually work with non-bootc ostree systems too. This approach will turn that off...which I don't think is wrong, but I wouldn't say it's right either.

What I think would be cleaner:

- Detect the version of ostree here (`ostree --version` parsing?)
- Patch bootc to do that version detection

